PR #11, opened 2022-08-10
chore: remove node ports

tiagolobocastro:
chore: remove node ports

Our kubectl-plugin no longer needs node ports to reach our services.

---

chore: temporarily pin io-engine to v0

This will be changed to v1 when the control-plane supports v1.

---

fix: run pre-commit hooks under the nix-shell

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -19,6 +19,7 @@ mkShell {
   buildInputs = [
     cargo-expand
     cargo-udeps
+    commitlint
     git
     pkg-config
     pre-commit
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,14 +12,14 @@ repos:
     -   id: rust-lint
         name: Rust lint
         description: Run cargo clippy on files included in the commit.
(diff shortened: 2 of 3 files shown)


Abhinandan-Purkait (approved):

vharsh (approved):
lgtm

shell.nix
```diff
@@ -20,6 +20,7 @@ mkShell {
     cargo-expand
     cargo-udeps
     git
+    nodejs
```
vharsh, line 23: where's nodejs required?

tiagolobocastro: https://github.com/openebs/mayastor-extensions/blob/13a77d12e2d0224ef457b1c86676721c4fd6b155/.pre-commit-config.yaml#L30
I've actually changed it now to install commitlint directly into the nix-shell, it'll still use nodejs because it's written for nodejs but it becomes a bit more explicit I think, what do you think?

vharsh: Thanks, this is good if the commitlint's version matches our requirements.

tiagolobocastro:
bors merge

PR #13, opened 2022-08-11
feat(upgrade): add nix build and layout

rajaSahil:
## Description
- Add Project Layout
  -  ```
      operators/
      ├── Cargo.toml
      ├── common
      │   └── src
      │       ├── helm
      │       │   ├── client
      │       │   └── mod.rs
      │       └── lib.rs
      └── upgrade
          └── src
              ├── bin
              │   └── main.rs
              ├── controller
              ├── crd
              │   ├── mod.rs
              │   └── versions
              └── rest

      ```
- Add nix build


upgrade-operator/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "upgrade-operator"
```
tiagolobocastro, line 2: Should this package be operators? and then we have a an operator-upgrade binary?

upgrade-operator/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+description = "Upgrade Operator"
+version = "0.1.0"
+edition = "2021"
+authors = ["Sahil Raja <sahil.raja@datacore.com>"]
```
tiagolobocastro, line 6: Nit - adding authors to a rust crate is no longer considered best practices in rust, and may eventually be deprecated and removed.

upgrade-operator/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[[bin]]
+name = "upgrade-operator"
+path = "controller/src/bin/main.rs"
+
+[lib]
+name = "components"
```
tiagolobocastro, line 13: components seems to be a very open ended name to a library.. what about operator?

upgrade-operator/Cargo.toml
rajaSahil: Yep, makes sense.

upgrade-operator/Cargo.toml
rajaSahil: Have used common instead of components.

tiagolobocastro: common also could be confusing because if other crates need to access the upgrade operator they'd use `common::SomeType` which doesn't make sense for them because it's only common for the upgrade operator itself right?
out of these 3, operator would be my choice.

rajaSahil: Okay. 
So name should be operator and directory name can be common?

Abhinandan-Purkait (approved):

upgrade-operator/Cargo.toml
tiagolobocastro: The rust standard is simply `src/lib.rs` but it's up to you.

tiagolobocastro (approved):
Thanks for doing the nix part :+1: we have another nixer now :)

upgrade-operator/Cargo.toml
rajaSahil: Okay. Will do changes of the layout while implementing different modules if required.

rajaSahil:
bors merge

PR #17, opened 2022-08-16
feat(upgrade): add crd and k8s api resources

rajaSahil:
Signed-off-by: rajaSahil <sahilraja242@gmail.com>

- Adds crd(UpgradeAction)
- Adds rbac
- Adds deployement, service for operator

### NOTE:
- There can be changes in the code while implementing different parts. Mainly addition of functions.

### DIFF
--- a/operators/src/lib.rs
+++ b/operators/src/lib.rs
@@ -1,0 +1,5 @@
+#![warn(missing_docs)]
+#![allow(dead_code)]
+//! library for upgrade operator
+/// The upgrade operator
+pub mod upgrade;
--- a/operators/src/upgrade/common/constants.rs
+++ b/operators/src/upgrade/common/constants.rs
@@ -1,0 +1,46 @@
+/// macros to define group used in the upgrade crd.
+#[macro_export]
+macro_rules! upgrade_group {
+    () => {
+        "openebs.io"
+    };
+    ($s:literal) => {
+        concat!(upgrade_group!(), "/", $s)
+    };
+}
+
+/// macros to define labels for upgrade operator.
+#[macro_export]
+macro_rules! upgrade_labels {
+    ($s:expr) => {
+        btreemap! {
+            APP => $s,
+            LABEL => $s,
+        }
+        .iter()
+        .map(|(k, v)| (k.to_string(), v.to_string()))
+        .collect()
+    };
+}
+
+/// label used for upgrade operator.
+pub(crate) const APP: &str = "app.kubernetes.io/component";
+/// label used for upgrade operator.
+pub(crate) const LABEL: &str = upgrade_group!("component");
+
+/// Upgrade operator.
+pub(crate) const UPGRADE_OPERATOR: &str = "upgrade-operator";
+/// Service account name for upgrade operator.
+pub(crate) const UPGRADE_OPERATOR_SERVICE_ACCOUNT: &str = "upgrade-operator-service-account";
+/// Role constant for upgrade operator.
+pub(crate) const UPGRADE_OPERATOR_CLUSTER_ROLE: &str = "upgrade-operator-role";
+/// Role binding constant for upgrade operator.
+pub(crate) const UPGRADE_OPERATOR_CLUSTER_ROLE_BINDING: &str = "upgrade-operator-role-binding";
+/// Deployment constant for upgrade operator.
+pub(crate) const UPGRADE_CONTROLLER_DEPLOYMENT: &str = "upgrade-operator-deployment";
+/// Service name constant for upgrade operator.
+pub(crate) const UPGRADE_OPERATOR_SERVICE: &str = "upgrade-operator-service";
+/// Service port constant for upgrade operator.
+pub(crate) const UPGRADE_OPERATOR_SERVICE_PORT: i32 = 8080;
+/// Service internal port constant for upgrade operator.
+pub(crate) const UPGRADE_OPERATOR_INTERNAL_PORT: i32 = 8080;
--- a/operators/src/upgrade/k8s/crd/mod.rs
+++ b/operators/src/upgrade/k8s/crd/mod.rs
@@ -1,0 +1,16 @@
+/// CRD definiton.
+pub mod v0;
+
+use lazy_static::lazy_static;
+pub use semver::Version;
+
+lazy_static! {
+    // Regex gathered from semver.org as the recommended semver validation regex.
+    static ref SEMVER_RE: regex::Regex = regex::Regex::new(
+        concat!(
+            r"^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)",
+            r"(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?",
+            r"(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"
+        ))
+        .expect("Invalid regex literal.");
+}
--- a/operators/src/upgrade/k8s/crd/v0.rs
+++ b/operators/src/upgrade/k8s/crd/v0.rs
@@ -1,0 +1,189 @@
+use crate::upgrade::k8s::crd::SEMVER_RE;
+use kube::CustomResource;
+use schemars::JsonSchema;
+pub use semver::Version;
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, str::FromStr};
+use validator::Validate;
+
+#[derive(
+    CustomResource,
+    Serialize,
+    Deserialize,
+    Debug,
+    Default,
+    Eq,
+    PartialEq,
+    Clone,
+    Validate,
+    JsonSchema,
+)]
+#[kube(
+    group = "openebs.io",
+    version = "v0",
+    kind = "UpgradeAction",
+    singular = "upgradeaction",
+    plural = "upgradeactions",
+    namespaced,
+    status = "UpgradeActionStatus",
+    derive = "Default",
+    derive = "PartialEq",
+    shortname = "ua",
+    printcolumn = r#"{"name":"Components state", "type":"string", "jsonPath":".status.components_state"}"#,
+    printcolumn = r#"{"name":"Target Version", "type":"string", "jsonPath":".spec.target_version"}"#,
+    printcolumn = r#"{"name":"Current Version", "type":"string", "jsonPath":".spec.current_version"}"#
+)]
+
+/// Define spec for upgrade action.
+pub struct UpgradeActionSpec {
+    /// Records the current version of the product.
+    #[validate(regex = "SEMVER_RE")]
+    current_version: String,
+    /// Records the desired version of the product.
+    #[validate(regex = "SEMVER_RE")]
+    target_version: String,
+    /// Records all the components present for the product.
+    components: HashMap<String, Vec<String>>,
+}
+
+impl UpgradeActionSpec {
+    /// Create a new upgrade action spec.
+    pub fn new(
+        current_version: Version,
+        target_version: Version,
+        componens: HashMap<String, Vec<String>>,
+    ) -> Self {
+        UpgradeActionSpec {
+            current_version: current_version.to_string(),
+            target_version: target_version.to_string(),
+            components: componens,
+        }
+    }
+
+    /// Returns the current version of the product.
+    pub fn current_version(&self) -> Version {
+        Version::from_str(&self.current_version).unwrap()
+    }
+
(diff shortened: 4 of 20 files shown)


operators/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+pub mod upgrade;
```
Abhinandan-Purkait, line 1: doc comments?

operators/src/upgrade/common/constants.rs
```diff
@@ -0,0 +1,36 @@
...
+            APP => $s,
+            LABEL => $s,
+        }
+        .iter()
+        .map(|(k, v)| (k.to_string(), v.to_string()))
+        .collect()
+    };
+}
+
+pub const APP: &str = "app.kubernetes.io/component";
```
Abhinandan-Purkait, line 25: doc comments?

operators/src/upgrade/common/constants.rs
```diff
@@ -0,0 +1,36 @@
...
+    };
+    ($s:literal) => {
+        concat!(upgrade_group!(), "/", $s)
+    };
+}
+
+#[macro_export]
+macro_rules! upgrade_labels {
+    ($s:expr) => {
+        btreemap! {
```
tiagolobocastro, line 15: what does this do?
Why does it need to be a macro? If possible functions are always more readable

operators/src/upgrade/common/constants.rs
tiagolobocastro, line 25: Why are all these pub, do we need to expose these to an api?

operators/src/upgrade/k8s/crd/mod.rs
```diff
@@ -0,0 +1,14 @@
+use lazy_static::lazy_static;
+pub use semver::Version;
+pub mod v0;
+
+lazy_static! {
+    // Regex gathered from semver.org as the recommended semver validation regex.
+    static ref SEMVER_RE: regex::Regex = regex::Regex::new(
```
tiagolobocastro, line 7: I'm confused, if we're using the semver crate already, why are we using our own regex?

operators/src/upgrade/k8s/crd/v0.rs
```diff
@@ -0,0 +1,223 @@
...
+    Debug,
+    Default,
+    Eq,
+    PartialEq,
+    Clone,
+    Validate,
+    JsonSchema,
+)]
+#[kube(
+    group = "openebs.io",
```
tiagolobocastro, line 24: this should be read from a const?

operators/src/upgrade/k8s/crd/v0.rs
```diff
@@ -0,0 +1,223 @@
...
+    shortname = "ua",
+    printcolumn = r#"{"name":"Components state", "type":"string", "jsonPath":".status.components_state"}"#,
+    printcolumn = r#"{"name":"Target Version", "type":"string", "jsonPath":".spec.target_version"}"#,
+    printcolumn = r#"{"name":"Current Version", "type":"string", "jsonPath":".spec.current_version"}"#
+)]
+
+pub struct UpgradeActionSpec{
+    /// Records the current version of the product
+    #[validate(required,regex="SEMVER_RE")]
+    current_version: Option<String>,
```
tiagolobocastro, line 42: rather than String, can this just be semver::Version ?

operators/src/upgrade/k8s/crd/v0.rs
```diff
@@ -0,0 +1,223 @@
...
+    #[validate(required,regex="SEMVER_RE")]
+    target_version:Option<String>,
+    /// Records all the components present for the product
+    #[validate(required)]
+    components:Option<HashMap<String,Vec<String>>>
+}
+
+impl UpgradeActionSpec{
+
+    pub fn new(current_version:Option<Version>, target_version:Option<Version>, componens: Option<HashMap<String,Vec<String>>>)-> Self{
```
tiagolobocastro, line 53: Do all these need to be pub?

tiagolobocastro (commented):
What is the plan with this crate, why do we need a library?

operators/src/upgrade/k8s/crd/v0.rs
```diff
@@ -0,0 +1,223 @@
...
+    /// Last time the condition transit from one status to another.
+    pub last_transition_time: Option<String>,
+
+    ///Human readable message indicating details about last transition.
+    pub message: Option<String>,
+
+    ///(one line) reason for the condition's last transition.
+    pub reason: Option<String>,
+
+    /// Status of the cpondition, one of True, False, Unknown.
```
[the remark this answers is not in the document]
rajaSahil: Controller updates the condition and when a new CR is created with all the component details, default condition is going to be created. Conditions and upgrade phases are inter linked to each other.
But instead of using conditions, I am thinking of using state and last_transition_time in UpgradeActionStatus. it would also work and can make our work easier.

rajaSahil:
> What is the plan with this crate, why do we need a library?

Suppose in future, we plan to build another operator for mayastor, we can simply include new bin dir and add that operator as a lib.

operators/src/upgrade/common/constants.rs
rajaSahil: No, pub is not required. Will update it.

operators/src/upgrade/common/constants.rs
rajaSahil: It creates labels for k8s resources, such as:
app.kubernetes.io/component: upgrade-operator
component:upgrade-operator

operators/src/upgrade/k8s/crd/mod.rs
rajaSahil: We would have to implement JsonSchema, Serialize and Deserialize for Version.

tiagolobocastro:
> > What is the plan with this crate, why do we need a library?
> 
> Suppose in future, we plan to build another operator for mayastor, we can simply include new bin dir and add that operator as a lib.

But we wouldn't need upgrade specific things in the library? The less we expose, the harder it is to misuse things.

operators/src/upgrade/common/constants.rs
tiagolobocastro: But why does it need to be a macro as opposed to a function?

operators/src/upgrade/k8s/crd/v0.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+#[derive(Deserialize, Serialize, Clone, Debug,Eq, PartialEq, JsonSchema)]
+pub struct UpgradeCondition{
+    /// Type of upgrade condition
+    #[validate(required)]
+    pub type_: Option<UpgradeConditionType>,
+
+    /// Last time the condition transit from one status to another.
+    pub last_transition_time: Option<String>,
```
tiagolobocastro, line 168: Should this not be Option<Utc> ?

operators/src/upgrade/common/constants.rs
rajaSahil: There is no reason for this, we can use function as well.

operators/src/upgrade/k8s/crd/v0.rs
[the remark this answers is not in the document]
tiagolobocastro: Could replace String with an enum, example:
enum ConditionStatus { Unknown, True, False } ?

operators/src/upgrade/k8s/crd/mod.rs
tiagolobocastro: To enable `Serialize/Deserialize` you have to enable the `serde` feature in the `semvercrate`. Why do you need JsonSchema? Change ` current_version: Option<String>` to ` current_version: semver::Version,` and it becomes required?

operators/src/upgrade/k8s/crd/v0.rs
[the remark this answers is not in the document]
Abhinandan-Purkait: > Controller updates the condition and when a new CR is created with all the component details, default condition is going to be created. Conditions and upgrade phases are inter linked to each other. But instead of using conditions, I am thinking of using state and last_transition_time in UpgradeActionStatus. it would also work and can make our work easier.

Yeah. Otherwise maintaining conditions will be an overhead and can generate wrong notions if not maintained properly.

operators/src/upgrade/k8s/crd/mod.rs
rajaSahil: JsonSchema is required for Spec of CRD. 
And to use semver::Version we will have to implement JsonSchema for semver::Version.

operators/src/upgrade/k8s/crd/v0.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub fn components(&self) -> Option<HashMap<String, Vec<String>>> {
+        self.components.clone()
+    }
+}
+
+/// 'UpgradePhase' defines the status of each components
+#[derive(Deserialize, Serialize, Clone, Debug, Eq, PartialEq, JsonSchema)]
+pub enum UpgradePhase {
+    Waiting,
+
```
tiagolobocastro, line 88: extra emptyline between states

operators/src/upgrade/k8s/crd/v0.rs
```diff
@@ -0,0 +1,189 @@
...
+    fn from(u: UpgradePhase) -> Self {
+        u.to_string()
+    }
+}
+
+/// 'UpgradeState' defines the status of upgradeaction resource
+#[derive(Deserialize, Serialize, Clone, Debug, Eq, PartialEq, JsonSchema)]
+pub enum UpgradeState {
+    NotStarted,
+
```
tiagolobocastro, line 128: emptyline between states

operators/src/upgrade/k8s/crd/v0.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// Returns all the components
+    pub fn components(&self) -> Option<HashMap<String, Vec<String>>> {
+        self.components.clone()
+    }
+}
+
+/// 'UpgradePhase' defines the status of each components
+#[derive(Deserialize, Serialize, Clone, Debug, Eq, PartialEq, JsonSchema)]
+pub enum UpgradePhase {
+    Waiting,
```
tiagolobocastro, line 87: missing doc comments on each enum state

operators/src/upgrade/k8s/crd/mod.rs
tiagolobocastro: If that's too much to get it working then I'm fine with the regex for now.

operators/src/upgrade/k8s/crd/v0.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
+pub struct UpgradeActionStatus {
+    /// UpgradeAction state
+    state: Option<UpgradeState>,
+    /// Last time the condition transit from one status to another.
+    last_transition_time: String,
+    /// Components State
+    #[validate(required)]
```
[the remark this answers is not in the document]
rajaSahil: Yes, this is required.

operators/src/upgrade/k8s/crd/v0.rs
```diff
@@ -0,0 +1,196 @@
...
+    shortname = "ua",
+    printcolumn = r#"{"name":"Components state", "type":"string", "jsonPath":".status.components_state"}"#,
+    printcolumn = r#"{"name":"Target Version", "type":"string", "jsonPath":".spec.target_version"}"#,
+    printcolumn = r#"{"name":"Current Version", "type":"string", "jsonPath":".spec.current_version"}"#
+)]
+
+/// Define spec for upgrade action.
+pub struct UpgradeActionSpec {
+    /// Records the current version of the product.
+    #[validate(required, regex = "SEMVER_RE")]
```
tiagolobocastro, line 42: If all of these are "required", why not simply drop the Option?

rajaSahil: Ahh, we cannot remove Option since validate_required(..) expects an Option value.

tiagolobocastro: That's what I mean, if we drop the Option don't need to add required because it'll be explicitly required?

rajaSahil: Oh Okay. Got it.

tiagolobocastro (approved):
Please squash your commits before merging

Abhinandan-Purkait (approved):

rajaSahil:
bors try

rajaSahil:
bors try

rajaSahil:
bors merge

rajaSahil:
bors merge

rajaSahil:
bors merge

rajaSahil:
bors merge

PR #26, opened 2022-09-06
feat(upgrade): add helm client

rajaSahil:
Signed-off-by: rajaSahil <sahilraja242@gmail.com>

- Add helm client
  - Runs helm commands in the container.

### DIFF
--- a/operators/src/upgrade/common/error.rs
+++ b/operators/src/upgrade/common/error.rs
@@ -1,0 +1,18 @@
+/// Helm error to be used for client.
+#[derive(Debug, Clone)]
+pub(crate) enum Error {
+    /// Helm not installed error.
+    HelmNotInstalled(String),
+    /// Error while running helm command.
+    HelmStd(String),
+    /// Error when specific helm version not found.
+    HelmVersionNotFound(String, String),
+    /// Error when helm chart is not present in the cluster.
+    HelmChartNotFound(String),
+    /// Error when converting utf8 to string.
+    Utf8(String),
+    /// Error while running helm get values command.
+    HelmGetValues(String),
+    /// Deserialization error for helm client.
+    SerdeDeserialization(String),
+}
--- a/operators/src/upgrade/common/mod.rs
+++ b/operators/src/upgrade/common/mod.rs
@@ -1,2 +1,4 @@
 /// Constants to be used in the crate.
 pub mod constants;
+/// Error module.
+pub(crate) mod error;
--- a/operators/src/upgrade/helm/client.rs
+++ b/operators/src/upgrade/helm/client.rs
@@ -1,0 +1,263 @@
+use serde::Deserialize;
+use std::{
+    ffi::OsStr,
+    path::PathBuf,
+    process::{Command, Output},
+};
+
+use crate::upgrade::common::error::Error;
+
+/// Helm arguments that are required to run helm commands.
+#[derive(Debug, Clone, Default)]
+struct HelmArgs {
+    name: Option<String>,
+    opts: Vec<(String, String)>,
+    namespace: Option<String>,
+    values: Vec<PathBuf>,
+}
+
+impl HelmArgs {
+    /// Set a name.
+    fn with_name(mut self, name: Option<String>) -> Self {
+        self.name = name;
+        self
+    }
+
+    /// Set a single option.
+    fn with_opt(mut self, key: String, value: String) -> Self {
+        self.opts.push((key, value));
+        self
+    }
+
+    /// Reset array of options.
+    fn with_opts(mut self, options: Vec<(String, String)>) -> Self {
+        self.opts = options;
+        self
+    }
+
+    /// Set namespace.
+    fn with_namespace(mut self, ns: Option<String>) -> Self {
+        self.namespace = ns;
+        self
+    }
+
+    /// Set values.
+    fn with_values(mut self, values: Vec<PathBuf>) -> Self {
+        self.values = values;
+        self
+    }
+
+    /// Set one value.
+    fn with_value(mut self, value: PathBuf) -> Self {
+        self.values.push(value);
+        self
+    }
+
+    /// Get a name.
+    fn name(&self) -> Option<String> {
+        self.name.clone()
+    }
+
+    /// Get options.
+    fn opts(&self) -> &Vec<(String, String)> {
+        &self.opts
+    }
+
+    /// Get namespace.
+    fn namespace(&self) -> Option<String> {
+        self.namespace.clone()
+    }
+
+    /// Get values.
+    fn values(&self) -> &Vec<PathBuf> {
+        &self.values
+    }
+
+    /// Apply arguments for helm command.
+    fn apply_args(&self, command: &mut Command) {
+        if self.name().is_some() {
+            command.arg(self.name().as_ref().unwrap());
+        }
+        if self.namespace().is_some() {
+            command
+                .arg("--namespace")
+                .arg(self.namespace().as_ref().unwrap());
+        }
+
+        for value_path in self.values() {
+            command.arg("--values").arg(value_path);
+        }
+
+        for (key, val) in self.opts() {
+            command.arg("--set").arg(format!("{}={}", key, val));
+        }
+    }
+
+    /// Run upgrade command.
+    fn upgrade(self) -> Result<Output, Error> {
+        self.run(["upgrade", "--output=json"])
+    }
+
+    /// Run get values command.
+    fn get_values(self) -> Result<Output, Error> {
+        self.run(["get", "values", "--output=json"])
+    }
+
+    /// Run helm version command.
+    fn version(self) -> Result<Output, Error> {
+        self.run(["version", "--short"])
+    }
+
+    /// Run helm ls command.
+    fn ls(self, exact_match: &String) -> Result<Output, Error> {
+        self.run(["ls", "--filter", exact_match, "--output=json"])
+    }
+
+    /// Create helm command and run.
+    fn run<I, S>(self, args: I) -> Result<Output, Error>
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<OsStr>,
+    {
+        let mut command = Command::new("helm");
+        for arg in args {
+            command.arg(arg.as_ref());
+        }
+        self.apply_args(&mut command);
+        let output = command
+            .output()
+            .map_err(|_| Error::HelmStd("Error while running helm get command".to_string()))?;
+        Ok(output)
+    }
+}
+
+/// Chart representation.
+#[derive(Deserialize, Debug, Clone, Default)]
+struct Chart {
+    name: String,
+    namespace: String,
+    revision: String,
+    updated: String,
+    status: String,
+    chart: String,
+    app_version: String,
+}
+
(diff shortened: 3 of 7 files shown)


operators/src/upgrade/common/error.rs
```diff
@@ -0,0 +1,18 @@
+/// Helm error to be used for client.
```
tiagolobocastro, line 1: why is everything pub?
Is the the upgrade operator a library or a binary?

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,249 @@
+use serde::Deserialize;
+use std::{path::PathBuf, process::{Command, Output}};
+
+use crate::upgrade::common::error::HelmError;
+
+/// helm arguments that are required to run helm commands
```
tiagolobocastro, line 6: Start with `C`apitals and end with `.`

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,249 @@
...
+pub struct HelmClient {
+    chart: Chart,
+    version: Option<String>,
+    args: HelmArgs,
+}
+
+impl HelmClient {
+    /// create a new helm client if helm is installed
+    pub fn new() -> Result<HelmClient, HelmError> {
+        let output = Command::new("helm")
```
tiagolobocastro, line 136: why not use a crate instead, such as https://github.com/stephenmuss/helm-api (haven't checked if it works or is appropriate, just a quick google search) ?

tiagolobocastro (commented):
Can we get a description of what the "helm client" is on the commit? Will it run inside the operator binary in a container etc

tiagolobocastro:
upgrade-operator should be operator-upgrade to conform with the nomenclature, please update the cargo and nix to reflect this.

operators/src/upgrade/helm/client.rs
rajaSahil: This is for helm 2 and tiller components have been removed from helm 3.

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,249 @@
...
+pub struct HelmArgs {
+    name: Option<String>,
+    opts: Vec<(String, String)>,
+    namespace: Option<String>,
+    values: Vec<PathBuf>,
+}
+
+impl HelmArgs {
+    /// set a name
+    pub fn set_name(&mut self, name: String) {
```
Abhinandan-Purkait, line 17: Should these be `pub(crate)` assuming it will be a binary.

operators/src/upgrade/common/error.rs
rajaSahil: Upgrade operator is a binary. Have updated the code.

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,249 @@
...
+        self.values = values;
+    }
+
+    /// Set one value.
+    fn set_value(&mut self, value: PathBuf) {
+        self.values.push(value);
+    }
+
+    /// Set a name.
+    fn get_name(&self) -> &String {
```
tiagolobocastro, line 50: nit: in rust we tend to skip the `get_`, example: `fn name(&self) -> ...`

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,249 @@
...
+            return Err(Error::HelmVersionNotFound(
+                "Helm version 3 not installed, installed version:{}".to_string(),
+                out_str,
+            ));
+        }
+
+        // If checks succeed, create Helm client.
+        Ok(Self {
+            args: HelmArgs::default(),
+            version: Some(out_str),
```
tiagolobocastro, line 157: why is version an option here we if we always assign it a value?

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,249 @@
...
+struct HelmArgs {
+    name: Option<String>,
+    opts: Vec<(String, String)>,
+    namespace: Option<String>,
+    values: Vec<PathBuf>,
+}
+
+impl HelmArgs {
+    /// Set a name.
+    fn set_name(&mut self, name: String) {
```
tiagolobocastro, line 20: nit: Looks like this is used to build the `HelmClient`. Consider using the builder pattern instead of set_ functions.

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,249 @@
...
+    fn set_opt(&mut self, key: String, value: String) {
+        self.opts.push((key, value));
+    }
+
+    /// Reset array of options.
+    fn set_opts(&mut self, options: Vec<(String, String)>) {
+        self.opts = options;
+    }
+
+    /// Set namepsace.
```
tiagolobocastro, line 34: namespace

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,249 @@
...
+    fn set_namespace(&mut self, ns: String) {
+        self.namespace = Some(ns);
+    }
+
+    /// Set one value.
```
tiagolobocastro, line 39: Isn't this setting all values?

operators/src/upgrade/helm/client.rs
tiagolobocastro, line 50: /// Get the name of the command?

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,249 @@
...
+    path::PathBuf,
+    process::{Command, Output},
+};
+
+use crate::upgrade::common::error::Error;
+
+/// Helm arguments that are required to run helm commands.
+#[derive(Debug, Clone, Default)]
+struct HelmArgs {
+    name: Option<String>,
```
tiagolobocastro, line 12: Is there a command without a name?

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        let x = String::from_utf8(output.stdout)
+            .map_err(|_| Error::Utf8Error("Unable to parse values into string".to_string()))?;
+        println!("{:?}", x);
+
+        Ok(())
+    }
+
+    /// Run helm commands.
+    fn run_command(&self, command_type_args: Vec<String>) -> Result<Output, Error> {
```
tiagolobocastro, line 237: Isn't it a bit confusing that we have `command_type_args` here, but within we add more `args` ?
What about using a builder like patter, to build a helm command on demand explicitly with types, example:

```rust
impl HelmClient { 
  fn args(&self) -> HelmArgs { self.default_args.clone() }
}
impl HelmArgs { fn with_xxx(self) -> Self {} fn run(self) -> Result<O,E> {} ... }
let client = HelmClient ...

/// this way it's explicit that we have an args
client.args().with_cmd(vec!["this", "that"]).run()?;
/// and easy to override options, example:
client.args().with_namespace("bob").with_cmd(vec!["this", "that"]).run()?;
```

operators/src/upgrade/common/error.rs
```diff
@@ -0,0 +1,18 @@
...
+#[derive(Debug, Clone)]
+pub enum Error {
```
tiagolobocastro, line 3: pub(crate) ?

operators/src/upgrade/common/mod.rs
```diff
@@ -1,2 +1,4 @@
 /// Constants to be used in the crate.
 pub mod constants;
+/// Error module.
+pub mod error;
```
tiagolobocastro, line 4: pub(crate) ?

operators/src/upgrade/helm/client.rs
```diff
@@ -9,62 +9,117 @@ use crate::upgrade::common::error::Error;
...
+        let mut cmd = self.clone().command(vec!["upgrade".to_string()]);
+        self.run(&mut cmd)
+    }
+
+    // Run get values command.
+    fn get_values(self) -> Result<Output, Error> {
+        let mut cmd = self
+            .clone()
+            .command(vec!["get".to_string(), "values".to_string()]);
+        self.run(&mut cmd)
```
tiagolobocastro, line 102: do we need to create a command at all?
Why not simply:
```rust
fn run(self, command_type_args: Vec<String>) -> {
  let mut command = Command::new("helm");
  for c in command_type_args {
      command.arg(c);
  }
  self.apply_args(&mut command);
  command.arg("--output=json");
  command.output()
      .map_err(|_| Error::HelmStd("Error while running helm get command".to_string()))
}
...
self.run(vec!["get".to_string(), "values".to_string()])

```

tiagolobocastro: Btw, since you're using `Command` why not implementing it like they do, and instead of using Vec<String> which required to do `vec!["get".to_string(), "values".to_string()]`, use traits so you can pass &str instead: https://doc.rust-lang.org/std/process/struct.Command.html#method.args
So it could become:
```rust
self.run(["get", "values"])
```

operators/src/upgrade/helm/client.rs
```diff
@@ -104,13 +159,12 @@ impl Chart {
             .arg("--output")
             .arg("json");
 
-        let output = command.output().map_err(|_| {
-            Error::HelmStdError("Error while running helm ls command".to_string())
-        })?;
+        let output = command
```
tiagolobocastro, line 162: Above, can we make use of our type instead of calling the binary directly?
```rust
HelmArgs::default()
            .with_name(None)
            .with_namespace(self.chart.namespace.clone())
            .run(["ls", "--filter", exact_match.as_str()])?;
```
Ah in this case we wouldn't have the name in the command? Maybe name should then be an option?

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,266 @@
...
+    pub(crate) fn with_chart(mut self, name: String, namespace: String) -> Result<Self, Error> {
+        match Chart::get_installed_chart_by_name(name, namespace) {
+            Ok(chart) => {
+                self.chart = chart;
+            }
+            Err(err) => {
+                return Err(err);
+            }
+        }
+        Ok(self)
```
tiagolobocastro, line 229:
```suggestion
        match Chart::get_installed_chart_by_name(name, namespace) {
            Ok(chart) => {
                self.chart = chart;
                Ok(self)
            }
            Err(err) => {
                Err(err)
            }
        }
```

operators/src/upgrade/helm/client.rs
```diff
@@ -0,0 +1,266 @@
...
+    fn namespace(&self) -> Option<String> {
+        self.namespace.clone()
+    }
+
+    /// Get values.
+    fn values(&self) -> &Vec<PathBuf> {
+        &self.values
+    }
+
+    // Apply arguments for helm command.
```
tiagolobocastro, line 76: comments above non-pub functions can still be doc comments (///)

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

rajaSahil:
bors try

rajaSahil:
bors try

rajaSahil:
bors merge

PR #20, opened 2022-08-22
feat(call-home): add transmitter

niladrih:
Signed-off-by: Niladri Halder <niladri.halder26@gmail.com>

This is a work in progress PR for the Swarm transmitter.

This depends on #18 .

PR #27, opened 2022-09-11
feat(call-home): added ci for call-home -transmitter

datacore-vkumar:
Added Ci for call-home-transmitter components

Note- It is dependent on call-home transmitter PR #20  
Signed-off-by: Vivek Kumar <vivek.kumar@datacore.com>


nix/pkgs/images/default.nix
```diff
@@ -2,15 +2,16 @@
...
   image_suffix = { "release" = ""; "debug" = "-debug"; "coverage" = "-coverage"; };
-  build-extensions-image = { pname, buildType, package, config ? { } }:
+  build-extensions-image = { pname, buildType, package, extraCommands ? '''', contents ? [ ], config ? { } }:
     dockerTools.buildImage {
       tag = extensions.version;
       created = "now";
       name = "mayadata/mayastor-${pname}${image_suffix.${buildType}}";
-      contents = [ package ];
+      contents = [ package ] ++ contents;
+      extraCommands = extraCommands;
```
tiagolobocastro, line 14:
```suggestion
      dockerTools.buildImage {
        inherit extraCommands;
        
        tag = extensions.version;
        created = "now";
        name = "mayadata/mayastor-${pname}${image_suffix.${buildType}}";
        contents = [ package ] ++ contents;
```

nix/pkgs/extensions/default.nix
```diff
@@ -44,6 +44,19 @@ let
         pname = "upgrade-operator";
       };
     };
+    call-home = rec {
```
tiagolobocastro, line 47: should `call-home` be part of `exporter`?

datacore-vkumar: In my opinion, No as call-home has a separate directory and it is not exposing any data through any HTTP endpoint instead it sends the encrypted JSON to a swarm server every 24 hours. So this data can not be read by Prometheus or in any other way by the user.

tiagolobocastro: That's how it currently is (directory structure) but I'm questioning if that's how we really want it to be. 
I thought `exporter` was just a generic name, it's exporting things, and 1 example is `metrics`. As per the nomenclature in the wiki, this is our naming convention: `<classification>-<specific_name>`.
I could see call home here as `exporter.call-home`?

datacore-vkumar: Yes, I got your point but this name was decided during the earlier discussions and it is also there in the system impact statement that's why I used this name.

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

datacore-vkumar:
bors try

datacore-vkumar:
bors merge

PR #28, opened 2022-09-11
feat(call-home): add helm chart components

niladrih:

This PR adds helm chart template, values entries and README.md updates for call-home.

**Note:**
This PR can be merged only after the container image is push into the docker repository.


chart/templates/mayastor/call-home/call-home-deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: call-home
+  template:
+    metadata:
+      labels:
+        app: call-home
+        openebs.io/logging: "true"
```
vharsh, line 17: Do we want supportability tooling to capture this log by default?

chart/values.yaml
```diff
@@ -414,3 +414,20 @@ loki-stack:
               - __meta_kubernetes_pod_uid
               - __meta_kubernetes_pod_container_name
               target_label: __path__
+
+call-home:
```
vharsh, line 418: I need to check if hyphens are considered okay now in template key-strings, iirc this was not considered a good practice and helm has a function to allow templating the values where the keys with hyphens exist.

vharsh (commented):
more changes might be required to making naming in sync

chart/templates/mayastor/call-home/call-home-deployment.yaml
vharsh: more than this, by default the Logging infra will capture & store it's logs by default for the set retention period

niladrih: No, this component does not need logging. Good catch, thanks.

chart/values.yaml
niladrih: RIght! Good catch.

niladrih: hyphens are not usable.

niladrih:
Changed component name to obs-callhome from call-home

chart/templates/mayastor/obs/obs-callhome-deployment.yaml
```diff
@@ -0,0 +1,45 @@
...
+        - name: obs-callhome
+          image: "{{ .Values.image.registry }}/{{ .Values.image.repo }}/mayastor-obs-callhome:{{ .Values.image.tag }}"
+          args:
+            - "-e http://{{ .Release.Name }}-api-rest:8081"
+            - "-n {{ .Release.Namespace }}"
+          env:
+            - name: RUST_LOG
+              value: {{ .Values.obs.callhome.logLevel }}
+            {{- if default .Values.base.logSilenceLevel .Values.obs.callhome.logSilenceLevel }}
+            - name: RUST_LOG_SILENCE
```
tiagolobocastro, line 34: Unless I'm mistaken, I don't think call home processes this env variable

tiagolobocastro (approved):

chart/templates/mayastor/obs/obs-callhome-deployment.yaml
niladrih: Right!

avishnu (approved):
lgtm

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors merge

PR #29, opened 2022-09-12
feat(csi-node): allow csi topology segments and nodeSelector constraint

tiagolobocastro:
feat(csi-node): allow configuring csi topology selector

This is "common" segment that all csi-nodes report as part of their topology.
This allows us to affinitize a pv to a csi-node (any).
When the csi-node is no longer running on a particular node the segment must be
"manually" removed from the kubernetes node labels.

---

feat(csi-node): add nodeSelector to the csi-node deployment

By default, the nodeSelector is disabled, which keeps in the current behaviour.
If enabled, each node where an application pod needs to run should be labelled
with the csi topology segments.


Abhinandan-Purkait (approved):

vharsh (approved):
Lgtm

tiagolobocastro:
bors merge

PR #31, opened 2022-09-14
fix(callhome): changed call-home to callhome

datacore-vkumar:
Changed the directory name of call-home from call-home to callhome

Signed-off-by: Vivek Kumar <vivek.kumar@datacore.com>


datacore-vkumar:
bors try

datacore-vkumar:
This change is not required as we're going to go with 'call-home' for the rust package name.

PR #37, opened 2022-09-20
fix(call-home): fix path join for output_filepath var

niladrih:

This PR fixes the default encrytion dir path for the gpg command. The directory path used in the default constant must not have a trailing slash.

### DIFF
--- a/call-home/src/common/constants.rs
+++ b/call-home/src/common/constants.rs
@@ -1,3 +1,5 @@
+use std::env;
+
 /// PRODUCT is the name of the project for which this call-home component is deployed.
 pub(crate) const PRODUCT: &str = "Mayastor";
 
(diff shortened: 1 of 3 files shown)


niladrih:
bors try

datacore-vkumar (approved):
lgtm

tiagolobocastro:
I don't think this is the right fix, the actual problem is that we're adding the paths using strings incorrectly: https://github.com/openebs/mayastor-extensions/blob/2a5cb7e68c339194c20021706a103ed1f3a66c28/call-home/src/transmitter/encryption.rs#L39
I think the fix is to do this instead:
```rust
let output_filepath = std::path::Path::new(&encryption_dir).join(random_name + ".gpg").to_string_lossy().to_string();
```

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

call-home/src/common/constants.rs
```diff
@@ -7,16 +7,10 @@ pub(crate) const PRODUCT: &str = "Mayastor";
...
 pub(crate) fn get_encryption_dir() -> String {
-    return match std::env::var("ENCRYPTION_DIR") {
-        Ok(input) => {
-            // This is a hack to eliminate a trailing slash.
-            match std::path::Path::new(&input).components().as_path().to_str() {
-                Some(path) => path.to_string(),
-                None => DEFAULT_ENCRYPTION_DIR_PATH.to_string(),
-            }
-        }
+    match std::env::var("ENCRYPTION_DIR") {
```
tiagolobocastro, line 10: nit: consider returning std::path::Path from these dir fn's.
nit: consider removing `get_` prefix from these functions

tiagolobocastro (approved):

call-home/src/common/constants.rs
niladrih: I wanted to return Path, but it said `the trait bound `Path: std::marker::Sized` is not satisfied [E0277] `Path` does not have a constant size known at compile-time`

I'll remote the `get_` I guess.

tiagolobocastro: Ah right, it would need to be a `PathBuf`, nvm

niladrih:
bors merge

call-home/src/main.rs
```diff
@@ -55,9 +55,9 @@ async fn run() -> anyhow::Result<()> {
     let endpoint = args.endpoint;
     let namespace = digest(args.namespace);
 
-    let sleep_duration = get_call_home_frequency();
-    let encryption_dir = get_encryption_dir();
-    let key_filepath = get_key_filepath();
+    let sleep_duration = call_home_frequency();
+    let encryption_dir = encryption_dir();
+    let key_filepath = key_filepath();
```
avishnu, line 60: @niladrih are these variables (encryption_dir and key_filepath) being consumed?

niladrih: Yes, these are being consumed inside the loop further down inside that function.
They are declared there to read input from the environment variables once, and the variable is used in the loop.

2 of the variables are cloned inside the loop because we are using them inside a `tokio::task::spawn_blocking()`. The spawn_blocking() function accepts a closure which can have arguments which live at least as long as the scope of spawn_blocking() call. In this case, that scope is the loop iteration of the spawn_blocking() call. The compiler cannot guarantee that the variables will continue to live for the entire duration of the spawn_blocking() call, as the arguments are `String` type (owned type). By redeclaring them inside the loop iteration, we have guaranteed that even if the outer scope variables are moved, the loop scoped clones will continue to exist at least as long as the iteration exists, and the iteration with will wait for the spawn_blocking() call to return before it exits and moves on to the next iteration.

niladrih: You may notice that the `report` variable does not have this problem, because the report variable will always be generated in each iteration and then moved into the closure. If the report variable was outside the loop, it will be consumed by the first iteration of the loop, and no longer usable, then we'd have to clone it inside the loop scope.

We're not using the functions encryption_dir() and key_filepath() inside the loop because these read from environment variables, and the input values are intended to be consistent throughout the duration of the process. So, input must be read once.

PR #41, opened 2022-09-23
feat(publish-tag): publish the helm chart on tag push

tiagolobocastro:
This is an interim solution for this next release until we sort out what the automated procedure will be like.
We'll keep the the versions and images pointing to the v2.0.0 release, so the chart will not be working from the git repo as is, without overriding the image tags.

When a new tag is pushed (eg: 2.0.0-alpha.2), we publish a helm chart with that semver for the appVersion and for the version. Exception: if the version has already increased, we'll simply bump the version.

A test has been added which shows how this works.

### DIFF
--- a/.github/workflows/release-chart.yaml
+++ b/.github/workflows/release-chart.yaml
@@ -1,18 +1,34 @@
 name: Release Helm Chart
 on:
   push:
-    branches:
-      - develop
-    paths:
-      - 'chart/**'
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+**'
 
 jobs:
   release-chart:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: cachix/install-nix-action@v17
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./scripts/helm/shell.nix
+      - name: Publish locally in the workspace
+        run: |
+          if [ "${{ github.ref_type }}" == "tag" ]; then
+            tag="${{ github.ref_name }}"
+            # Publish the Chart.yaml locally
+            # Note this does not commit the Chart.yaml changes this the branch
+            nix-shell --pure --run "./scripts/helm/publish-chart-yaml.sh --app-tag "$tag"" ./scripts/helm/shell.nix
+          else
+            echo "Only tag pushes are supported!"
+            exit 1
+          fi
       - name: Publish Mayastor Helm chart
         uses: stefanprodan/helm-gh-pages@v1.5.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           charts_dir: .
+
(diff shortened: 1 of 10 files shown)


tiagolobocastro:
bors try

.github/workflows/release-chart.yaml
```diff
@@ -1,18 +1,34 @@
...
   release-chart:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: cachix/install-nix-action@v17
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell -p semver-tool -p yq-go --run "echo"
```
vharsh, line 17: should a `--pure` be added to ensure that only `echo` from nix is used?
I checked that `-p` is not for pure but for `packages`

Jenkinsfile
```diff
@@ -125,6 +126,16 @@ pipeline {
...
           }
         }
+        stage('chart publish test') {
+          when {
+            expression { helm_test == true }
+          }
+          agent { label 'nixos-mayastor' }
+          steps {
+            sh 'printenv'
+            sh 'nix-shell --run "./scripts/helm/test-publish-chart-yaml.sh"'
```
vharsh, line 136: please run it with `--pure` command to ensure sem-ver & jq are picked up only from nix environ(I guess this will be default in most cases as `$PATH` usually has the nix-path in the front

chart/Chart.yaml
```diff
@@ -16,13 +16,13 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.0.0
+version: 2.0.0
```
vharsh, line 19: Has it been decided to match the helm chart versions to appVersions always? On one side, it's good that upgrade code will likely have one less thing to worry about.

vharsh (approved):
Lgtm, it looks using `--pure` might not be a good idea.

[optional] Aside, if removing `nix` can reduce build time, lets get rid of nix and depend upon GitHub actions container to supply us semver, echo, etc.

Jenkinsfile
tiagolobocastro: Yeah it should move to the front of the PATH. Using pure is most useful ensure we're not using any tool/lib which is not defined in the nix, otherwise it might not work on someone else's box.
I'll add pure here and in the actions :+1:

chart/Chart.yaml
tiagolobocastro: For now only for this next release to keep things simpler; initially we thought maybe we can leave the path version differ and pin major and minor, but that might have drawbacks, so we probably need to come up with a longer term strategy going forward - including who's suppose to increment the chart version, each PR that goes in, or a bot?

tiagolobocastro:
> Lgtm, it looks using `--pure` might not be a good idea.
> 
> [optional] Aside, if removing `nix` can reduce build time, lets get rid of nix and depend upon GitHub actions container to supply us semver, echo, etc.

It only takes about 27s to get semver and yq; semver and yq are not present on the ubuntu container, so we'd have to install them as well. I've added a shell.nix to the helm script to tidy it up a bit and clarify the requirements.

tiagolobocastro:
bors merge

PR #47, opened 2022-09-30
feat(ha-cluster-agent): added helm template for ha cluster agent

Ab-hishek:
HA cluster agent is added as a separate new container to the `agent-core` deployment since cluster agents depends on core-agent for its working and hence should have the dynamic behaviour as that of the core-agent. Therefore instead of adding `ha-cluster-agent` as a separate deployment, it is added alongside `core-agent`.

Signed-off-by: Abhishek Agarwal <abhishek.agarwal@mayadata.io>

### DIFF
--- a/chart/templates/mayastor/agents/core/agent-core-and-ha-cluster-deployment.yaml
+++ b/chart/templates/mayastor/agents/core/agent-core-and-ha-cluster-deployment.yaml
@@ -1,18 +1,18 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ .Release.Name }}-agent-core
+  name: {{ .Release.Name }}-agent-core-and-ha-cluster
   labels:
-    app: agent-core
+    app: agent-core-and-ha-cluster
 spec:
   replicas: 1
   selector:
     matchLabels:
-      app: agent-core
+      app: agent-core-and-ha-cluster
   template:
     metadata:
       labels:
-        app: agent-core
+        app: agent-core-and-ha-cluster
         openebs.io/logging: "true"
     spec:
       serviceAccount: {{ .Release.Name }}-service-account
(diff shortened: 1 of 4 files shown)


chart/templates/mayastor/agents/core/agent-core-and-ha-cluster-deployment.yaml
```diff
@@ -1,18 +1,18 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ .Release.Name }}-agent-core
+  name: {{ .Release.Name }}-agent-core-and-ha-cluster
```
tiagolobocastro, line 4: I'm not sure about this, I think in this case we should still let this be the core agent, otherwise as we bring features in we'd end up with a never ending name.
Also we have tests which depend on `app: agent-core` so I don't think we should change this.

chart/templates/mayastor/agents/core/agent-core-and-ha-cluster-deployment.yaml
```diff
@@ -57,3 +57,34 @@ spec:
...
+            limits:
+              cpu: {{ .Values.agents.ha.cluster.resources.limits.cpu | quote }}
+              memory: {{ .Values.agents.ha.cluster.resources.limits.memory | quote }}
+            requests:
+              cpu: {{ .Values.agents.ha.cluster.resources.requests.cpu | quote }}
+              memory: {{ .Values.agents.ha.cluster.resources.requests.memory | quote }}
+          image: "{{ .Values.image.registry }}/{{ .Values.image.repo }}/mayastor-agent-ha-cluster:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          args:
+            - "-g=0.0.0.0:11500"
```
tiagolobocastro, line 71:
```suggestion
            - "--grpc-server-addr=0.0.0.0:50051"
```
nit: use long name of arguments to make it clearer.

chart/templates/mayastor/agents/core/agent-core-and-ha-cluster-deployment.yaml
```diff
@@ -57,3 +57,34 @@ spec:
...
+            - "--store {{ .Release.Name }}-etcd:{{ .Values.etcd.service.port }}"
+            - "--core-grpc=0.0.0.0:50051"
```
tiagolobocastro, line 73:
```suggestion
            "--core-grpc=https://{{ .Release.Name }}-agent-core:50051"
```

chart/values.yaml
```diff
@@ -131,6 +131,19 @@ agents:
...
+      logLevel: info
+      resources:
+        limits:
+          # cpu limits for ha cluster agent
+          cpu: "100m"
+          # memory limits for ha cluster agent
+          memory: "32Mi"
+        requests:
+          # cpu requests for ha cluster agent
+          cpu: "500m"
```
tiagolobocastro, line 144: Is this a typo, should be 100m?

chart/templates/mayastor/agents/core/agent-core-and-ha-cluster-deployment.yaml
Ab-hishek: Ahh! Right!

chart/templates/mayastor/agents/core/agent-core-and-ha-cluster-deployment.yaml
Ab-hishek: Okay.

chart/values.yaml
Ab-hishek: Ahh, yes! Will change it.

chart/templates/mayastor/agents/core/agent-core-and-ha-cluster-deployment.yaml
tiagolobocastro: Although since we're running on the same pod, maybe we can just connect directly however we'd need to add some life/ready probing to make sure we don't see crashes/failures if they start out of order?

abhilashshetty04 (approved):

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

tiagolobocastro:
This should not have been merged. For reference please make sure the changes to helm actually work as the image was not even being built by control-plane yet.

Abhinandan-Purkait:
> This should not have been merged. For reference please make sure the changes to helm actually work as the image was not even being built by control-plane yet.

There are some issues and its not working as expected, will be raising a PR to fix them.

tiagolobocastro:
Just to make it clear the issue is not that an image has bugs, but rather if it's not present, then helm fails to deploy successfully and that breaks e2e meaning we don't get any test results for the day. 
Hopefully this won't be an issue once we have the new "stable" helm being tested on a k8s cluster. It will add more delay to system-test but at least we should always have test results.

PR #48, opened 2022-10-06
build!: set openebs as image registry

GlennBullingham:
BREAKING CHANGE: Build container images using openebs not mayadata as the registry

Signed-off-by: GlennBullingham <glenn.bullingham@datacore.com>


tiagolobocastro:
Think you also need to change the values.yaml file?

PR #55, opened 2022-10-17
fix(call-home): avoid panic when transmission fails

niladrih:

At this point, if the call-home transmission module fails to reach the receiver API server, it retries 3 times and panics.
This PR changes this behavior, and after 3 retries, the error is logged and we don't panic.


Abhinandan-Purkait (approved):

tiagolobocastro (approved):

niladrih:
bors merge

PR #57, opened 2022-10-31
feat(nvme): add ctrl_loss_tmo parameter

tiagolobocastro:
Add support for modifying nvme controller loss timeout. Nvme-tcp is not able to reconnect a device after the max_reconnections has been hit and the device is lost, so we should set a known limit which is sufficiently large to allow for failover. Conversely a value that is too large will make rebooting tricky as the kernel will keep trying..


tiagolobocastro:
Depends on: https://github.com/openebs/mayastor-control-plane/pull/371

Abhinandan-Purkait (approved):

chriswldenyer (approved):
LGTM

tiagolobocastro:
bors merge

PR #70, opened 2022-12-02
feat(upgrade): refactor upgrade controller and kubectl plugin

niladrih:
Upgrade operator changes:
- run controller asynchronously/respond to apply upgrade REST call
- add check to see if the io-engine pods all come to running state after upgrade
- add check to verify if all volumes are unpublished before restarting io-engine
- add loop to retry data-plane upgrade
- add check to verify if io-engine pods are restarted

Kubectl plugin changes:
- update upgrade operator deployment args flags
- add get, list, create, delete, patch permissions for poddisruptionbudget resource

### DIFF
--- a/k8s/plugin/src/constant.rs
+++ b/k8s/plugin/src/constant.rs
@@ -39,7 +39,7 @@ pub(crate) const UPGRADE_OPERATOR_SERVICE: &str = "operator-upgrade";
 pub(crate) const UPGRADE_OPERATOR_SERVICE_PORT: i32 = 8080;
 /// Service internal port constant for upgrade operator.
 pub(crate) const UPGRADE_OPERATOR_INTERNAL_PORT: i32 = 8080;
-/// Upgrade image tag
+/// This is the upgrade-operator container image.
 pub(crate) const UPGRADE_IMAGE: &str = "openebs/mayastor-operator-upgrade:develop";
 /// The service port for upgrade operator.
 pub const UPGRADE_OPERATOR_HTTP_PORT: &str = "http";
--- a/k8s/plugin/src/resources/objects.rs
+++ b/k8s/plugin/src/resources/objects.rs
@@ -5,7 +5,7 @@ use crate::{
         UPGRADE_OPERATOR_HTTP_PORT, UPGRADE_OPERATOR_INTERNAL_PORT, UPGRADE_OPERATOR_SERVICE,
         UPGRADE_OPERATOR_SERVICE_ACCOUNT, UPGRADE_OPERATOR_SERVICE_PORT,
     },
-    upgrade_labels, CliArgs,
+    upgrade_labels,
 };
 
 use k8s_openapi::{
@@ -19,8 +19,6 @@ use k8s_openapi::{
     },
     apimachinery::pkg::{apis::meta::v1::LabelSelector, util::intstr::IntOrString},
 };
-
-use plugin::rest_wrapper::RestClient;
 
 use kube::core::ObjectMeta;
 use maplit::btreemap;
@@ -160,7 +158,7 @@ pub(crate) fn upgrade_operator_cluster_role(
             PolicyRule {
                 api_groups: Some(vec!["rbac.authorization.k8s.io"].into_vec()),
                 resources: Some(vec!["clusterroles"].into_vec()),
-                verbs: vec!["create", "list", "delete", "get", "patch"].into_vec(),
+                verbs: vec!["create", "list", "delete", "get", "patch", "escalate"].into_vec(),
                 ..Default::default()
             },
             PolicyRule {
@@ -190,6 +188,12 @@ pub(crate) fn upgrade_operator_cluster_role(
             PolicyRule {
                 api_groups: Some(vec!["scheduling.k8s.io"].into_vec()),
                 resources: Some(vec!["priorityclasses"].into_vec()),
+                verbs: vec!["create", "list", "delete", "get", "patch"].into_vec(),
+                ..Default::default()
+            },
+            PolicyRule {
+                api_groups: Some(vec!["policy"].into_vec()),
+                resources: Some(vec!["poddisruptionbudgets"].into_vec()),
                 verbs: vec!["create", "list", "delete", "get", "patch"].into_vec(),
                 ..Default::default()
             },
@@ -229,18 +233,21 @@ pub(crate) fn upgrade_operator_cluster_role_binding(
 
 /// Defines the upgrade-operator deployment.
 pub(crate) fn upgrade_operator_deployment(
-    namespace: Option<String>,
+    namespace: &str,
     upgrade_image: String,
     release_name: String,
 ) -> Deployment {
-    let rest_endpoint_arg = format!("--rest-endpoint={}", RestClient::get_or_panic().uri());
-    let namespace_arg = format!("--namespace={}", CliArgs::args().namespace);
-    let chart_name_arg = format!("--chart-name={}", &release_name);
+    let rest_endpoint_clusterip_url = format!("http://{}-api-rest:8081", &release_name);
+    let rest_endpoint_arg = format!("--rest-endpoint={}", rest_endpoint_clusterip_url);
+
+    let namespace_arg = format!("--namespace={}", namespace);
+    let chart_release_name_arg = format!("--release-name={}", &release_name);
+
     Deployment {
         metadata: ObjectMeta {
             labels: Some(upgrade_labels!(UPGRADE_OPERATOR)),
             name: Some(upgrade_group(&release_name, UPGRADE_CONTROLLER_DEPLOYMENT)),
-            namespace: namespace.clone(),
+            namespace: Some(namespace.to_string()),
             ..Default::default()
         },
         spec: Some(DeploymentSpec {
@@ -256,12 +263,15 @@ pub(crate) fn upgrade_operator_deployment(
             template: PodTemplateSpec {
                 metadata: Some(ObjectMeta {
                     labels: Some(btreemap! { LABEL.to_string() => UPGRADE_OPERATOR.to_string()}),
-                    namespace,
                     ..Default::default()
                 }),
                 spec: Some(PodSpec {
                     containers: vec![Container {
-                        args: Some(vec![rest_endpoint_arg, namespace_arg, chart_name_arg]),
+                        args: Some(vec![
+                            rest_endpoint_arg,
+                            namespace_arg,
+                            chart_release_name_arg,
+                        ]),
                         image: Some(upgrade_image),
                         image_pull_policy: Some("Always".to_string()),
                         name: UPGRADE_OPERATOR.to_string(),
--- a/operators/Cargo.toml
+++ b/operators/Cargo.toml
@@ -21,7 +21,8 @@ schemars = "0.8.10"
 serde = "1.0.143"
 serde_json = "1.0.83"
 serde_yaml = "0.9.9"
-tokio = {version="1.20.1", features = ["macros", "rt-multi-thread", "time"]}
+tokio = { version = "1.20.2", features = ["rt-multi-thread", "macros", "time"] }
+actix-web = "4.2.1"
 validator = {version="0.16.0",features=["derive"]}
 tracing = "0.1.36"
 tracing-subscriber = {version = "0.3.15", features = ["env-filter", "std"]}
@@ -33,7 +34,6 @@ once_cell = "1.14.0"
 openapi = {path = "../dependencies/control-plane/openapi"}
 clap = { version = "3.2.16", features = ["cargo", "derive"] }
 url = "2.3.1"
-actix-web = "4.2.1"
 thiserror = "1.0.37"
 humantime = "2.1.0"
 file = "1.1.2"
--- a/operators/src/upgrade/config.rs
+++ b/operators/src/upgrade/config.rs
@@ -8,14 +8,14 @@ use url::Url;
 
 use super::{helm::client::HelmClient, k8s::client::K8sClient};
 
-static CONFIG: OnceCell<UpgradeOperatorConfig> = OnceCell::new();
+static CONFIG: OnceCell<UpgradeConfig> = OnceCell::new();
 
 /// Cli Args to initialize rest-endpoint, namespace and chart.
 #[derive(Parser)]
 #[clap(author, version, about)]
 pub struct CliArgs {
     /// An URL endpoint to the control plane's rest endpoint.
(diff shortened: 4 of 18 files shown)


operators/src/upgrade/common/constants.rs
```diff
@@ -3,6 +3,7 @@ use std::{
     env,
     path::{Path, PathBuf},
 };
+use tracing::info;
```
tiagolobocastro, line 6: not used?

operators/src/upgrade/controller/reconciler.rs
```diff
@@ -416,33 +451,37 @@ async fn reconcile(ua: Arc<UpgradeAction>, ctx: Arc<ControllerContext>) -> Resul
     }
 }
 
+pub(crate) enum Command {
```
tiagolobocastro, line 454: missing doc comment

operators/src/upgrade/controller/reconciler.rs
```diff
@@ -416,33 +451,37 @@ async fn reconcile(ua: Arc<UpgradeAction>, ctx: Arc<ControllerContext>) -> Resul
...
+            .run(reconcile, error_policy, Arc::new(context))
+            .for_each(|res| async move {
+                match res {
+                    Ok(o) => {
+                        trace!(?o);
+                    }
+                    Err(e) => {
+                        trace!(?e);
+                    }
+                }
```
tiagolobocastro, line 482: trace!(?res) ?

operators/src/upgrade/controller/reconciler.rs
```diff
@@ -248,21 +252,52 @@ impl ResourceContext {
...
+        for pod in pods.list(&lp).await? {
+            match pod.status {
+                Some(status) => {
+                    match status.conditions {
+                        Some(conditions) => {
+                            for condition in conditions {
+                                if condition.type_.eq("Ready") && condition.status.eq("True") {
+                                    continue
+                                } else {
+                                    warn!("Couldn't verify the ready condition of io-engine Pod '{}' in namespace '{}' to be true", pod.name_any(), pod.namespace().unwrap());
```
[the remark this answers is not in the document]
niladrih: Added a closure.

operators/src/upgrade/controller/reconciler.rs
```diff
@@ -235,6 +235,10 @@ impl ResourceContext {
                 debug!(?status, "Deleted collection of pods: status");
             }
         }
+
+        // Give kubernetes reconciler enough time to delete the pods.
+        tokio::time::sleep(Duration::from_secs(5)).await;
```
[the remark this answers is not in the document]
niladrih: Refactored this section, added a loop.

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

operators/Cargo.toml
```diff
@@ -21,7 +21,15 @@ schemars = "0.8.10"
...
 serde_json = "1.0.83"
 serde_yaml = "0.9.9"
-tokio = {version="1.20.1", features = ["macros", "rt-multi-thread", "time"]}
+
+# The tokio crate version should always match the one used in actix-web.
+# The upgrade controller is deployed as a tokio task, and the actix-web must be able to
+# use the same runtime for its listener and workers.
+# E.g., the tokio version used to build actix-web 4.2.1 is tokio 1.13.1.
+# Ref: https://github.com/actix/actix-web/blob/web-v4.2.1/actix-web/Cargo.toml#L121
+tokio = { version = "1.13.1", features = ["rt-multi-thread", "macros", "time"] }
```
tiagolobocastro, line 30: hmm this should not be true, only needs to be semver compatible which 1.20 is. In the control-plane we use 1.20 with no issues. What issues did you encounter?

operators/src/upgrade/config.rs
```diff
@@ -8,23 +8,23 @@ use url::Url;
...
 
     /// The namespace we are supposed to operate in.
     #[clap(short, long, default_value = "mayastor")]
     namespace: String,
 
     /// The chart_name we are supposed to operate in.
-    #[clap(short, long, default_value = "mayastor")]
-    chart_name: String,
+    #[clap(short = 'r', long = "helm-release", default_value = "mayastor")]
+    helm_chart_release_name: String,
```
tiagolobocastro, line 27: nit: release_name ?

operators/src/upgrade/controller/reconciler.rs
```diff
@@ -216,15 +221,32 @@ impl ResourceContext {
...
+        for pod in initial_io_engine_pod_list.iter() {
+            match pod.metadata.uid.as_ref() {
+                Some(uid) => {
+                    // Building a set of io-engine Pod uid values. These values may be used to check
+                    // for io-engine Pod restarts.
+                    initial_io_engine_pod_uids.insert(uid);
+                }
+                None => {
+                    // It's ok to panic here because all kubernetes objects should have uids.
+                    panic!("Could not list all io-engine Pods' metadata.uids before restart");
```
tiagolobocastro, line 242: Nit: it's fine but since this functions has Result<_,Error> we could still return an Error though?

operators/src/upgrade/helm/client.rs
```diff
@@ -292,4 +298,16 @@ impl HelmClient {
             String::from_utf8(output.stdout).map_err(|error| Error::Utf8 { source: error })?;
         Ok(output)
     }
+    /*
+       pub(crate) fn show_chart(&mut self) -> ChartData {
```
tiagolobocastro, line 302: can remove this ?

k8s/plugin/src/resources/objects.rs
```diff
@@ -233,9 +237,12 @@ pub(crate) fn upgrade_operator_deployment(
     upgrade_image: String,
     release_name: String,
 ) -> Deployment {
-    let rest_endpoint_arg = format!("--rest-endpoint={}", RestClient::get_or_panic().uri());
-    let namespace_arg = format!("--namespace={}", CliArgs::args().namespace);
-    let chart_name_arg = format!("--chart-name={}", &release_name);
+    let rest_endpoint_clusterip_url = format!("http://{}-api-rest:8081", &release_name);
```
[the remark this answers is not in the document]
niladrih: No, it's not. The uri is being generated in the [build()](https://github.com/openebs/mayastor-extensions/blob/52a6888dd3bf784d1091deaa3dd12760018018e0/k8s/plugin/src/main.rs#L202) method, and only being used for forwarding.

operators/src/upgrade/config.rs
niladrih: Ok!

operators/src/upgrade/controller/reconciler.rs
niladrih: Yes, okay, done!

operators/src/upgrade/helm/client.rs
niladrih: Yes.. ✔️

operators/Cargo.toml
niladrih: That's right. I misread this SO answer and wrote that up --> https://stackoverflow.com/questions/66416174/using-actix-from-a-tokio-app-mixing-actix-webmain-and-tokiomain. Thanks!

niladrih: Tested and verified that this works with 1.20.

niladrih:
bors try

operators/Cargo.toml
```diff
@@ -21,7 +21,14 @@ schemars = "0.8.10"
 serde = "1.0.143"
 serde_json = "1.0.83"
 serde_yaml = "0.9.9"
-tokio = {version="1.20.1", features = ["macros", "rt-multi-thread", "time"]}
+
+# The tokio crate version should be compatible with the one used in actix-web.
+# The upgrade controller is deployed as a tokio task, and the actix-web must be able to
+# use the same runtime for its listener and workers.
+# E.g., the tokio version used to build actix-web 4.2.1 is tokio 1.13.1.
```
tiagolobocastro, line 28: we can remove this comment as this is generally true for all crates with shared deps.
Also the way we set dependencies with the default resolver you're not guaranteed to get this exact version btw, it will be the "best for all"

operators/src/upgrade/controller/reconciler.rs
```diff
@@ -216,15 +221,35 @@ impl ResourceContext {
...
+            match pod.metadata.uid.as_ref() {
+                Some(uid) => {
+                    // Building a set of io-engine Pod uid values. These values may be used to check
+                    // for io-engine Pod restarts.
+                    initial_io_engine_pod_uids.insert(uid);
+                }
+                None => {
+                    // All kubernetes objects should have uids.
+                    error!("Could not list all io-engine Pods' metadata.uids before restart");
+                    return Err(Error::ReconcileError {
```
tiagolobocastro, line 243: Ah what I meant was add a more appropriate error for this situation so that you could match on it and decide to either retry or fail at that top level.

operators/src/upgrade/controller/reconciler.rs
```diff
@@ -235,7 +260,28 @@ impl ResourceContext {
                 debug!(?status, "Deleted collection of pods: status");
             }
         }
-        Ok(Action::await_change())
+        debug!("Deleted io-engine Pods");
+
+        debug!("Waiting for new io-engine Pods to start...");
```
tiagolobocastro, line 265: We're deleting all io-engine pods at once here, which means all volumes will be broken. Should we check that there are no published volumes?

operators/src/upgrade/helm/client.rs
```diff
@@ -6,6 +6,7 @@ use std::{
 };
 
 use crate::upgrade::common::{constants::chart_dir_path, error::Error};
+//use crate::upgrade::helm::chart_data::ChartData;
```
tiagolobocastro, line 9: remove this line

operators/src/upgrade/k8s/crd/v0.rs
```diff
@@ -80,15 +80,15 @@ impl UpgradeActionSpec {
 /// 'UpgradePhase' defines the status of each components.
 #[derive(Deserialize, Serialize, Clone, Debug, Eq, PartialEq, JsonSchema)]
 pub enum UpgradePhase {
-    /// Components in Waiting phase.
+    // Components in Waiting phase.
```
tiagolobocastro, line 83: Why did you remove the doc comment?

niladrih:
> These are the arguments for the operator itself right?
I wonder if we can define the operator via helm (and then the api-rest and what not is filled in via the template) and then the plugin would simply "apply" that.
Maybe this would solve some "issues", example: we're hardcoding RUST_LOG to info and image_pull_policy to Always, might be nice if we could tweak those via helm the same way we do for the other components.

One alternative could be to do something like... the user adds `--dry-run -o yaml` to the install command to get the yaml manifests. Then he can edit and apply it with configuration that he has validated.

tiagolobocastro:
> > These are the arguments for the operator itself right?
> > I wonder if we can define the operator via helm (and then the api-rest and what not is filled in via the template) and then the plugin would simply "apply" that.
> > Maybe this would solve some "issues", example: we're hardcoding RUST_LOG to info and image_pull_policy to Always, might be nice if we could tweak those via helm the same way we do for the other components.
> 
> One alternative could be to do something like... the user adds `--dry-run -o yaml` to the install command to get the yaml manifests. Then he can edit and apply it with configuration that he has validated.

We could generate the yaml at build time, and then the operator simply has to apply it.
Maybe we an discuss this as a future enhancement (not for this PR)

operators/src/upgrade/k8s/crd/v0.rs
niladrih: Because, the kube-rs library uses the doc comment as input for the 'description' field in the CRD openapi spec json/yaml. That description is not openapi3 compliant, and leads to an invalid openapi3 spec.

niladrih:
bors try

operators/src/upgrade/k8s/crd/v0.rs
tiagolobocastro: oh I see.. what would it take to make it compliant, have description would actually be good to have?

k8s/plugin/src/constant.rs
```diff
@@ -40,7 +40,7 @@ pub(crate) const UPGRADE_OPERATOR_SERVICE_PORT: i32 = 8080;
 /// Service internal port constant for upgrade operator.
 pub(crate) const UPGRADE_OPERATOR_INTERNAL_PORT: i32 = 8080;
 /// Upgrade image tag
-pub(crate) const UPGRADE_IMAGE: &str = "openebs/mayastor-operator-upgrade:develop";
+pub(crate) const UPGRADE_IMAGE: &str = "niladrih/mayastor-operator-upgrade:final3";
```
tiagolobocastro, line 43: please revert this: probably better to add a cli option or something like that to avoid merging the wrong thing

operators/src/upgrade/controller/utils.rs
```diff
@@ -0,0 +1,95 @@
...
+            .get_volumes(max_entries, None, starting_token)
+            .await
+        {
+            Ok(vols) => {
+                let v = vols.into_body();
+                volumes.extend(v.entries);
+                starting_token = v.next_token;
+            }
+            Err(error) => {
+                panic!("Failed to list volumes. error: {}", error);
```
tiagolobocastro, line 82: why is this a panic, this should be handled as errors via rest may occur.
This fn should return Result.

k8s/plugin/src/resources/objects.rs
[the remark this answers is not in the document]
niladrih: ImagePullPolicy, image registry, repository and tag has to be customizable. I'll take up this change in a separate PR.

k8s/plugin/src/constant.rs
niladrih: Oops

niladrih: Fixed it, I'm not adding a cli flag for this at the moment.

operators/src/upgrade/controller/utils.rs
niladrih: Added an error return for this.

niladrih:
bors try

niladrih:
bors try

tiagolobocastro (approved):

niladrih:
bors merge

Abhinandan-Purkait (approved):

niladrih:
bors merge

PR #74, opened 2022-12-07
chore(chart): add labels and update container registry for CSI containers

niladrih:
Changes:
- Add a label with key `helm-release` to Mayastor components whose value will be the release name of the helm chart release.
- Replace k8s.gcr.io container registry for registry.k8s.io.
- Use {{ .Chart.Name }} instead of 'mayastor'.


niladrih:
bors try

chart/templates/mayastor/priority-class/priority-class.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
   name: {{ .Release.Name }}-priority-class
+  labels:
+    helm-release: {{ .Release.Name }}
```
tiagolobocastro, line 6: since you're here, can you replace mayastor on line 9 with {{ .Release.Name }} ?

chart/templates/storageclass.yaml
```diff
@@ -2,6 +2,8 @@ kind: StorageClass
 apiVersion: storage.k8s.io/v1
 metadata:
   name: {{ .Release.Name }}-single-replica
+  labels:
```
tiagolobocastro, line 5: do we need it for the sc?

chart/values.yaml
```diff
@@ -183,7 +183,7 @@ apis:
 csi:
   image:
     # image registry to pull all CSI Sidecar images
-    registry: k8s.gcr.io
+    registry: registry.k8s.io
```
tiagolobocastro, line 186: 👍

tiagolobocastro (approved):

chart/templates/mayastor/priority-class/priority-class.yaml
niladrih: Why though? That 'mayastor' refers to the mayastor project, as in the helm chart itself.... like https://github.com/openebs/mayastor-extensions/blob/develop/chart/Chart.yaml#L2. I don't think this refers to a specific release.

chart/templates/storageclass.yaml
niladrih: Not really. Just putting it on all of the mayastor components.

niladrih:
These changes are included in #75 .

PR #84, opened 2022-12-26
chore: cherry-pick old release branch actions into develop

tiagolobocastro:
feat(publish-tag): cherry-pick publish the helm chart on tag push

This is an interim solution for this next release until we sort out what the automated
procedure will be like.
We'll keep the the versions and images pointing to the v2.0.0 release, so the chart will not
be working from the git repo as is, without overriding the image tags.

When a new tag is pushed (eg: 2.0.0-alpha.2), we publish a helm chart with that semver for
the appVersion and for the version. Exception: if the version has already increased, we'll
simply bump the version.

A test has been added which shows how this works.

---

chore(ci): add action to fixup the release branch

When a branch is created we must update the Chart.yaml.
This action will create a PR with the necessary changes, if required.

---

style: run yq on values.yaml

Also ignore blank line diffs from now on.


Jenkinsfile
```diff
@@ -41,7 +41,7 @@ def mainBranches() {
 // TODO: Use multiple choices
 run_linter = true
 rust_test = true
-bdd_test = true
```
avishnu, line 44: Is this not needed?

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

pchandra19 (approved):

Jenkinsfile
tiagolobocastro: No, copy&paste from controller repo.

tiagolobocastro:
bors merge

PR #93, opened 2023-01-16
feat(upgrade-plugin) : refactor plugin code

sinhaashish:
Signed-off-by: sinhaashish <ashi.sinha.87@gmail.com>

### DIFF
--- a/k8s/plugin/src/resources/upgrade.rs
+++ b/k8s/plugin/src/resources/upgrade.rs
@@ -19,13 +19,22 @@ use kube::{
     core::ObjectList,
     Client,
 };
-use std::{path::PathBuf, time::Duration};
+use std::path::PathBuf;
 
 /// The types of resources that support the upgrade operator.
 #[derive(clap::Subcommand, Debug)]
 pub enum UpgradeOperator {
     /// Install, Uninstall upgrade resources.
     UpgradeOperator,
+}
+
+#[derive(clap::Subcommand, Debug)]
+/// Actions to be performed.
+pub enum Actions {
+    /// Action to create object.
+    Create,
+    /// Action to delete object.
+    Delete,
 }
 
 /// K8s resources needed for upgrade operator.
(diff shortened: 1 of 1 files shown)


sinhaashish:
bors try

niladrih (approved):
LGTM!
nit: In the log messages, you could remove the spaces before the ':'.
nit: In the log messages, you could use the k8s object names as they are, e.g. ServiceAccount or serviceaccount instead of service account or Service Account, and ClusterRole or ClusterRoleBinding....

k8s/plugin/src/resources/upgrade.rs
```diff
@@ -54,57 +61,101 @@ impl UpgradeResources {
...
+                        Actions::Create => {
                             let ns = Some(ns.to_string());
                             let service_account = objects::upgrade_operator_service_account(
                                 ns,
-                                uo.release_name.clone(),
+                                self.release_name.clone(),
                             );
                             let pp = PostParams::default();
-                            match uo.service_account.create(&pp, &service_account).await {
+                            match self.service_account.create(&pp, &service_account).await {
```
tiagolobocastro, line 77: what's the difference between sa (line 67), and service_account (72) ?
I'm trying to see if we can do the match action at the root and not do the sa.iter.count.

k8s/plugin/src/resources/upgrade.rs
```diff
@@ -54,57 +61,101 @@ impl UpgradeResources {
...
+                } else {
+                    match action {
+                        Actions::Create => {
                             println!(
                                 "Service Account : {} in namespace : {} already exist.",
-                                sa.items[0].metadata.name.as_ref().unwrap(),
-                                sa.items[0].metadata.namespace.as_ref().unwrap(),
+                                sa.items.get(0).unwrap().metadata.name.as_ref().unwrap(),
+                                sa.items
+                                    .get(0)
```
tiagolobocastro, line 103: you can tidy is by changing the if above to if let Some... and then you get the inner from the Some.

k8s/plugin/src/resources/upgrade.rs
```diff
@@ -54,57 +61,101 @@ impl UpgradeResources {
...
+            }
+        }
+        Ok(())
+    }
+
+    /// Create/Delete cluster role
+    pub async fn cluster_role_actions(&self, ns: &str, action: Actions) -> Result<(), ()> {
+        match self.get_cluster_role().await {
+            Ok(cr) => {
+                if cr.iter().count() == 0 {
```
tiagolobocastro, line 149: same comment here as for the service_account.

k8s/plugin/src/resources/upgrade.rs
```diff
@@ -270,148 +372,156 @@ impl UpgradeResources {
...
+    }
 
+    /// Install the upgrade resources
+    pub async fn install(ns: &str) {
+        match UpgradeResources::new(ns).await {
+            Ok(uo) => {
+                // Create Service Account
+                uo.service_account_actions(ns, Actions::Create)
+                    .await
+                    .map_err(|err| println!("{:?}", err))
```
tiagolobocastro, line 470: we print the errors but ignore them, is this intended?

k8s/plugin/src/resources/upgrade.rs
sinhaashish: Line 67 `sa` is the list of Service account.   Basically this ` pub async fn get_service_account(&self) -> Result<ObjectList<ServiceAccount>, Error>` . I am doing an `sa.iter().count()` as its return an object list . We can return `ServiceAccount` only instead of the list, but that IMHO wont be good as we will be restricting to just a ServiceAccount at index 0. Returning the list is good as this code can be reused whenever we need to have a list of service accounts.

tiagolobocastro: I get that, what I'm asking what is the different between those two variables, `sa` and `service_account`?

sinhaashish: changes the code where in i am using `get_apt` rather than list.

k8s/plugin/src/resources/upgrade.rs
sinhaashish: yes, as we need to continue even if the resources are present.

tiagolobocastro: Should we specifically ignore AlreadyExists error then? Seems we'll just ignore any error.

sinhaashish: Resolved it. PTAL

sinhaashish:
bors try

k8s/plugin/src/resources/errors.rs
```diff
@@ -0,0 +1,58 @@
+use snafu::Snafu;
+// use thiserror::Error;
```
tiagolobocastro, line 2: hmm I see thiserror which actually probably pairs up better with anyhow, any reason to use snafu instead?

k8s/plugin/src/resources/upgrade.rs
```diff
@@ -54,187 +62,371 @@ impl UpgradeResources {
...
-                                    std::process::exit(1);
-                                }
-                            }
-                        } else {
-                            println!(
-                                "Service Account : {} in namespace : {} already exist.",
-                                sa.items[0].metadata.name.as_ref().unwrap(),
-                                sa.items[0].metadata.namespace.as_ref().unwrap(),
-                            );
+    /// Create/Delete ServiceAction
```
tiagolobocastro, line 65: not for this PR, but please move all this code to a library as this is becoming too unwieldy to be in the plugin, instead should be imported from a library.

k8s/plugin/src/resources/upgrade.rs
```diff
@@ -54,187 +62,371 @@ impl UpgradeResources {
...
+                            &upgrade_group(&self.release_name, UPGRADE_OPERATOR_SERVICE_ACCOUNT),
+                            &DeleteParams::default(),
+                        )
+                        .await
+                    {
+                        Ok(_) => {
+                            println!("ServiceAccount deleted");
+                        }
+                        Err(e) => {
+                            return Err(UpgradeError::ServiceAccountDelete { error: e });
```
tiagolobocastro, line 100: nit: if you rename e to error, the can simply have `Err(UpgradeError::ServiceAccountDelete { error })`

k8s/plugin/src/resources/upgrade.rs
```diff
@@ -54,187 +62,371 @@ impl UpgradeResources {
...
+                Actions::Create => {
+                    let ns = Some(ns.to_string());
+                    let role_binding = objects::upgrade_operator_cluster_role_binding(
+                        ns,
+                        self.release_name.clone(),
+                    );
+                    let pp = PostParams::default();
+                    match self.cluster_role_binding.create(&pp, &role_binding).await {
+                        Ok(crb) => {
+                            println!("ClusterRoleBinding: {} created", crb.metadata.name.unwrap(),);
```
tiagolobocastro, line 249:
```suggestion
                            println!("ClusterRoleBinding: {} created", crb.metadata.name.unwrap());
```

k8s/plugin/src/resources/upgrade.rs
```diff
@@ -54,187 +62,371 @@ impl UpgradeResources {
...
+                UPGRADE_CONTROLLER_DEPLOYMENT,
+            ))
+            .await?
+        {
+            match action {
+                Actions::Create => {
+                    println!(
+                        "Deployment: {} in namespace: {} already exist",
+                        deployment.metadata.name.as_ref().unwrap(),
+                        deployment.metadata.namespace.as_ref().unwrap(),
```
tiagolobocastro, line 279:
```suggestion
                        deployment.metadata.namespace.as_ref().unwrap()
```

k8s/plugin/src/resources/upgrade.rs
```diff
@@ -54,187 +62,371 @@ impl UpgradeResources {
...
+        }
+        Ok(())
+    }
+
+    /// Install the upgrade resources
+    pub async fn install(ns: &str) {
+        match UpgradeResources::new(ns).await {
+            Ok(uo) => {
+                // Create Service Account
+                if let Err(UpgradeError::ServiceAccountCreate { .. }) =
```
tiagolobocastro, line 398: please use match statement instead of having all this `if's`

k8s/plugin/src/resources/upgrade.rs
```diff
@@ -28,6 +29,13 @@ pub enum UpgradeOperator {
     UpgradeOperator,
 }
 
+/// Actions to be performed.
+#[derive(clap::Subcommand, Debug)]
+pub enum Actions {
+    Create,
```
tiagolobocastro, line 35: this needs a doc comment?

tiagolobocastro (approved):

k8s/plugin/src/resources/errors.rs
sinhaashish: removed the errors file and did a simple match on errors

k8s/plugin/src/resources/upgrade.rs
tiagolobocastro: I still don't see the doc comment here.

k8s/plugin/Cargo.toml
```diff
@@ -37,6 +37,7 @@ async-trait = "0.1.51"
 serde = "1.0.140"
 serde_json = "1.0.82"
 url = "2.2.2"
+snafu = "0.7.1"
```
tiagolobocastro, line 40: guess no longer needed?

tiagolobocastro (approved):

tiagolobocastro (changes requested):
Why are you downgrading the control-plane submodule?

tiagolobocastro (approved):

sinhaashish:
bors merge

PR #99, opened 2023-01-17
chore: make rest api node port service configurable

Abhinandan-Purkait:
Signed-off-by: Abhinandan Purkait <purkaitabhinandan@gmail.com>

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -178,6 +178,16 @@ apis:
         cpu: "50m"
         # -- Memory requests for rest
         memory: "32Mi"
+    # Rest service parameters define how the rest service is exposed
+    service:
+      # -- Rest K8s service type
+      type: ClusterIP
+      # Ports from where rest endpoints are accessible from outside the cluster, only valid if type is NodePort
+      nodePorts:
+        # NodePort associated with http port
+        http: 30011
+        # NodePort associated with https port
+        https: 30010
 
 csi:
   image:
--- a/chart/README.md
+++ b/chart/README.md
@@ -73,6 +73,7 @@ $ helm install my-release openebs/mayastor
 | apis.&ZeroWidthSpace;rest.&ZeroWidthSpace;resources.&ZeroWidthSpace;limits.&ZeroWidthSpace;memory | Memory limits for rest | `"64Mi"` |
 | apis.&ZeroWidthSpace;rest.&ZeroWidthSpace;resources.&ZeroWidthSpace;requests.&ZeroWidthSpace;cpu | Cpu requests for rest | `"50m"` |
 | apis.&ZeroWidthSpace;rest.&ZeroWidthSpace;resources.&ZeroWidthSpace;requests.&ZeroWidthSpace;memory | Memory requests for rest | `"32Mi"` |
+| apis.&ZeroWidthSpace;rest.&ZeroWidthSpace;service.&ZeroWidthSpace;type | Rest K8s service type | `"ClusterIP"` |
 | base.&ZeroWidthSpace;cache_poll_period | Cache timeout for core agent & diskpool deployment | `"30s"` |
 | base.&ZeroWidthSpace;default_req_timeout | Request timeout for rest & core agents | `"5s"` |
 | base.&ZeroWidthSpace;imagePullSecrets.&ZeroWidthSpace;enabled | Enable imagePullSecrets for pulling our container images | `false` |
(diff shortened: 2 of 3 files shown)


chart/values.yaml
```diff
@@ -178,6 +178,17 @@ apis:
         cpu: "50m"
         # -- Memory requests for rest
         memory: "32Mi"
+    # Rest service parameters defines how the rest service is exposed
+    service:
+      # K8s service type
+      type: ClusterIP
+
+      # Ports from where rest endpoints are accessible from outside cluster, only valid if type is NodePort
+      nodePorts:
```
blaisedias, line 187: Should these lines be commented out by default?

avishnu: Yes IMO.

chart/values.yaml
```diff
@@ -178,6 +178,17 @@ apis:
         cpu: "50m"
         # -- Memory requests for rest
         memory: "32Mi"
+    # Rest service parameters defines how the rest service is exposed
+    service:
+      # K8s service type
+      type: ClusterIP
+
```
avishnu, line 185: Blank line can be removed.

chart/values.yaml
```diff
@@ -178,6 +178,17 @@ apis:
         cpu: "50m"
         # -- Memory requests for rest
         memory: "32Mi"
+    # Rest service parameters defines how the rest service is exposed
```
avishnu, line 181: typo: define

chart/values.yaml
```diff
@@ -178,6 +178,17 @@ apis:
...
+    service:
+      # K8s service type
+      type: ClusterIP
+
+      # Ports from where rest endpoints are accessible from outside cluster, only valid if type is NodePort
```
avishnu, line 186: outside the cluster

avishnu:
Thanks @Abhinandan-Purkait . Do we need similar changes for Loki and etcd?

chart/values.yaml
Abhinandan-Purkait: No, these values are also configurable. It will not work if we comment them.

chart/values.yaml
Abhinandan-Purkait: Done.

Abhinandan-Purkait:
> Thanks @Abhinandan-Purkait . Do we need similar changes for Loki and etcd?

For loki and etcd its already there.

chart/values.yaml
Abhinandan-Purkait: Done.

chart/values.yaml
avishnu: Seeing similar configuration for Loki and etcd as well. Even though, by default, service type is clusterIp, there is a nodePort value as well.

Abhinandan-Purkait: No, we have an if conditional in the template that would add/remove this based on the types.
This in similar to what other chart providers follow.:- ref etcd https://github.com/bitnami/charts/blob/610f860fd8197f6f1dc1906511b56a96df42dac9/bitnami/etcd/values.yaml#L517

avishnu: Having said that, we can still comment out these lines by default, specifying in the comment that "Uncomment below lines if serviceType is nodePort" - @Abhinandan-Purkait

Abhinandan-Purkait: Okay. Shall I comment them for `etcd` and `loki-stack` as well, so that its similar throughout.

tiagolobocastro: I don't think we should comment them out. They're there as the default in case node port is used.

avishnu: As discussed, looks good, no need to comment.

avishnu (approved):
lgtm

chart/values.yaml
```diff
@@ -178,6 +178,16 @@ apis:
...
         # -- Memory requests for rest
         memory: "32Mi"
+    # Rest service parameters define how the rest service is exposed
+    service:
+      # -- Rest K8s service type
+      type: ClusterIP
+      # Ports from where rest endpoints are accessible from outside the cluster, only valid if type is NodePort
+      nodePorts:
+        # NodePort associated with http port
+        httpPort: 30011
```
tiagolobocastro, line 188:
```suggestion
        http: 30011
```

chart/values.yaml
```diff
@@ -178,6 +178,16 @@ apis:
...
+        # NodePort associated with https port
+        httpsPort: 30010
```
tiagolobocastro, line 190:
```suggestion
        https: 30010
```

tiagolobocastro (approved):

chart/values.yaml

chart/values.yaml

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #103, opened 2023-01-31
feat(upgrade): add labels to helm chart

niladrih:
Changes:
- add the helm-release label to the storageclasses for loki and etcd, jaeger crd and cd
- add the helm-chart-version label to all helm chart templates


niladrih:
bors try

pchandra19 (approved):

tiagolobocastro (commented):
lgtm but I wonder if we should qualify those labels, eg: openebs.io/release?

niladrih:
> lgtm but I wonder if we should qualify those labels, eg: openebs.io/release?

I thought the same for this. Awesome!

chart/templates/_helpers.tpl
```diff
@@ -107,3 +107,12 @@ Usage:
...
 {{- end }}
 {{- end }}
+
+{{/*
+Adds the project name
+Usage:
+{{ include "label_prefix" . }}/release: {{ .Release.Name }}
+*/}}
+{{- define "label_prefix" -}}
+    {{ print "openebs.io" }}
```
tiagolobocastro, line 117: should this be coming from a values.yaml var?

niladrih: Values yaml parameters are user-configurable. This is essentially a part of the template and not intended to be altered. Altering this may interfere with upgrade.

tiagolobocastro: Makes sense, then let's read it from a file. Perhaps other things in the values should also come from a file.

niladrih: I don't think helm template functions allow filesystem access. Any idea how I could approach this one?

niladrih: Added a yaml with 'openebs.io' as one of the constants. Good one!

tiagolobocastro: Awesome :+1:

tiagolobocastro (approved):

chart/templates/storageclass.yaml
```diff
@@ -2,8 +2,6 @@ kind: StorageClass
 apiVersion: storage.k8s.io/v1
 metadata:
   name: {{ .Release.Name }}-single-replica
-  labels:
```
avishnu, line 5: Why are we removing this label?

niladrih: We don't need to pick out the default storageclass by label, or grab the release name from it (we're using the rest api Pod's label for that). Also the storageclass plays no role in the upgrade, so we need not verify it's version. We might not need this label any more.

niladrih:
bors merge

PR #105, opened 2023-02-03
chore: adding access on get and update verbs of crd

abhilashshetty04:
Currently in DSP operator init we check if crd is diskpool crd is present on cluster. If no, then only we create crd defined in the image. In case crd is already present, We dont update the crd with new changes. This PR modifies access to crd resource. Adding get and update verb to the cluster role mayastor-cluster-role

Signed-off-by: Abhilash Shetty <abhilash.shetty@datacore.com>


niladrih (approved):

tiagolobocastro (approved):

abhilashshetty04:
bors merge

PR #106, opened 2023-02-06
fix(upgrade): remove labels carrying the chart version from LabelSelectors for Services, Deployments, DaemonSets, StatefulSets

niladrih:
**Why is this change required?**

v1.LabelSelectors cannot be patched during helm upgrade, hence it should only carry labels which don't change across helm chart versions.


avishnu (approved):
seems fine.

tiagolobocastro (approved):

niladrih:
bors merge

PR #113, opened 2023-02-15
add promt for single replica volume

sinhaashish:
Add a note for the hot upgrade : `Note : During upgrade if you create single replica volumes then you may loose them`

Add `upgrade_operator_endpoint` to upgrade and upgrade status command  only.

### DIFF
--- a/console-logger/src/console.rs
+++ b/console-logger/src/console.rs
@@ -1,0 +1,15 @@
+use nu_ansi_term::Color::{Cyan, Red};
+
+/// Print info on console.
+pub fn info(message: &str) {
+    println!("{}", Cyan.bold().italic().paint(message));
+}
+
+/// Print warning on console.
+pub fn warn(data: &str, message: &str) {
+    println!(
+        "{} {} ",
+        Red.bold().italic().paint(message),
+        Cyan.bold().italic().paint(data)
+    );
+}
--- a/k8s/plugin/Cargo.toml
+++ b/k8s/plugin/Cargo.toml
@@ -37,7 +37,7 @@ async-trait = "0.1.64"
 serde = "1.0.152"
 serde_json = "1.0.93"
 url = "2.3.1"
-
+nu-ansi-term = "0.46.0"
 # Tracing
 tracing = "0.1.37"
 opentelemetry = { version = "0.18.0", features = ["rt-tokio-current-thread"] }
--- a/k8s/plugin/Cargo.toml
+++ b/k8s/plugin/Cargo.toml
@@ -19,6 +19,7 @@ tls = [ "openapi/tower-client-tls", "rest-plugin/tls" ]
 [dependencies]
 openapi = { path = "../../dependencies/control-plane/openapi", default-features = false, features = [ "tower-trace" ] }
 utils = { path = "../../dependencies/control-plane/utils/utils-lib" }
+console-logger = { path = "../../console-logger" }
 rest-plugin = { path = "../../dependencies/control-plane/control-plane/plugin", default-features = false }
 kube = { version = "0.78.0", features = [ "derive" ] }
 supportability = { path = "../supportability" }
@@ -37,7 +38,6 @@ async-trait = "0.1.64"
 serde = "1.0.152"
 serde_json = "1.0.93"
 url = "2.3.1"
-nu-ansi-term = "0.46.0"
 # Tracing
 tracing = "0.1.37"
 opentelemetry = { version = "0.18.0", features = ["rt-tokio-current-thread"] }
--- a/k8s/plugin/src/main.rs
+++ b/k8s/plugin/src/main.rs
@@ -13,12 +13,13 @@ use plugin::{
 use std::{env, path::PathBuf};
 mod resources;
 use crate::resources::GetResourcesK8s;
-use http::Uri;
 use resources::{
     upgrade::{UpgradeOperator, UpgradeResources},
     Operations,
 };
 pub mod constant;
+use nu_ansi_term::Color::Cyan;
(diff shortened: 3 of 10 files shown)


k8s/plugin/src/main.rs
```diff
@@ -171,14 +166,16 @@ async fn execute(cli_args: CliArgs) {
...
-                UpgradeResources::apply(
-                    cli_args.upgrade_operator_endpoint,
-                    &cli_args.namespace,
-                    cli_args.kube_config_path,
-                    cli_args.timeout,
-                )
-                .await;
+            Operations::Upgrade(resources) => {
+                println!("{}",
+                Cyan.bold().italic().paint("Note : During upgrade if you create single replica volumes then you may loose them"));
```
niladrih, line 171:
```suggestion
                Cyan.bold().italic().paint("Note: Volumes which make use of a single volume replica instance will be unavailable for some time during upgrade. It is recommended that you do not create new volumes which make use of only one volume replica."));
```

niladrih (commented):
I don't see the code for a prompt. This looks like a message.

k8s/plugin/Cargo.toml
```diff
@@ -37,7 +37,7 @@ async-trait = "0.1.64"
 serde = "1.0.152"
 serde_json = "1.0.93"
 url = "2.3.1"
-
+ansi_term = "0.12.1"
```
tiagolobocastro, line 40: should we use `nu-ansi-term` instead since we already depend on it for tracing to avoid using 2 different crates for the same thing?

k8s/plugin/src/main.rs
tiagolobocastro: Should this be behind an logging interface?
warn("Note: Volumes which make use of a single volume replica instance will be unavailable for some time during upgrade. It is recommended that you do not create new volumes which make use of only one volume replica.");

On another note if these messages are important for the user, maybe we should keep them somewhere common? Would also useful for localization.

sinhaashish:
I tried `tracing::warn!("<<>>");` but this is only printed when we want the trace. The requirement here is to add this prompt as default , so that the user when the run `kubectl mayastor upgrade` knows about the warning.

sinhaashish:
bors try

tiagolobocastro:
> I tried `tracing::warn!("<<>>")` but this is only printed when we want the trace. The requirement here is to add this prompt as default , so that the user when the run `kubectl mayastor upgrade` knows about the warning.

Oh I mean that maybe we should add our own, where we can customise how we show messages for the users, for example if we want to ask the user a yes/no question.

console/src/console.rs
```diff
@@ -0,0 +1,55 @@
+use nu_ansi_term::Color::{Cyan, Red};
+
+/// Console print to users console
+pub trait Console {
+    fn print(&self);
+}
+
+/// Info prints information for users
```
tiagolobocastro, line 8: Seems a bit clunky to have to specify a struct in order in order to print these.
Why not simply expose different methods, such as the usual log crate, eg: info, warn, error?
You can also multiple args to the fn if necessary, eg: your data and message.

console/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "console"
```
tiagolobocastro, line 2: nit: I would suggest a more suggestive name, eg: `console-logger`

console/src/console.rs
sinhaashish: modified

console-logger/src/console.rs
```diff
@@ -0,0 +1,15 @@
+use nu_ansi_term::Color::{Cyan, Red};
+
+/// Print info on console.
+pub fn info(message: String) {
+    println!("{}", Cyan.bold().italic().paint(message));
+}
+
+/// Print warning on console.
+pub fn warn(data: String, message: String) {
```
tiagolobocastro, line 9: can take &str or AsRec<str> instead

k8s/plugin/src/main.rs
```diff
@@ -171,14 +167,15 @@ async fn execute(cli_args: CliArgs) {
...
-                    cli_args.upgrade_operator_endpoint,
-                    &cli_args.namespace,
-                    cli_args.kube_config_path,
-                    cli_args.timeout,
-                )
-                .await;
+            Operations::Upgrade(resources) => {
+                console_logger::info(user_prompt::UPGRADE_WARNING.to_string());
+                resources
+                    .implement(
```
tiagolobocastro, line 173: I don't understand the details here, but IMHO apply seemed better since under the hood is also apply.

k8s/plugin/Cargo.toml
```diff
@@ -19,6 +19,7 @@ tls = [ "openapi/tower-client-tls", "rest-plugin/tls" ]
 [dependencies]
 openapi = { path = "../../dependencies/control-plane/openapi", default-features = false, features = [ "tower-trace" ] }
 utils = { path = "../../dependencies/control-plane/utils/utils-lib" }
+console-logger = { path = "../../console-logger"}
```
tiagolobocastro, line 22:
```suggestion
console-logger = { path = "../../console-logger" }
```

tiagolobocastro (approved):

k8s/plugin/src/main.rs
sinhaashish: changed

sinhaashish:
bors try

sinhaashish:
bors try

sinhaashish:
bors merge

sinhaashish:
bors merge

PR #118, opened 2023-02-25
remove clusterIP: none to some chart services to improve resilience

ryandellolio:
**Observation**
When clusterIP is explicitly set to none in the agent core service chart (most importantly) and the metrics service (less critical), I have observed lack of network connectivity the will ultimately cause the mayastor cluster to fail.

**Recommendation**
Make this chart more resilient for clusters where these services cannot run headless by removing the `clusterIP: None` directive to allow the cluster to apply its default behavior

**Discussion questions**
- Is there any downside to this approach?  I don't think setting a cluster IP on the service will hurt anything, but not 100% confident. This was the fix for me at least
- Is this unique to bare metal/talos cluster or the chosen CNI? Don't want to solve for an edge case
- This could also be set in `values.yaml` as a "better" more configurable fix, with headless services as an option instead of doing this directly in the chart.  Let me know if you want me to take that approach, but I think that's dependent on the answer to questions #1 and #2

**References**
Thanks to @avishnu for the conversation on this [relevant slack thread](https://kubernetes.slack.com/archives/CUAKPFU78/p1677125940176639?thread_ts=1677043218.818379&cid=CUAKPFU78) where this PR was originally proposed


Abhinandan-Purkait:
Hey, @ryandellolio  Can you please rebase the branch with origin/develop and sign your commit?

ryandellolio:
Yes will get to this tomorrow

niladrih (approved):

niladrih (changes requested):
You'd have to DCO sign your commit @ryandellolio. This command should work, if you're using git:
```
git commit --amend -s --no-edit
```

niladrih (approved):

ryandellolio:
Per my conversation with @niladrih on Slack, this should now be all set.  Thanks for the guidance. Let me know if you need anything else.

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
bors merge

PR #123, opened 2023-03-06
fix(call-home): set call-home product_version and header version to 2.0.1

niladrih:
This sets the version in the request body of the transmission module to 2.0.1. It does not use v2.0.1, as previously the call-home server-client codebase has been tested to work with versions similar to rust crate versions. Using a 'v' in the version requires testing.


call-home/src/main.rs
```diff
@@ -50,7 +50,7 @@ async fn main() {
 
 async fn run() -> anyhow::Result<()> {
     let args = CliArgs::args();
-    let version = clap::crate_version!().to_string();
+    let version = "2.0.1".to_string();
```
avishnu, line 53: We may not need to change this. Below change for CAStor version should be sufficient.

niladrih: We are using this... CAstor version is left in because to header requirements by the server.

niladrih: I could bump the crate version, if that is preferred.

niladrih:
bors try

niladrih:
bors try

datacore-vvarakantham (approved):
LGTM

niladrih:
bors merge

call-home/src/common/constants.rs
```diff
@@ -62,3 +62,6 @@ pub(crate) fn call_home_frequency() -> std::time::Duration {
         })
         .unwrap()
 }
+
+/// Release version
+pub(crate) const RELEASE_VERSION: &str = "2.0.1";
```
tiagolobocastro, line 67: We should not hardcode this, we'll keep getting it wrong as we forget, please use the version_info crate from the control-plane dep.

niladrih: https://github.com/openebs/mayastor-dependencies/tree/develop/version-info

.. this one?

tiagolobocastro: here: https://github.com/openebs/mayastor-extensions/blob/efe832075f7f2fd770a0a58286f8cd8fbf212403/k8s/plugin/src/main.rs#L20

PR #125, opened 2023-03-07
feat(call-home): use version_info crate to generate git version

niladrih:
This change uses the version_info crate from the openebs/mayastor-dependencies repo (via dependencies/control-plane git submodule's dependency) to generate the version for the call-home payload and the HTTP header.

This uses the git tag, if present, otherwise defaults to the commit hash for the HEAD commit of the repo branch.

### DIFF
--- a/call-home/Cargo.toml
+++ b/call-home/Cargo.toml
@@ -14,7 +14,7 @@ kube = { version = "0.78.0", features = ["runtime", "derive"] }
 k8s-openapi = { version = "0.17.0", features = ["v1_20"] }
 futures = "0.3.26"
 tokio = { version = "1.25.0", features = ["full"] }
-clap = { version = "4.1.4", features = ["cargo", "derive"] }
+clap = { version = "4.1.8", features = ["cargo", "derive", "string"] }
 serde_json = "1.0.93"
 serde_yaml = "0.9.17"
 serde = { version = "1.0.152", features = ["derive"] }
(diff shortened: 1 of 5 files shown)


niladrih:
bors try

niladrih:
bors try

avishnu (approved):
lgtm

call-home/src/main.rs
```diff
@@ -50,7 +50,7 @@ async fn main() {
 
 async fn run() -> anyhow::Result<()> {
     let args = CliArgs::args();
-    let version = RELEASE_VERSION.to_string();
+    let version = release_version();
```
tiagolobocastro, line 53: missing change to the clap to set the correct version, see [here](https://github.com/openebs/mayastor-extensions/blob/efe832075f7f2fd770a0a58286f8cd8fbf212403/k8s/plugin/src/main.rs#L20) as example

call-home/Cargo.toml
```diff
@@ -32,3 +32,4 @@ chrono = "0.4.23"
 rand = "0.8.5"
 tempfile = "3.3.0"
 humantime = "2.1.0"
+version-info = {path = "../dependencies/control-plane/utils/dependencies/version-info"}
```
tiagolobocastro, line 35: Add default-features = false or perhaps simply use utils crate from control-plane which is what other crates use.

niladrih: Using the utils crate.

call-home/src/main.rs
niladrih: Added the clap attributes.

PR #126, opened 2023-03-07
feat(upgrade): add function to find rebuild in progress

sinhaashish:


sinhaashish:
bors merge

sinhaashish:
bors try

operators/src/upgrade/controller/reconciler.rs
```diff
@@ -552,3 +552,46 @@ pub async fn is_draining() -> Result<bool, Error> {
...
+                error
+            })?;
+
+        let v = vols.into_body();
+        volumes.extend(v.entries);
+        starting_token = v.next_token;
+    }
+
+    for volume in volumes.into_iter() {
+        is_rebuilding = if let Some(target) = &volume.state.target {
```
tiagolobocastro, line 583: is_rebuilding can be a local variable, or even better completely skipped, just break on the if.
We could also move this bit of code to inside the previous while as if we find a rebuild there's no need to keep probing.

PR #129, opened 2023-03-09
feta(upgrade): restart io engine upon upgrade

sinhaashish:


k8s/upgrade/src/constant.rs
```diff
@@ -40,7 +40,8 @@ pub(crate) const UPGRADE_OPERATOR_SERVICE_PORT: i32 = 8080;
 /// Service internal port constant for upgrade operator.
 pub(crate) const UPGRADE_OPERATOR_INTERNAL_PORT: i32 = 8080;
 /// This is the upgrade-operator container image.
-pub(crate) const UPGRADE_IMAGE: &str = "openebs/mayastor-operator-upgrade:develop";
+//pub(crate) const UPGRADE_IMAGE: &str = "openebs/mayastor-operator-upgrade:develop";
+pub(crate) const UPGRADE_IMAGE: &str = "sinhaashish/mayastor-operator-upgrade:v3";
```
tiagolobocastro, line 44: hmm, how should we set this properly?

operators/src/upgrade/common/error.rs
```diff
@@ -118,6 +118,10 @@ pub enum Error {
     /// Error for when all volumes are not unpublished.
     #[error("All volumes must be unpublished before upgrade: {}", reason)]
     VolumesNotUnpublishedError { reason: String },
+
+    /// When agent-core or api-rest or etcd pods are down
+    #[error("Error: {0}")]
```
tiagolobocastro, line 123: Add more descriptive error message?

operators/src/upgrade/rest/service.rs
```diff
@@ -125,9 +138,179 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
     }
 }
 
+pub async fn upgrade_data_plane() -> Result<(), Error> {
```
tiagolobocastro, line 141: missing doc comment

operators/src/upgrade/rest/service.rs
```diff
@@ -125,9 +138,179 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+    let initial_io_engine_pod_list: ObjectList<Pod> = pods.list(&io_engine_listparam).await?;
+    for pod in initial_io_engine_pod_list.iter() {
+        // Fetch the node name on which the io-engine pod is running
+        let node_name = pod
+            .spec
+            .as_ref()
+            .ok_or_else(|| anyhow::anyhow!("pod.spec.node_name is invalid"))?
+            .node_name
+            .as_ref()
+            .ok_or_else(|| anyhow::anyhow!("pod.spec.node_name is invalid"))?
```
tiagolobocastro, line 157: invalid/empty ?

operators/src/upgrade/rest/service.rs
```diff
@@ -125,9 +138,179 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+        match UpgradeConfig::get_config()
+            .rest_client()
+            .nodes_api()
+            .put_node_drain(node_name, DRAIN_FOR_UPGRADE)
+            .await
+        {
+            Ok(_) => {
+                // Wait for any node drain process to complete across the cluster.
+                while is_draining().await? {
+                    tokio::time::sleep(Duration::from_secs(10_u64)).await;
```
tiagolobocastro, line 176: should this be configurable?
Do we need a timeout, or no because we'll just wait forever anyway?

operators/src/upgrade/rest/service.rs
```diff
@@ -125,9 +138,179 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+                }
+                tracing::info!("Mayastor node: {:#?} drained successfully", node_name);
+
+                // Wait for 30 seconds for any rebuilds to kick in.
+                tokio::time::sleep(Duration::from_secs(30_u64)).await;
```
tiagolobocastro, line 181: why do rebuilds kick in now, and why 30s?

operators/src/upgrade/rest/service.rs
```diff
@@ -125,9 +138,179 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+                tracing::info!(
+                    "Deleting io-engine pod: {:#?} present on node: {:#?}",
+                    pod_name,
+                    node_name,
+                );
+
+                pods.delete(pod_name, &DeleteParams::default())
+                    .await?
+                    .map_left(|o| tracing::debug!("Deleting Pod: {:?}", o.status))
+                    .map_right(|s| tracing::info!("Deleted Pod: {:?}", s));
```
tiagolobocastro, line 199: If result is `Either::Left`, do we need to wait for deleted?

operators/src/upgrade/rest/service.rs
```diff
@@ -125,9 +138,179 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+                    .rest_client()
+                    .nodes_api()
+                    .delete_node_cordon(node_name, DRAIN_FOR_UPGRADE)
+                    .await
+                {
+                    Ok(_) => {
+                        tracing::info!("Node {:#?} uncordoned!", node_name);
+                    }
+                    Err(error) => {
+                        return Err(error.into());
```
tiagolobocastro, line 212: can just use ?; above

operators/src/upgrade/rest/service.rs
```diff
@@ -125,9 +138,179 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+                // Validate the control plane pod is up and running
+                is_control_plane_running().await?
+            }
+            Err(error) => {
+                return Err(error.into());
+            }
+        }
+    }
+    Ok(())
+}
```
tiagolobocastro, line 228: How about using a state machine here, would making the boundaries of each step clearer?

operators/src/upgrade/rest/service.rs
```diff
@@ -125,9 +138,179 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+        UpgradeConfig::get_config().k8s_client().client(),
+        UpgradeConfig::get_config().namespace(),
+    );
+
+    let pod_list: ObjectList<Pod> = pods
+        .list(&ListParams::default().labels(AGENT_CORE_POD_LABEL))
+        .await?;
+    if !all_pods_are_ready(pod_list) {
+        return Err(Error::ControlPlanePodsDown(
+            "Helm not installed".to_string(),
```
tiagolobocastro, line 286: I don't understand this, why is help not installed?

operators/src/upgrade/rest/service.rs
```diff
@@ -125,9 +138,179 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+        ));
+    }
+    Ok(())
+}
+
 /// Get  upgrade.
 #[get("/upgrade")]
 pub async fn get_upgrade() -> impl Responder {
+    upgrade_data_plane().await;
```
tiagolobocastro, line 313: oh, we're upgrading the data plane on a get? Seems to violate http rules... gets should be safe operations

operators/src/upgrade/rest/service.rs
sinhaashish: The replica rebuild present in volume target is cached. So if 
 some operation is performed against the cluster, you might have to wait for some time to have the latest value. and that value is 30 seconds here

operators/src/upgrade/rest/service.rs
sinhaashish: What do you mean by state machine here ?

operators/src/upgrade/rest/service.rs
```diff
@@ -125,6 +138,176 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+            .node_name
+            .as_ref()
+            .ok_or_else(|| anyhow::anyhow!("pod.spec.node_name is empty"))?
+            .as_str();
+
+        tracing::info!(
+            "Upgrade starting for data-plane pod: {:#?} present on node: {:#?} ",
+            pod.metadata.name,
+            node_name
+        );
```
tiagolobocastro, line 165: Use field attributes on all tracing events, eg:
```suggestion
        tracing::info!(
            pod.name = %pod.metadata.name.unwrap_or_default(),
            node.name = %node_name,
            "Upgrade starting for data-plane pod"
        );
```
This makes it more machine readable and easy to search across the traces.

operators/src/upgrade/rest/service.rs
tiagolobocastro: State machine itself is probably not really necessary, simply making each state its own function for example would already help tidy things up here I think, eg states:
Drain Node, WaitDrain, RestartDataPlane, etc...
Then we could have:
```rust
let node_name = "node-1";
issue_node_drain(&node_name).await?;
wait_node_drain(&node_name).await?;
restart_data_plane(&node_name).await?;
```

sinhaashish: done , thanks

operators/src/upgrade/rest/service.rs
sinhaashish: no, we dont ned to wait, we just need to issue the delete command on pod.

operators/src/upgrade/rest/service.rs
```diff
@@ -125,6 +138,239 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+        .await?
+        .map_left(|o| tracing::debug!("Deleting Pod: {:?}", o.status))
+        .map_right(|s| tracing::info!("Deleted Pod: {:?}", s));
+    Ok(())
+}
+
+/// Wait for the data plane pod to come up on the given node.
+pub async fn wait_node_drain() -> Result<(), Error> {
+    while is_draining().await? {
+        tracing::info!("Mayastor node is still draining");
```
tiagolobocastro, line 238: Do we need these (here and below), could fill the log?

operators/src/upgrade/controller/reconciler.rs
```diff
@@ -533,18 +533,19 @@ pub async fn is_draining() -> Result<bool, Error> {
...
-        {
-            match cordondrainstate {
-                CordonDrainState::cordonedstate(_) => false,
-                CordonDrainState::drainingstate(_) => true,
-                CordonDrainState::drainedstate(_) => false,
+        let node_spec = match &node.spec {
+            Some(node_spec) => node_spec,
+            None => {
+                return Err(Error::NodeSpecNotPresent {
+                    node: node.clone().id,
```
tiagolobocastro, line 540: oh, node.id should suffice

operators/src/upgrade/controller/reconciler.rs
```diff
@@ -553,6 +554,30 @@ pub async fn is_draining() -> Result<bool, Error> {
...
+        .rest_client()
+        .nodes_api()
+        .get_node(node_name)
+        .await?;
+    let node_body = &node.into_body();
+    let node_spec = match &node_body.spec {
+        Some(node_spec) => node_spec,
+        None => {
+            return Err(Error::NodeSpecNotPresent {
+                node: node_body.clone().id,
```
tiagolobocastro, line 568: node_body.id should suffice, remove the & from 563.

operators/src/upgrade/rest/service.rs
```diff
@@ -125,6 +138,234 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+        Ok(_) => {
+            tracing::info!(node.name = node_name, "Node is uncordoned");
+        }
+        Err(_) => {
+            Error::NodeUncordonError {
+                node_name: node_name.to_string(),
+            };
+        }
+    }
+    Ok(())
```
tiagolobocastro, line 211: Semicolon here is not right:

```suggestion
    {
        Ok(_) => {
            tracing::info!(node.name = node_name, "Node is uncordoned");
            Ok(())
        }
        Err(_) => {
            Error::NodeUncordonError {
                node_name: node_name.to_string(),
            }
        }
    }
```

operators/src/upgrade/rest/service.rs
```diff
@@ -125,6 +138,234 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+    while is_data_plane_pod_running(node_name).await? {
+        tokio::time::sleep(Duration::from_secs(10_u64)).await;
+    }
+    Ok(())
+}
+
+///  Wait for the rebuild to complete if any
+pub async fn wait_for_rebuild() -> Result<(), Error> {
+    // Wait for 30 seconds for any rebuilds to kick in.
+    tokio::time::sleep(Duration::from_secs(30_u64)).await;
```
tiagolobocastro, line 253: 30s might not be enough here, maybe should be conservative and set to 1m.
Later we can add api to bypass cache.

operators/src/upgrade/rest/service.rs
```diff
@@ -125,6 +138,234 @@ pub async fn apply_upgrade() -> Result<HttpResponse, RestError> {
...
+        Ok(_) => {
+            tracing::info!(
+                node.name = %node_name,
+                "Drain started"
+            );
+        }
+        Err(_) => {
+            Error::NodeDrainError {
+                node_name: node_name.to_string(),
+            };
```
tiagolobocastro, line 277: same thing here with the semicolon

tiagolobocastro (commented):
lgtm, but is the target branch correct?

sinhaashish:
bors try

tiagolobocastro (approved):

niladrih (approved):

sinhaashish:
bors merge

PR #131, opened 2023-03-12
feat(upgrade): add rebuild validation as preflight check 

sinhaashish:


niladrih (approved):

operators/src/upgrade/common/error.rs
```diff
@@ -138,6 +138,9 @@ pub enum Error {
     /// Node uncordon error.
     #[error("Node uncordon error {}", node_name)]
     NodeUncordonError { node_name: String },
+    // /// Error in
+    // #[error("No replica rebuild {}", reason)]
+    // FetchRebuildingStatusError { reason: String },
```
tiagolobocastro, line 143: why is this commented out?

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -0,0 +1,20 @@
...
+use operators::upgrade::{common::error, controller::reconciler};
+
+/// Validation to be done before applying upgrade.
+pub async fn preflight_check() -> Result<(), error::Error> {
+    console_logger::info(user_prompt::UPGRADE_WARNING);
+    match reconciler::is_rebuilding().await {
+        Ok(is_rebuilding) => {
+            if is_rebuilding {
+                console_logger::warn(user_prompt::REBUILD_WARNING, "");
+                std::process::exit(1);
```
tiagolobocastro, line 11: Shouldn't we return an error instead of exit?

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -0,0 +1,20 @@
+use crate::user_prompt;
+use operators::upgrade::{common::error, controller::reconciler};
+
+/// Validation to be done before applying upgrade.
+pub async fn preflight_check() -> Result<(), error::Error> {
+    console_logger::info(user_prompt::UPGRADE_WARNING);
+    match reconciler::is_rebuilding().await {
```
tiagolobocastro, line 7: maybe split the rebuild into a separate function? So if we ever have more checks it can be more "modular"

k8s/upgrade/src/preflight_validations.rs
sinhaashish: Its not an error . It just writes to console and exit.

Abhinandan-Purkait (approved):

sinhaashish:
bors try

sinhaashish:
bors merge

k8s/upgrade/src/preflight_validations.rs
tiagolobocastro: What I mean is shouldn't we return an error and then the code handling the error decide to exit?
Seems like we have a bunch of process exit all over the place ...

PR #135, opened 2023-03-15
feat(upgrade): list single volume replica before upgrade

sinhaashish:
The o/p of kubeclt-mayastor upgrade
``` 
ashish@vm:~/code/rust/mayastor-extensions(single-replica)$ ./target/debug/kubectl-mayastor  upgrade

Below are the list of cordoned nodes. Since there are some cordoned nodes,
 the rebuild will happen on another nodes so please have enough available free space or else the rebuild will get stuck forever.
 node-0-64716

The list below shows the single replica volumes in cluster.
These single replica volumes may not be accessible during upgrade.
To skip this validation, please retry  with '--skip-single-replica-volume-validation` flag.
 ms-volume-claim
```

### DIFF
--- a/k8s/upgrade/src/preflight_validations.rs
+++ b/k8s/upgrade/src/preflight_validations.rs
@@ -1,26 +1,143 @@
-use crate::user_prompt;
-use operators::upgrade::{common::error, controller::reconciler};
+use crate::{
+    constant::SINGLE_REPLICA_VOLUME, upgrade_lib, upgrade_resources::upgrade::get_pvc_from_uuid,
+    user_prompt,
+};
+use openapi::models::CordonDrainState;
+use operators::upgrade::common::error;
+use std::{collections::HashSet, path::PathBuf};
+use tracing::error;
 
 /// Validation to be done before applying upgrade.
-pub async fn preflight_check() -> Result<(), error::Error> {
+pub async fn preflight_check(
+    kube_config_path: Option<PathBuf>,
+    timeout: humantime::Duration,
+) -> Result<(), error::Error> {
     console_logger::info(user_prompt::UPGRADE_WARNING);
-    rebuild_in_progress_validation().await?;
+    // Initialise the REST client.
+    let config = kube_proxy::ConfigBuilder::default_api_rest()
+        .with_kube_config(kube_config_path.clone())
+        .with_timeout(*timeout)
+        .build()
+        .await?;
+
+    let rest_client = upgrade_lib::RestClient::new_with_config(config);
+
+    rebuild_in_progress_validation(&rest_client).await?;
+
+    already_cordoned_nodes_validation(&rest_client).await?;
+
+    single_volume_replica_validation(&rest_client).await?;
+    Ok(())
+}
+
+/// Check for already cordoned node
+pub async fn already_cordoned_nodes_validation(
+    client: &upgrade_lib::RestClient,
+) -> Result<(), error::Error> {
+    let mut cordoned_nodes_list = Vec::new();
+    let nodes = client.nodes_api().get_nodes().await?;
+    let nodelist = nodes.into_body();
+    for node in nodelist {
+        let node_spec = match node.spec {
+            Some(node_spec) => node_spec,
+            None => return Err(error::Error::NodeSpecNotPresent { node: node.id }),
+        };
+        if matches!(
+            node_spec.cordondrainstate,
+            Some(CordonDrainState::cordonedstate(_))
+        ) {
+            cordoned_nodes_list.push(node.id);
+        }
+    }
+    if !cordoned_nodes_list.is_empty() {
+        console_logger::warn(
+            user_prompt::CORDONED_NODE_WARNING,
+            &cordoned_nodes_list.join("\n"),
+        );
+    }
+    Ok(())
+}
+
+/// Check single replica volumes.
+pub async fn single_volume_replica_validation(
+    client: &upgrade_lib::RestClient,
+) -> Result<(), error::Error> {
+    // let mut single_replica_volumes = Vec::new();
+    // The number of volumes to get per request.
+    let max_entries = 200;
+    let mut starting_token = Some(0_isize);
+    let mut volumes = Vec::with_capacity(max_entries as usize);
+
+    // The last paginated request will set the `starting_token` to `None`.
+    while starting_token.is_some() {
+        let vols = client
+            .volumes_api()
+            .get_volumes(max_entries, None, starting_token)
+            .await
+            .map_err(|error| {
+                error!(?error, "Failed to list volumes");
+                error
+            })?;
+
+        let v = vols.into_body();
+        let single_rep_vol_ids: Vec<String> = v
+            .entries
+            .into_iter()
+            .filter(|volume| volume.spec.num_replicas == SINGLE_REPLICA_VOLUME)
+            .map(|volume| volume.spec.uuid.to_string())
+            .collect();
+        volumes.extend(single_rep_vol_ids);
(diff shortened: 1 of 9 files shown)


k8s/upgrade/src/preflight_validations.rs
```diff
@@ -1,26 +1,101 @@
...
+            .volumes_api()
+            .get_volumes(max_entries, None, starting_token)
+            .await
+            .map_err(|error| {
+                error!(?error, "Failed to list volumes");
+                error
+            })?;
+
+        let v = vols.into_body();
+        volumes.extend(v.entries);
```
Abhinandan-Purkait, line 39: nit:- Instead of appending all the volumes in the vector you can simply filter the single volume replicas here and append in the list. That would save the below loop.

```
while let Some(starting_token) = starting_token {
    let vols = RestClient::client()
        .volumes_api()
        .get_volumes(max_entries, None, starting_token)
        .await
        .map_err(|error| {
            error!(?error, "Failed to list volumes");
            error
        })?;

    let single_rep_vol_ids = vols
        .into_body()
        .entries
        .into_iter()
        .filter(|volume| volume.spec.num_replicas == SINGLE_REPLICA_VOLUME)
        .map(|volume| volume.spec.uuid.to_string())
        .collect();

    volumes.extend(single_replica_volumes);

    starting_token = vols.into_body().next_token;
}
```

Something like this probably. Then you can directly create the hashset from the volume vector.

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -1,26 +1,101 @@
...
+    while starting_token.is_some() {
+        let vols = RestClient::client()
+            .volumes_api()
+            .get_volumes(max_entries, None, starting_token)
+            .await
+            .map_err(|error| {
+                error!(?error, "Failed to list volumes");
+                error
+            })?;
+
```
Abhinandan-Purkait, line 80: You can do the same here as well, filter here and return from here. If we get one rebuilding then no need of making further paginated api calls.

k8s/upgrade/Cargo.toml
```diff
@@ -10,6 +10,7 @@ path = "src/lib.rs"
 
 [dependencies]
 openapi = { path = "../../dependencies/control-plane/openapi", default-features = false, features = [ "tower-trace" ] }
+rest-plugin = { path = "../../dependencies/control-plane/control-plane/plugin", default-features = false }
```
tiagolobocastro, line 13: This doesn't seem right pulling the rest-plugin here. We already have a client elsewhere as well right, why don't we use that?

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -1,26 +1,101 @@
...
+        } else {
+            false
+        };
+
+        if is_rebuilding {
+            break;
         }
     }
-    Ok(())
+    Ok(is_rebuilding)
```
tiagolobocastro, line 100: Can avoid collecting all altogether

```suggestion
let volumes = vols.into_body();
starting_token = volumes.next_token;
for volume in volumes.entries {
        if let Some(target) = &volume.state.target {
            if target
                .children
                .iter()
                .any(|child| child.rebuild_progress.is_some()) {
                   return Ok(true);
                }
        }
}
Ok(false)
```

k8s/upgrade/src/upgrade_resources/upgrade.rs
```diff
@@ -598,6 +598,30 @@ pub async fn get_deployment_for_rest(ns: &str) -> Result<ObjectList<Deployment>,
     Ok(deployment.list(&lp).await?)
 }
 
+pub async fn get_pvc_from_uuid(uuid_list: HashSet<String>) -> Result<Vec<String>, Error> {
+    let mut single_replica_volumes_pvc = Vec::new();
```
tiagolobocastro, line 602: In Rust we declare variables close to usage rather than at the top like C.

k8s/upgrade/Cargo.toml
sinhaashish: IMHO, i dont find any place where its used directly . 
One thing which i tried earlier was `UpgradeConfig::get_config().rest_client()` , but this didn't worked as UpgradeConfig is not initialized by the time i am calling this validation . 

Can you suggest anything here.

tiagolobocastro: Another point is that we can't use `RestClient` at all here since it won't do the port forwarding.
We have to get `UpgradeConfig` initialized before this call is done.

sinhaashish: @tiagolobocastro  PTAL , created a new client

sinhaashish:
bors try

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -1,26 +1,147 @@
...
+            .entries
+            .into_iter()
+            .filter(|volume| volume.spec.num_replicas == SINGLE_REPLICA_VOLUME)
+            .map(|volume| volume.spec.uuid.to_string())
+            .collect();
+        volumes.extend(single_rep_vol_ids);
+        starting_token = v.next_token;
+    }
+
+    let single_volume_replicas_list: Vec<String> = volumes.iter().map(|s| s.to_string()).collect();
```
Abhinandan-Purkait, line 95: Why do we again need to convert to string? The `volumes` vector already has the String right?

Abhinandan-Purkait (approved):

k8s/upgrade/src/preflight_validations.rs
sinhaashish: ah, thanks removed it

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -1,26 +1,145 @@
...
+    rebuild_in_progress_validation(&rest_client).await?;
+
+    already_cordoned_nodes_validation(&rest_client).await?;
+
+    single_volume_replica_validation(&rest_client).await?;
+    Ok(())
+}
+
+/// Check single replica volumes.
+pub async fn already_cordoned_nodes_validation(
```
tiagolobocastro, line 34: comment this to be pertaining to the function below

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -1,26 +1,145 @@
...
+            None => return Err(error::Error::NodeSpecNotPresent { node: node.id }),
+        };
+        match node_spec.cordondrainstate {
+            Some(CordonDrainState::cordonedstate(_)) => {
+                cordoned_nodes_list.push(node.id);
+            }
+            Some(CordonDrainState::drainingstate(_)) => (),
+            Some(CordonDrainState::drainedstate(_)) => (),
+            None => (),
+        };
```
tiagolobocastro, line 52:
```suggestion
        if matches!(node_spec.cordondrainstate, Some(CordonDrainState::cordonedstate(_))) {
                cordoned_nodes_list.push(node.id);
        }
```
or
```suggestion
        match node_spec.cordondrainstate {
            Some(CordonDrainState::cordonedstate(_)) => {
                cordoned_nodes_list.push(node.id);
            }
            _ => {}
        }
```

tiagolobocastro (approved):
Do we also need to check for draining nodes?

sinhaashish:
> Do we also need to check for draining nodes?
Discussed this with @vishnuitta  and its updated in https://dcsw-bolt.atlassian.net/browse/GTM-441. 
It was stated that we can have a warning in the console about the cordoned nodes and the rebuild will take time. 
The concept of  150% free space in cluster needed for the volumes present on cordoned node is dropped.

PR #136, opened 2023-03-21
feat(upgrade): add the dry run flag

sinhaashish:


k8s/plugin/README.md
```diff
@@ -334,10 +334,31 @@ SUBCOMMANDS:
...
+
+  Usage: kubectl-mayastor upgrade [OPTIONS]
+
+  Options:
+    -r, --rest <REST>
+        The rest endpoint to connect to
+    -u, --upgrade-operator-endpoint <UPGRADE_OPERATOR_ENDPOINT>
+        Endpoint of upgrade operator service, if left empty then it will try to parse endpoints from upgrade operator service(K8s service resource)
+    -d, --dry-run
+        If set then it will display all the validations o/p but wont execute upgrade
```
Abhinandan-Purkait, line 350: `display all the validations o/p but wont execute upgrade` , we don't need the  `If set then it will`

Abhinandan-Purkait (approved):

k8s/plugin/src/main.rs
```diff
@@ -174,13 +174,15 @@ async fn execute(cli_args: CliArgs) {
...
                     std::process::exit(1);
                 });
-                resources
-                    .apply(
-                        &cli_args.namespace,
-                        cli_args.kube_config_path,
-                        cli_args.timeout,
-                    )
-                    .await;
+                if !resources.dry_run {
```
tiagolobocastro, line 177: Would we want to dry_run the apply as well?
Example: show which operations we would do on apply?

tiagolobocastro (approved):

k8s/plugin/src/main.rs
sinhaashish: Will add that in another PR , For now merging this PR

sinhaashish:
bors merge

PR #137, opened 2023-03-22
Add k8s Job for upgrade

sinhaashish:
As we are moving from operator to K8s job for upgrade.
This PR removes the deployment service and operator related code and add the K8s Job.


k8s/upgrade/src/upgrade_resources/upgrade.rs
```diff
@@ -378,229 +384,100 @@ impl UpgradeResources {
...
+                    println!("Job does not exist");
                 }
             }
         }
         Ok(())
     }
 
-    /// Install the upgrade resources
-    pub async fn install(ns: &str) {
+    /// Create the resources for upcrate
```
tiagolobocastro, line 394:
```suggestion
    /// Create the resources for upgrade
```

k8s/upgrade/src/user_prompt.rs
```diff
@@ -2,10 +2,10 @@
...
 
 /// Warning to users before doing an upgrade.
-pub const REBUILD_WARNING: &str =  "\nThe cluster is rebuilding replica of some volumes.\nPlease try after some time or skip this validation by retrying with '--ignore-rebuild` flag.";
+pub const REBUILD_WARNING: &str =  "\nThe cluster is rebuilding replica of some volumes.\nPlease try after some time or skip this validation by retrying with '--ignore-replica-rebuild` flag.";
 
 /// Warning to users before doing an upgrade
-pub const SINGLE_REPLICA_VOLUME_WARNING: &str =  "\nThe list below shows the single replica volumes in cluster.\nThese single replica volumes may not be accessible during upgrade.\nTo skip this validation, please retry  with '--skip-single-replica-volume-validation` flag.";
+pub const SINGLE_REPLICA_VOLUME_WARNING: &str =  "\nThe list below shows the single replica volumes in cluster.\nThese single replica volumes may not be accessible during upgrade.\nTo skip this validation, please retry  with '--bypass-single-replica-volume-validation` flag.";
```
tiagolobocastro, line 8: suggest keeping skip since kubectl also uses skip.

Also seems `--ignore-replica-rebuild` should also be `--skip-replica-rebuild-validation` ?
A question for the future, should this be a rebuild validation or a volume validation, eg: by default only proceed if all volumes are online which effectively means there's no rebuilds?

tiagolobocastro (approved):

k8s/plugin/src/main.rs
```diff
@@ -155,33 +146,19 @@ async fn execute(cli_args: CliArgs) {
...
                 });
-                if !resources.dry_run {
-                    resources
-                        .apply(
-                            &cli_args.namespace,
-                            cli_args.kube_config_path,
-                            cli_args.timeout,
-                        )
-                        .await;
+                if resources.dry_run {
```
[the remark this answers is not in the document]
sinhaashish: Validation soul still be in client only, The upgrade library creates a job which has the rest client . 
The validations are done before creating the K8s job. If any validations fails we dont create the JOB.

k8s/upgrade/src/user_prompt.rs
sinhaashish: IMHO its a rebuild validation, which actually means that some volume are not healthy and thus rebuild is in progress.
Since we have a single replica volume validation , going with rebuild validations here. Can modify as suggested in other PR.

sinhaashish:
bors try

sinhaashish:
bors merge

PR #141, opened 2023-03-28
fix(plugin/upgrade): pass on --skip-data-plane-restart flag to upgrade

niladrih:
The kubectl-mayastor tool receives the optional flag `--skip-data-plane-restart` as an option to the upgrade subcommand. This flag currently incorrectly enables data-plane restart. This PR uses the value from the same flag and passes it to the upgrade Kubernetes Job verbatim.

Additionally, the flag is made optional for the Job, and only passed in if it is enabled. This is done because Clap v4's derive API cannot interpret string literals as bool inputs to flags.


sinhaashish:
bors try

niladrih:
bors retry

sinhaashish (approved):
lgtm

PR #153, opened 2023-04-03
feat(upgrade): add  version tag to upgrade resources &  add error file

sinhaashish:
```
ServiceAccount: mayastor-upgrade-service-account-v2-0-1 in namespace: mayastor already exist.
ClusterRole: mayastor-upgrade-role-v2-0-1 already exist
ClusterRoleBinding: mayastor-upgrade-role-binding-v2-0-1 already exist
Job: mayastor-upgrade-v2-0-1 in namespace: mayastor already exist
```

### DIFF
--- a/k8s/upgrade/Cargo.toml
+++ b/k8s/upgrade/Cargo.toml
@@ -10,6 +10,7 @@ path = "src/lib.rs"
 
 [dependencies]
 openapi = { path = "../../dependencies/control-plane/openapi", default-features = false, features = [ "tower-trace" ] }
+utils = { path = "../../dependencies/control-plane/utils/utils-lib" }
 kube-proxy = { path = "../proxy" }
 console-logger = { path = "../../console-logger" }
 operators = { path = "../../operators" }
@@ -25,5 +26,6 @@ http = "0.2.8"
 async-trait = "0.1.64"
 serde = "1.0.152"
 serde_json = "1.0.93"
+thiserror = "1.0.38"
 # Tracing
 tracing = "0.1.37"
--- a/k8s/upgrade/src/constant.rs
+++ b/k8s/upgrade/src/constant.rs
@@ -1,3 +1,5 @@
+use utils::version_info;
+
 /// This is used to create labels for the upgrade job.
 #[macro_export]
 macro_rules! upgrade_labels {
@@ -12,8 +14,37 @@ macro_rules! upgrade_labels {
 }
 
 /// Append the release name to k8s objects.
-pub(crate) fn upgrade_name_concat(release_name: &str, component_name: &str) -> String {
-    format!("{release_name}-{component_name}")
+pub(crate) fn upgrade_name_concat(
+    release_name: &str,
+    component_name: &str,
+    upgrade_to_branch: Option<&String>,
+) -> String {
+    let version = match upgrade_to_branch {
+        Some(tag) => tag.to_string(),
+        None => get_image_tag(),
+    };
+    format!("{release_name}-{component_name}-{version}")
+}
+
+/// Fetch the image tag to append to upgrade resources.
+pub(crate) fn get_image_tag() -> String {
+    let version = release_version();
+    match version {
+        Some(upgrade_job_image_tag) => {
+            let tag: Vec<_> = upgrade_job_image_tag.split('.').collect();
+            let major = *(tag.first().unwrap_or(&""));
+            let minor = *(tag.get(1).unwrap_or(&""));
+            let patch = *(tag.get(2).unwrap_or(&""));
+            format!("v{major}-{minor}-{patch}")
+        }
+        None => UPGRADE_JOB_TO_DEVELOP_TAG.to_string(),
+    }
+}
+
+/// Returns the git tag version (if tag is found) or simply returns the commit hash (12 characters).
+pub(crate) fn release_version() -> Option<String> {
+    let version_info = version_info!();
+    version_info.version_tag
 }
 
 /// Append the tag to image in upgrade.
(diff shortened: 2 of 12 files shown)


k8s/upgrade/src/constant.rs
```diff
@@ -13,7 +13,20 @@ macro_rules! upgrade_labels {
 
 /// Append the release name to k8s objects.
 pub(crate) fn upgrade_name_concat(release_name: &str, component_name: &str) -> String {
-    format!("{release_name}-{component_name}")
+    let version = get_image_tag();
+    format!("{release_name}-{component_name}-{version}")
+}
+
+pub(crate) fn get_image_tag() -> String {
```
tiagolobocastro, line 20: missing doc comments

k8s/upgrade/src/constant.rs
```diff
@@ -22,7 +35,7 @@ pub(crate) fn upgrade_image_concat(image_repo: &str, image_name: &str, image_tag
 }
 
 /// Upgrade job container image tag.
-pub(crate) const UPGRADE_JOB_IMAGE_TAG: &str = "develop";
+pub(crate) const UPGRADE_JOB_IMAGE_TAG: &str = "2.0.1";
```
tiagolobocastro, line 38: shouldn't this come from the current tag?
Can probably use utils from control-plane, might need some modifications to avoid fallback and revert back to branch name in that case.

sinhaashish:
bors try

sinhaashish:
bors try

k8s/upgrade/Cargo.toml
```diff
@@ -25,5 +26,6 @@ http = "0.2.8"
 async-trait = "0.1.64"
 serde = "1.0.152"
 serde_json = "1.0.93"
+thiserror = "1.0.38"
 # Tracing
 tracing = "0.1.37"
```
Abhinandan-Purkait, line 31: Please add a blank line below.

k8s/upgrade/src/constant.rs
```diff
@@ -12,17 +14,45 @@ macro_rules! upgrade_labels {
...
 /// Append the release name to k8s objects.
-pub(crate) fn upgrade_name_concat(release_name: &str, component_name: &str) -> String {
-    format!("{release_name}-{component_name}")
+pub(crate) fn upgrade_name_concat(
+    release_name: &str,
+    component_name: &str,
+    upgrade_to_develop: bool,
+) -> String {
+    if upgrade_to_develop {
+        "develop".to_string()
```
Abhinandan-Purkait, line 23: Should this be taken from a constant?

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -118,7 +121,7 @@ pub async fn rebuild_in_progress_validation(
 }
 
 /// Check for rebuild in progress.
-pub async fn rebuild_in_progress(client: &upgrade_lib::RestClient) -> Result<bool, error::Error> {
+pub async fn rebuild_in_progress(client: &upgrade_lib::RestClient) -> Result<bool, Error> {
```
Abhinandan-Purkait, line 124: nit: `rebuild_in_progress` --> `check_rebuild_in_progress`. I know its in doc comment, but the method name should also reflect that?

k8s/upgrade/Cargo.toml
```diff
@@ -25,5 +26,6 @@ http = "0.2.8"
 async-trait = "0.1.64"
 serde = "1.0.152"
 serde_json = "1.0.93"
+thiserror = "1.0.38"
```
tiagolobocastro, line 29: @niladrih is using snafu, let's either use snafu or thiserror, not both please

k8s/upgrade/src/constant.rs
```diff
@@ -12,17 +14,45 @@ macro_rules! upgrade_labels {
...
+/// Fetch the image tag to append to upgrade resources.
+pub(crate) fn get_image_tag() -> String {
+    let version = release_version();
+    match version {
+        Some(upgrade_job_image_tag) => {
+            let tag: Vec<_> = upgrade_job_image_tag.split('.').collect();
+            let major = *(tag.first().unwrap_or(&""));
+            let minor = *(tag.get(1).unwrap_or(&""));
+            let patch = *(tag.get(2).unwrap_or(&""));
+            format!("v{major}-{minor}-{patch}")
```
tiagolobocastro, line 39: need to handle pre-release

k8s/upgrade/src/constant.rs
tiagolobocastro: We should be able to have upgrade_to_branch instead of upgrade to devel? Ex: when we want to test the release branch but there's no version.

k8s/upgrade/Cargo.toml
niladrih: The upgrade crate and the upgrade-job crates are largely unrelated. No shared resources whatsoever. @tiagolobocastro do you mean this in the sense that using the same error crate reduces repo dependencies?

tiagolobocastro: They're not the same crate?? They should be in the same create IMHO, they're all about upgrade. A crate can have multiple targets.

niladrih: That is true, but development for each of them was started independently at different points in time. The k8s/upgrade-job crate is meant to replace the operators/upgrade crate, and the k8s/upgrade crate was extracted from the k8s/plugin crate after it had started getting bulky.

We can maybe put in a TODO or a GitHub issue with the integration plan.

k8s/upgrade/src/constant.rs
Abhinandan-Purkait: Yep, that makes sense.

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

sinhaashish:
bors merge

PR #154, opened 2023-04-04
feat: enable setting priorityClassName references

xUnholy:
Enable setting `priorityClassName` reference for scheduling https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/

This is critical for scheduling workloads that are significantly reliant upon within the cluster.

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -137,7 +137,7 @@ $ helm install my-release openebs/mayastor
 | loki-stack.&ZeroWidthSpace;promtail.&ZeroWidthSpace;enabled | Enables promtail for scraping logs from nodes | `true` |
 | loki-stack.&ZeroWidthSpace;promtail.&ZeroWidthSpace;tolerations | Disallow promtail from running on the master node | `[]` |
 | nodeSelector | Node labels for pod assignment ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/ Note that if multi-arch images support 'kubernetes.io/arch: amd64' should be removed and set 'nodeSelector' to empty '{}' as default value. | <pre>{<br>"kubernetes.io/arch":"amd64"<br>}</pre> |
-| priorityClassName |Pod scheduling priority ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/ | `""` |
+| priorityClassName | Pod scheduling priority ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/ | `""` |
 | obs.&ZeroWidthSpace;callhome.&ZeroWidthSpace;enabled | Enable callhome | `true` |
 | obs.&ZeroWidthSpace;callhome.&ZeroWidthSpace;logLevel | Log level for callhome | `"info"` |
 | obs.&ZeroWidthSpace;callhome.&ZeroWidthSpace;resources.&ZeroWidthSpace;limits.&ZeroWidthSpace;cpu | Cpu limits for callhome | `"100m"` |
--- a/chart/README.md
+++ b/chart/README.md
@@ -137,6 +137,7 @@ $ helm install my-release openebs/mayastor
 | loki-stack.&ZeroWidthSpace;promtail.&ZeroWidthSpace;enabled | Enables promtail for scraping logs from nodes | `true` |
 | loki-stack.&ZeroWidthSpace;promtail.&ZeroWidthSpace;tolerations | Disallow promtail from running on the master node | `[]` |
 | nodeSelector | Node labels for pod assignment ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/ Note that if multi-arch images support 'kubernetes.io/arch: amd64' should be removed and set 'nodeSelector' to empty '{}' as default value. | <pre>{<br>"kubernetes.io/arch":"amd64"<br>}</pre> |
+| priorityClassName |Pod scheduling priority ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/ | `""` |
 | obs.&ZeroWidthSpace;callhome.&ZeroWidthSpace;enabled | Enable callhome | `true` |
 | obs.&ZeroWidthSpace;callhome.&ZeroWidthSpace;logLevel | Log level for callhome | `"info"` |
 | obs.&ZeroWidthSpace;callhome.&ZeroWidthSpace;resources.&ZeroWidthSpace;limits.&ZeroWidthSpace;cpu | Cpu limits for callhome | `"100m"` |
@@ -148,4 +149,3 @@ $ helm install my-release openebs/mayastor
 | operators.&ZeroWidthSpace;pool.&ZeroWidthSpace;resources.&ZeroWidthSpace;limits.&ZeroWidthSpace;memory | Memory limits for diskpool operator | `"32Mi"` |
 | operators.&ZeroWidthSpace;pool.&ZeroWidthSpace;resources.&ZeroWidthSpace;requests.&ZeroWidthSpace;cpu | Cpu requests for diskpool operator | `"50m"` |
 | operators.&ZeroWidthSpace;pool.&ZeroWidthSpace;resources.&ZeroWidthSpace;requests.&ZeroWidthSpace;memory | Memory requests for diskpool operator | `"16Mi"` |
-
--- a/chart/templates/mayastor/agents/core/agent-core-deployment.yaml
+++ b/chart/templates/mayastor/agents/core/agent-core-deployment.yaml
@@ -25,11 +25,11 @@ spec:
         {{- include "base_pull_secrets" . }}
       initContainers:
         {{- include "base_init_core_containers" . }}
+      priorityClassName: {{ default "system-cluster-critical" .Values.priorityClassName }}
       {{- if .Values.nodeSelector }}
       nodeSelector: {{- toYaml .Values.nodeSelector | nindent 8 }}
       {{- end }}
       tolerations: {{- toYaml .Values.earlyEvictionTolerations | nindent 8}}
-      priorityClassName: system-cluster-critical # Priority class provided by k8s by default.
       containers:
         - name: agent-core
           resources:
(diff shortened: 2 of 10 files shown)


chart/templates/mayastor/agents/core/agent-core-deployment.yaml
```diff
@@ -25,6 +25,9 @@ spec:
         {{- include "base_pull_secrets" . }}
       initContainers:
         {{- include "base_init_core_containers" . }}
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName }}
```
tiagolobocastro, line 29: this is already defined a few lines below

chart/templates/mayastor/apis/api-rest-deployment.yaml
```diff
@@ -24,6 +24,9 @@ spec:
         {{- include "base_pull_secrets" . }}
       initContainers:
         {{- include "base_init_containers" . }}
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName }}
```
tiagolobocastro, line 28: already defined below

tiagolobocastro (commented):
Perhaps each comment should have a class defined in values yaml, as you did with io-engine? 
This would allow us to control things a bit, eg: one may not need or want to have call home or the diskpool operator on the same priority as the core-agent.
CC @Abhinandan-Purkait

chart/templates/mayastor/agents/core/agent-core-deployment.yaml
Abhinandan-Purkait: As @tiagolobocastro said this should probably come from `values.yaml` for all the components. And specifically for the control-plane components we should default to `system-cluster-critical`, if the value is not set.

xUnholy: Argh yes I did not see this one had it hardcoded, assumed that it wouldn't due to all the others not also having it, will amend, thanks for checking 👍

chart/templates/mayastor/apis/api-rest-deployment.yaml
xUnholy: Will remove the one below as it's hardcoded

xUnholy:
> Perhaps each comment should have a class defined in values yaml, as you did with io-engine? This would allow us to control things a bit, eg: one may not need or want to have call home or the diskpool operator on the same priority as the core-agent. CC @Abhinandan-Purkait

The reason I avoided doing this although I thought it was a good idea was due to all the other conditions like `nodeSelectors` and such would also need to be uplifted to be consistent, ideally for each service it goes one way or the other, once we get this merged I'd be happy to make another iteration where we put it on each service uniquely.

tiagolobocastro:
> The reason I avoided doing this although I thought it was a good idea was due to all the other conditions like `nodeSelectors` and such would also need to be uplifted to be consistent, ideally for each service it goes one way or the other, once we get this merged I'd be happy to make another iteration where we put it on each service uniquely.

This way though we're setting everything to `system-cluster-critical` by default?
I would rather wait for the changes where we add the class to each component or alternatively you could simply leave the global priorityClassName `""` on the components where it was already set use `system-cluster-critical` as default, ex: 
```
      priorityClassName: {{ default "system-cluster-critical" .Values.priorityClassName }}
```
and for the other ones we still have the if:
```
      {{- if .Values.priorityClassName }}
      priorityClassName: {{ .Values.priorityClassName }}
      {{- end }}
```

xUnholy:
@tiagolobocastro I'm not really a fan of setting `priorityClassName: {{ default "system-cluster-critical" .Values.priorityClassName }}` as this makes a lot of assumptions and should be determined by the engineer installing the chart - Ideally the default is in the chart and they either unset it which indicated they don't want to use PriorityClasses or they use a bespoke/predined one as is the case in my scenario.

Additionally I think we can both agree the conditions should be consistent in behaviour.

chart/values.yaml
```diff
@@ -284,6 +288,9 @@ io_engine:
   nodeSelector:
     openebs.io/engine: mayastor
     kubernetes.io/arch: amd64
+  # -- Pod scheduling priority
+  # ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/
+  priorityClassName: ""
```
avishnu, line 293: IMO, this is not needed, make use of the global field and refer the same in the io-engine-daemonset yaml template.

xUnholy: Updated.

avishnu (approved):

avishnu:
bors try

xUnholy:
bors try

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
bors try

chart/README.md
```diff
@@ -137,6 +137,7 @@ $ helm install my-release openebs/mayastor
 | loki-stack.&ZeroWidthSpace;promtail.&ZeroWidthSpace;enabled | Enables promtail for scraping logs from nodes | `true` |
 | loki-stack.&ZeroWidthSpace;promtail.&ZeroWidthSpace;tolerations | Disallow promtail from running on the master node | `[]` |
 | nodeSelector | Node labels for pod assignment ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/ Note that if multi-arch images support 'kubernetes.io/arch: amd64' should be removed and set 'nodeSelector' to empty '{}' as default value. | <pre>{<br>"kubernetes.io/arch":"amd64"<br>}</pre> |
+| priorityClassName |Pod scheduling priority ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/ | `""` |
```
avishnu, line 140: Please give a space between '|' and 'Pod'

Abhinandan-Purkait, line 140: You seem to be missing a space after `|`.

```
| priorityClassName | Pod scheduling priority ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/ | `""` |
```

Please change it to the above line.

chart/README.md
```diff
@@ -148,4 +149,3 @@ $ helm install my-release openebs/mayastor
 | operators.&ZeroWidthSpace;pool.&ZeroWidthSpace;resources.&ZeroWidthSpace;limits.&ZeroWidthSpace;memory | Memory limits for diskpool operator | `"32Mi"` |
 | operators.&ZeroWidthSpace;pool.&ZeroWidthSpace;resources.&ZeroWidthSpace;requests.&ZeroWidthSpace;cpu | Cpu requests for diskpool operator | `"50m"` |
 | operators.&ZeroWidthSpace;pool.&ZeroWidthSpace;resources.&ZeroWidthSpace;requests.&ZeroWidthSpace;memory | Memory requests for diskpool operator | `"16Mi"` |
-
```
avishnu, line 151: This line may be needed.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

chart/README.md
tiagolobocastro: hmm I don't think it can, the README file is autogenerated, whatever the commit hook does, it should be kep.

niladrih:
Hi Michael, here are the steps to generate the README.md file for the helm chart (we don't modify it directly):

1. Install nix -- https://nixos.org/download.html
2. Edit the /etc/nix/nix.conf file to add the line `experimental-features = nix-command` to it. Save and exit.
3. [Optional] Install rustup. This is only required if you want to build the rust projects. You don't need this for this helm chart change.
4. Log out and log back in, this is required after installing nix.
5. Move into the mayastor-extensions repo directory.
6. Execute the command `nix-shell`.
7. Make sure that you have your changes in the chart/values.yaml file.
8. Restore the chart/README.md file to its initial state. Run `git checkout 9e9dc6e8bf4c86b647ab1e579cd065c5c10ae114 -- chart/README.md`. The hash is the commit hash that was at the top of the develop branch before you committed your changes. Alternatively, just remove line 140, and add the newline back at the end of the file.
9. Execute the shell script 'scripts/helm/generate-readme.sh'... run `./scripts/helm/generate-readme.sh`
10. Run git status or any alternative version management tool equivalent command to confirm that the README.md includes changes which correspond to your chart/values.yaml changes.

xUnholy:
> Hi @xUnholy, here are the steps to generate the README.md file for the helm chart (we don't modify it directly):
> 
> 1. Install nix -- https://nixos.org/download.html
> 2. Edit the /etc/nix/nix.conf file to add the line `experimental-features = nix-command` to it. Save and exit.
> 3. [Optional] Install rustup. This is only required if you want to build the rust projects. You don't need this for this helm chart change.
> 4. Log out and log back in, this is required after installing nix.
> 5. Move into the mayastor-extensions repo directory.
> 6. Execute the command `nix-shell`.
> 7. Make sure that you have your changes in the chart/values.yaml file.
> 8. Restore the chart/README.md file to its initial state. Run `git checkout 9e9dc6e8bf4c86b647ab1e579cd065c5c10ae114 -- chart/README.md`. The hash is the commit hash that was at the top of the develop branch before you committed your changes. Alternatively, just remove line 140, and add the newline back at the end of the file.
> 9. Execute the shell script 'scripts/helm/generate-readme.sh'... run `./scripts/helm/generate-readme.sh`
> 10. Run git status or any alternative version management tool equivalent command to confirm that the README.md includes changes which correspond to your chart/values.yaml changes.

I'm happy for you to close this PR if you wish - I'm not really impressed by the lack of UX to make a simple commit to a helm chart, I appreciate the effort for everyone here but I simply don't have the time to go through all these steps to update such a simple use case of enabling `PrioirtyClasses` I'll just patch it on my own chart usage. Thanks 🙏

tiagolobocastro:
I appreciate that list is quite long, but it was sent in good faith for a complete setup including installing the shell that will install all dependencies for you.
In the simple case the git commit hook will automatically update the doc. If the commit hook is not installed you can simply run the ./scripts/helm/generate-readme.sh. 
We ought to document this really.. 
At any case, please feel free to close this and we’ll take this up and merge it, thank you!

PR #157, opened 2023-04-05
build(helm/upgrade): prepare chart before packing container

tiagolobocastro:
As things stand the tree is always pointing to the stable helm version, eg: 2.1.0 
When building upgrade images for pre-releases or patches we need to patch the chart so that it may point to the correct version.
This is done using the same publish script that is used to publish the chart to the helm repo.

If we're building without a tag then for now we simply ignore this..

### DIFF
--- a/nix/pkgs/extensions/cargo-project.nix
+++ b/nix/pkgs/extensions/cargo-project.nix
@@ -60,7 +60,6 @@ let
     "console-logger"
     "operators"
     "call-home"
-    "scripts"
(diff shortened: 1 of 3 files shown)


niladrih (approved):

niladrih:
Thanks @tiagolobocastro.

sinhaashish (approved):

tiagolobocastro:
bors merge

PR #158, opened 2023-04-06
fix: update chart version for main branch

r1jt:
Update the chart version from 0.0.0-$DATE_TIME to 0.0.0-main

### DIFF
--- a/scripts/helm/publish-chart-yaml.sh
+++ b/scripts/helm/publish-chart-yaml.sh
@@ -49,7 +49,7 @@ branch_chart_version()
     echo "0.0.0"
   elif [ "$CHECK_BRANCH" == "main" ]; then
     # Main has no meaningful version, other than the date-time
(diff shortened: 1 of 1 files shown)


scripts/helm/publish-chart-yaml.sh
```diff
@@ -49,7 +49,7 @@ branch_chart_version()
     echo "0.0.0"
   elif [ "$CHECK_BRANCH" == "main" ]; then
     # Main has no meaningful version, other than the date-time
-    echo "0.0.0-$DATE_TIME"
+    echo "0.0.0-main"
```
tiagolobocastro, line 52: hmm don't think we want to do this right? We want to support both ways, only if date-time is not set we should use the main postfix.

tiagolobocastro (commented):
Something is wrong with this PR, it's showing commits which are already on develop, please rebase or start with fresh branch off of develop.

PR #161, opened 2023-04-07
feat(upgrade): rectify image tag

sinhaashish:

PR #164, opened 2023-04-08
fix(upgrade-job): add  '--set' options for image.repoTags options

niladrih:
This is required because the helm template function 'default' uses the values set in image.repoTags.* keys. Leaving it empty while using the '--reuse-values' flag will result in a nil pointer error in helm.

Adds an extra `--set` flag to the umbrella chart `helm upgrade command` to update the release.version (e.g. https://github.com/openebs/charts/blob/main/charts/openebs/values.yaml#L21)

### DIFF
--- a/k8s/upgrade/src/constant.rs
+++ b/k8s/upgrade/src/constant.rs
@@ -61,7 +61,7 @@ pub(crate) fn upgrade_event_selector(release_name: &str, component_name: &str) -
     format!("{kind},{name_key}={name_value}")
 }
 
-/// Upgrade conatainers to develop.
+/// Upgrade containers to develop.
 pub(crate) const UPGRADE_JOB_TO_DEVELOP_TAG: &str = "develop";
 /// Upgrade job container image repository.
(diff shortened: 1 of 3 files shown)


k8s/upgrade/src/constant.rs
```diff
@@ -61,7 +61,7 @@ pub(crate) fn upgrade_event_selector(release_name: &str, component_name: &str) -
     format!("{kind},{name_key}={name_value}")
 }
 
-/// Upgrade conatainers to develop.
+/// Upgrade containers to develop.
```
niladrih, line 64: @avishnu -- PTAL
Ref: #161

niladrih:
bors try

niladrih:
bors try-

niladrih:
bors try

niladrih:
bors try

tiagolobocastro (approved):

sinhaashish (approved):

niladrih:
bors merge

PR #165, opened 2023-04-08
feat(upgrade-job): support for continuous eventing for the upgrade duration

niladrih:
This includes this behavior:
- Events are now published from a separate tokio::task.
- Events are now re-published every 20 min, if no new event is generated. This let's a user probe for the latest event, in case an event is lost to the k8s events' expiration duration of 1 hour.

Also increases helm upgrade timeout to 15 min (from the default 5m).

Also, removes some unused errors and fixes serde container attribute for renaming struct members to camelCase.

P. S.: I deleted the Cargo.lock and regenerated it with `cargo build`.

### DIFF
--- a/k8s/upgrade-job/src/events/event_recorder.rs
+++ b/k8s/upgrade-job/src/events/event_recorder.rs
@@ -1,7 +1,7 @@
 use crate::common::{
     constants::PRODUCT,
     error::{
-        EventPublish, EventRecorderOptionsAbsent, GetPod, JobPodHasTooManyOwners,
+        EventChannelSend, EventPublish, EventRecorderOptionsAbsent, GetPod, JobPodHasTooManyOwners,
         JobPodOwnerIsNotJob, JobPodOwnerNotFound, Result, SerializeEventNote,
     },
     kube_client::KubeClientSet,
@@ -10,10 +10,11 @@ use k8s_openapi::{api::core::v1::ObjectReference, serde_json};
 use kube::runtime::events::{Event, EventType, Recorder};
 use serde::Serialize;
 use snafu::{ensure, ResultExt};
-use std::fmt::Display;
+use std::{fmt::Display, time::Duration};
+use tokio::{select, sync::mpsc, task, time::sleep};
 
 #[derive(Serialize, Debug)]
-#[serde(rename(serialize = "camelCase"))]
+#[serde(rename_all(serialize = "camelCase"))]
 pub(crate) struct EventNote {
     from_version: String,
     to_version: String,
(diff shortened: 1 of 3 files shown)


niladrih:
bors try

niladrih:
bors try

Cargo.lock
```diff
@@ -313,9 +313,9 @@ checksum = "7de8ce5e0f9f8d88245311066a578d72b7af3e7088f32783804676302df237e4"
 
 [[package]]
 name = "async-stream"
-version = "0.3.4"
+version = "0.3.5"
```
tiagolobocastro, line 316: Should we be upgrading dependencies so close to release?

k8s/upgrade-job/src/events/event_recorder.rs
```diff
@@ -159,12 +160,39 @@ impl EventRecorderBuilder {
             resource_version: None,
         };
 
+        let (tx, mut rx) = mpsc::channel::<Event>(10);
```
tiagolobocastro, line 163: Consider using unbound here, or a much larger buffer perhaps. Otherwise the sender may block or fail, please check the behaviour of the channel type you are using.

k8s/upgrade-job/src/events/event_recorder.rs
```diff
@@ -159,12 +160,39 @@ impl EventRecorderBuilder {
...
+            while let Some(event) = &current_event {
+                // Hacky clone for the Event.
+                let event = Event {
+                    type_: event.type_,
+                    reason: event.reason.clone(),
+                    note: event.note.clone(),
+                    action: event.action.clone(),
+                    secondary: event.secondary.clone(),
+                };
+                if let Err(error) = event_handler.publish(event).await.context(EventPublish) {
```
tiagolobocastro, line 181: Should we have a queue of events that need to be published? Otherwise if there's an intermittent failure we'll miss events. If we keep a queue of failed publishes, we retry them?

Cargo.lock
niladrih: I'll restore the Cargo.lock, I guess.

k8s/upgrade-job/src/events/event_recorder.rs
niladrih: I'm planning on including this in a future release, but not this one.

k8s/upgrade-job/src/events/event_recorder.rs
niladrih: Done. Added unbounded channel.

niladrih:
bors try

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

PR #189, opened 2023-04-13
feat(k8s/upgrade): use release-2.1 tagged upgrade image on the release/2.1 branch

niladrih:
This change was brought in when cherry-picking the k8s/upgrade crate from the develop branch.

### DIFF
--- a/k8s/upgrade/src/constant.rs
+++ b/k8s/upgrade/src/constant.rs
@@ -37,7 +37,7 @@ pub(crate) fn get_image_tag() -> String {
             let patch = *(tag.get(2).unwrap_or(&""));
             format!("v{major}-{minor}-{patch}")
         }
-        None => UPGRADE_JOB_TO_DEVELOP_TAG.to_string(),
+        None => UPGRADE_JOB_IMAGE_TAG.to_string(),
     }
 }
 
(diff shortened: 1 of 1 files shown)


sinhaashish (approved):

k8s/upgrade/src/constant.rs
```diff
@@ -70,7 +70,7 @@ pub(crate) const HELM_RELEASE_NAME_LABEL: &str = "openebs.io/release";
 
 pub(crate) const DEFAULT_IMAGE_REGISTRY: &str = "docker.io";
 /// Upgrade containers to develop.
-pub(crate) const UPGRADE_JOB_TO_DEVELOP_TAG: &str = "develop";
+pub(crate) const UPGRADE_JOB_TO_DEVELOP_TAG: &str = "release-2.1";
```
avishnu, line 73: Why is the variable named as "to develop"?

niladrih: Used a more generic name for the variable.

avishnu (approved):
LGTM. But I'd like to see how do we address when the tagged image comes out (eg. 2.1.0)

niladrih:
> LGTM. But I'd like to see how do we address when the tagged image comes out (eg. 2.1.0)

We'd have to do it manually, as it stands now.

niladrih:
bors merge

PR #199, opened 2023-04-15
feat(k8s/upgrade-job): add wait for all event-publishes 

niladrih:
Presently, the drop implementation for EventRecorder cancels the event loop task before all of the events could be pulished. This change adds a channel to receive an acknowledgement message, after closing the primary event channel.

No cancellation is done, the worker should exit on the next line after the acknowledgement.

Additionally, adds/modifies some logs.

### DIFF
--- a/k8s/upgrade-job/src/events/event_recorder.rs
+++ b/k8s/upgrade-job/src/events/event_recorder.rs
@@ -1,8 +1,9 @@
 use crate::common::{
     constants::PRODUCT,
     error::{
-        EventChannelSend, EventPublish, EventRecorderOptionsAbsent, GetPod, JobPodHasTooManyOwners,
-        JobPodOwnerIsNotJob, JobPodOwnerNotFound, Result, SerializeEventNote,
+        EventChannelSend, EventFinalAcknowledgementSend, EventPublish, EventRecorderOptionsAbsent,
+        GetPod, JobPodHasTooManyOwners, JobPodOwnerIsNotJob, JobPodOwnerNotFound, Result,
+        SerializeEventNote,
     },
     kube_client::KubeClientSet,
 };
@@ -11,7 +12,8 @@ use kube::runtime::events::{Event, EventType, Recorder};
 use serde::Serialize;
 use snafu::{ensure, ResultExt};
 use std::{fmt::Display, time::Duration};
-use tokio::{select, sync::mpsc, task, time::sleep};
+use tokio::{select, sync::mpsc, time::sleep};
+use tracing::error;
 
 #[derive(Serialize, Debug)]
 #[serde(rename_all(serialize = "camelCase"))]
@@ -43,7 +45,6 @@ impl EventNote {
 pub(crate) struct EventRecorderBuilder {
     pod_name: Option<String>,
     namespace: Option<String>,
-    reporter_name: Option<String>,
     from_version: Option<String>,
     to_version: Option<String>,
 }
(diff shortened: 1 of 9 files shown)


niladrih:
bors try

k8s/upgrade-job/src/events/event_recorder.rs
```diff
@@ -179,38 +180,73 @@ impl EventRecorderBuilder {
...
+                    if self.ack_sender.is_some() {
+                        let sender = self.ack_sender.clone().unwrap();
+                        self.ack_sender = None;
+                        drop(sender);
+                    }
+                    return;
+                }
+                Err(warning) => {
+                    warn!(%warning);
+                    std::thread::sleep(retry_duration);
```
tiagolobocastro, line 243: You shouldn't block like this.. it'll prevent async code on the same executor from running.

niladrih:
bors try

k8s/upgrade-job/src/events/event_recorder.rs
```diff
@@ -179,38 +180,73 @@ impl EventRecorderBuilder {
...
 
 impl Drop for EventRecorder {
     fn drop(&mut self) {
         // Cancel the tokio task.
-        self.event_loop_killswitch.abort();
+        if self.event_sender.is_some() {
+            let sender = self.event_sender.clone().unwrap();
+            self.event_sender = None;
+            drop(sender);
+        }
```
tiagolobocastro, line 223:
```suggestion
        let _ = self.event_sender.take();
```

k8s/upgrade-job/src/events/event_recorder.rs
tiagolobocastro: You probably shouldn't write this kind of code in the drop, rust has no async drop yet sadly...

tiagolobocastro (commented):
Do you even need to abort anything, won't it end with grace if you simply don't take the task's abort handle?

niladrih:
>Do you even need to abort anything, won't it end with grace if you simply don't take the task's abort handle?

Well, I failed to notice that there was literall no way for that worker to go anywhere. Removed the acknowledgement Sender drop.

k8s/upgrade-job/src/events/event_recorder.rs
niladrih: Moved it to a different method, no longer on the drop.

niladrih:
bors try

sinhaashish (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

k8s/upgrade-job/src/events/event_recorder.rs
```diff
@@ -179,41 +179,38 @@ impl EventRecorderBuilder {
...
-        })
-        .abort_handle();
 
-        Ok(EventRecorder {
-            event_loop_killswitch: aborter,
-            event_sender: tx,
-            from_version,
-            to_version,
-        })
+            // Acknowledgement for exit of event loop, i.e. all events published, nothing left
```
tiagolobocastro, line 191: Don't think you need this, on the other side you just need to await the task completion?

k8s/upgrade-job/src/events/event_recorder.rs
```diff
@@ -274,6 +272,13 @@ impl EventRecorder {
...
             .await
-            .map_err(|error| tracing::error!(%error, "Failed to upgrade {PRODUCT}"));
+            .map_err(|error| error!(%error, "Failed to upgrade {PRODUCT}"));
+    }
+
+    /// Shuts down the event channel which makes the event loop worker exit its loop and return.
+    pub(crate) async fn shutdown_worker(mut self) {
+        let _ = self.event_sender.take();
+
+        self.ack_receiver.recv().await;
```
tiagolobocastro, line 282: here, a self.task.join().await would have done it I think

niladrih: Ooooh.... great idea

PR #204, opened 2023-04-20
feat(rebuild): provide faulted child wait setting via Helm

dsharma-dc:
The core-agent takes an argument which defines how much time core-agent waits for a faulted child's replica to be online again so that same child can be made online to initiate fast(log-based) rebuild. This wait duration is being made configurable via helm values by this change.


chart/values.yaml
```diff
@@ -154,6 +154,7 @@ agents:
         cpu: "500m"
         # -- Memory requests for core agents
         memory: "32Mi"
+    fast_rebuild_wait_period: "0s"
```
avishnu, line 157: Is the default value supposed to be 0s ?

avishnu: The variable should use camelCase naming convention, please refer other names.

chart/templates/mayastor/agents/core/agent-core-deployment.yaml
```diff
@@ -50,6 +50,7 @@ spec:
             - "--pool-commitment={{ .Values.agents.core.capacity.thin.poolCommitment }}"
             - "--volume-commitment-initial={{ .Values.agents.core.capacity.thin.volumeCommitmentInitial }}"
             - "--volume-commitment={{ .Values.agents.core.capacity.thin.volumeCommitment }}"
+            - "--faulted-child-wait-period={{ .Values.agents.core.fast_rebuild_wait_period }}"
```
avishnu, line 53: Name the cmdline argument similar to the helm field, refer the above volume-commitment example.

dsharma-dc: I thought of keeping it same but "faulted child wait" may not be as easy to comprehend for users compared to a more abstracted "fast rebuild wait" ?

chart/values.yaml
dsharma-dc: Yes default is supposed to be zero, in which case core-agent uses internally defined values and ignores this.
Will change to camelCase.

chart/templates/mayastor/agents/core/agent-core-deployment.yaml
avishnu: The field description and helm documentation would be sufficient imo

niladrih (approved):

chart/values.yaml
avishnu: It's better to keep the default value same as the one used by core agent internally

dsharma-dc: The 'default' value used by core-agent internally would also be dynamic. And the current logic checks if cli provided value is zero, then decide upon a default value internally. This is stated in documentation.

chart/templates/mayastor/agents/core/agent-core-deployment.yaml
dsharma-dc: Alright.

chart/values.yaml
avishnu: Shouldn't it be simple 1:1 mapping? If not, what's the calculation, how can the value be figured out dynamically?

dsharma-dc: Recalling the control-plane design discussion around this - the wait value will be taking into consideration factors like replica size, number of children etc because we wouldn't want to wait for same amount of time for all the volumes.

If the non-zero cli value is provided - it overrides internal decisions and we'll take cli value into consideration.
If cli value is zero - coreagent will decide the values.

tiagolobocastro: This is confusing. Just set it to "" and if it's "" then don't provide the argument and core agent will use the default value.

dsharma-dc: oh if `""` is allowed then that's good because core agent will consider it zero.

```
   #[clap(long, default_value = "0s")]
    pub(crate) faulted_child_wait_period: humantime::Duration,
```

dsharma-dc: We never wanted anyone to have a 'choice' between full and partial rebuild. That wasn't the idea behind this. This is so that system tests can reduce the total runtime of partial rebuild suite.

tiagolobocastro: Could be useful for testing as well or if we find a bug in partial rebuild for example

abhilashshetty04 (approved):

chart/values.yaml
```diff
@@ -154,6 +154,10 @@ agents:
         cpu: "500m"
         # -- Memory requests for core agents
         memory: "32Mi"
+    # -- How long core agent waits for a faulted child's replica to come back online so that
+    # the agent can attempt to make same child online and in turn utilize fast rebuild capability.
+    # A blank value "" means internally derived value will be used.
+    faultedChildWaitPeriod: "10m"
```
tiagolobocastro, line 160: Why is it 10m now, don’t we want it as “”?

dsharma-dc: It's "" , 10m came in from outdated diff.

chart/templates/mayastor/agents/core/agent-core-deployment.yaml
```diff
@@ -50,6 +50,7 @@ spec:
             - "--pool-commitment={{ .Values.agents.core.capacity.thin.poolCommitment }}"
             - "--volume-commitment-initial={{ .Values.agents.core.capacity.thin.volumeCommitmentInitial }}"
             - "--volume-commitment={{ .Values.agents.core.capacity.thin.volumeCommitment }}"
+            - "--faulted-child-wait-period={{ .Values.agents.core.faultedChildWaitPeriod }}"
```
tiagolobocastro, line 53:
```suggestion
            - "--volume-commitment={{ .Values.agents.core.capacity.thin.volumeCommitment }}"{{ if .Values.agents.core.faultedChildWaitPeriod }}
            - "--faulted-child-wait-period={{ .Values.agents.core.faultedChildWaitPeriod }}"{{ end }}
```

tiagolobocastro (commented):
Should we rename faultedChildWaitPeriod to something else, wait period might not be very meaningful unless you're familiar with the implementation?
Example: partialRebuildTimeout... something like that
@avishnu wdyt?

chart/templates/mayastor/agents/core/agent-core-deployment.yaml
dsharma-dc: > Should we rename faultedChildWaitPeriod to something else, wait period might not be very meaningful unless you're familiar with the implementation?
Example: partialRebuildTimeout... something like that
@avishnu wdyt?

This was discussed here:
https://github.com/openebs/mayastor-extensions/pull/204#discussion_r1172122374

dsharma-dc:
bors try

dsharma-dc:
bors merge

chart/templates/mayastor/agents/core/agent-core-deployment.yaml
avishnu: My preference would be to keep the helm field name similar to the code variable name. Isn't it the norm?

chart/values.yaml
```diff
@@ -154,6 +154,10 @@ agents:
         cpu: "500m"
         # -- Memory requests for core agents
         memory: "32Mi"
+    # -- How long core agent waits for a faulted child's replica to come back online so that
+    # the agent can attempt to make same child online and in turn utilize fast rebuild capability.
+    # A blank value "" means internally derived value will be used.
```
tiagolobocastro, line 159: I think we should reword this, a user doesn't know what a child is and doesn't know what child_online rpc is.
I suggest:
```suggestion
# -- If a faulted replica comes back online within this time period then it will be rebuilt using the partial rebuild capability (using a log of missed IO). Otherwise it will be fully rebuilt.
     # A blank value "" means an internally derived value will be used.
```

dsharma-dc: Sounds good. Will modify the documentation.

PR #207, opened 2023-04-23
fix: Correct etcd storage class helm values

zimmertr:
Docs [here](https://mayastor.gitbook.io/introduction/quickstart/deploy-mayastor#installation-via-helm) say the following:

> Installing Mayastor via the Helm chart sets the StorageClass as 'default' for Loki and etcd StatefulSets. This will work in clusters which have a StorageClass named 'default'. If there is no 'default' StorageClass in the cluster, storage provisioning will fail for Loki and etcd. In such cases, one can change the StorageClass to 'manual' which will provision a PersistentVolume of type hostPath. To do this, make the following changes in the values.yaml file:
> * loki.persistence.storageClassName: manual
> * etcd.persistence.storageClassName: manual

This PR updates the `etcd` storageclass value to correlate with that as well as be consistent with the Loki value. 

Related: https://github.com/openebs/mayastor-docs/pull/123


zimmertr:
Also fixed the loki-stack conditional in `local-py.yml`

niladrih:
I don't thing we can work this into the helm chart very easily. Also, I don't think we have a need for this kind of symmetry amongst the values options. Closing this PR.

PR #218, opened 2023-05-02
feat(upgrade): add time stamp to tag for testing envt

sinhaashish:
In the testing environment the images are pushed with timestamp. Since the upgrade image were not time stamped , the testing environment used older images,

A new flag `image_tag` needs to be passed if users wants to use a custom tag
```
ashish@vm:~/code/rust/mayastor-extensions(image-for-develop)$ curl <<CI_REGISTRY_URL>>/v2/openebs/mayastor-upgrade-job/tags/list | jq . | grep "release"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1816  100  1816    0     0   9634      0 --:--:-- --:--:-- --:--:--  9659
    "release-2.1-2023-04-16-11-54-07",
    "release-2.1-2023-04-18-00-07-11",
    "release-2.1-2023-04-24-08-58-13",
    "release-2.1-2023-04-17-14-37-17",
    "release-2.1-2023-04-24-00-07-09",
    "release-2.1-2023-04-29-00-51-14",
    "release-2.1-2023-04-20-00-07-13",
    "release-2.1-2023-04-20-12-52-20",
    "release-2.1-2023-04-15-00-51-10",
    "release-2.1-2023-04-21-00-07-13",
    "release-2.1-2023-04-19-00-07-09",
    "release-2.1-2023-04-22-00-51-12",
    "release-2.1-2023-04-25-00-07-13",
    "release-2.1",
    "release-2.1-2023-04-26-00-07-12",
    "release-2.1-2023-04-27-00-07-11",
```
At present the testing is using the image tag as `release-2.1` and 
Same is the case with develop branch as well.

@w3aman @blaisedias @rohan2794  you need to fetch the latest build tag and pass it  using`--image_tag`  flag like `--image_tag release-2.1-2023-04-27-00-07-11` so  that the upgrade job takes the image which has the same tag as passed by user.

### DIFF
--- a/k8s/upgrade/src/constant.rs
+++ b/k8s/upgrade/src/constant.rs
@@ -37,15 +37,20 @@ pub(crate) fn upgrade_obj_suffix() -> String {
 }
 
 /// Fetch the image tag for upgrade job's pod.
-pub(crate) fn get_image_version_tag() -> String {
-    let version = release_version();
-    match version {
-        Some(upgrade_job_image_tag) => upgrade_job_image_tag,
-        None => UPGRADE_JOB_IMAGE_TAG.to_string(),
+pub(crate) fn get_image_version_tag(image_tag: Option<&String>) -> String {
+    match image_tag {
+        Some(tag) => tag.to_string(),
+        None => {
+            let version = release_version();
+            match version {
+                Some(upgrade_job_image_tag) => upgrade_job_image_tag,
+                None => UPGRADE_JOB_IMAGE_TAG.to_string(),
+            }
+        }
     }
 }
 
-/// Returns the git tag version (if tag is found) or simply returns the commit hash (12 characters).
+/// Returns the git tag version (if tag is found)
 pub(crate) fn release_version() -> Option<String> {
     let version_info = version_info!();
     version_info.version_tag
--- a/k8s/upgrade/src/upgrade_resources/upgrade.rs
+++ b/k8s/upgrade/src/upgrade_resources/upgrade.rs
@@ -39,7 +39,8 @@ pub enum DeleteResources {
 /// Delete Upgrade resource.
 #[derive(clap::Args, Debug)]
 pub struct DeleteUpgradeArgs {
-    /// If set then upgrade will skip the io-engine pods restart
+    /// Used for fetching the name of the upgrade resources.
+    /// Pass the name of the branch to which upgrade has happened earlier.
     #[clap(global = true, hide = true, long, short = 'u')]
     pub upgrade_to_branch: Option<String>,
 }
@@ -97,6 +98,10 @@ pub struct UpgradeArgs {
     /// Upgrade to the specified branch.
     #[clap(global = true, hide = true, long, short = 'u')]
     pub upgrade_to_branch: Option<String>,
+
+    /// Upgrade tag for testing, contains timestamp.
+    #[clap(hide = true, long)]
+    pub image_tag: Option<String>,
 }
 
 impl UpgradeArgs {
@@ -107,6 +112,7 @@ impl UpgradeArgs {
             namespace,
             self.skip_data_plane_restart,
             self.upgrade_to_branch.as_ref(),
+            self.image_tag.as_ref(),
         )
         .await;
         console_logger::info(UPGRADE_JOB_STARTED, "");
@@ -487,6 +493,7 @@ impl UpgradeResources {
         action: Actions,
         skip_data_plane_restart: bool,
         upgrade_to_branch: Option<&String>,
+        image_tag: Option<&String>,
     ) -> Result<(), Error> {
         let job_name = upgrade_name_concat(
             &self.release_name,
@@ -515,7 +522,7 @@ impl UpgradeResources {
         } else {
             match action {
                 Actions::Create => {
-                    let upgrade_job_image_tag = get_image_version_tag();
+                    let upgrade_job_image_tag = get_image_version_tag(image_tag);
                     let rest_deployment = get_deployment_for_rest(ns).await?;
                     let img = ImageProperties::try_from(rest_deployment)?;
                     let upgrade_deploy = objects::upgrade_job(
@@ -562,6 +569,7 @@ impl UpgradeResources {
         ns: &str,
         skip_data_plane_restart: bool,
         upgrade_to_branch: Option<&String>,
+        image_tag: Option<&String>,
     ) {
         match UpgradeResources::new(ns).await {
             Ok(uo) => {
@@ -599,6 +607,7 @@ impl UpgradeResources {
                         Actions::Create,
                         skip_data_plane_restart,
                         upgrade_to_branch,
+                        image_tag,
                     )
                     .await
                     .map_err(|error| {
@@ -616,7 +625,7 @@ impl UpgradeResources {
             Ok(uo) => {
                 // Delete the job
                 let _job = uo
-                    .job_actions(ns, Actions::Delete, false, upgrade_to_branch)
+                    .job_actions(ns, Actions::Delete, false, upgrade_to_branch, None)
                     .await
                     .map_err(|error| {
                         println!("Failure: {error}");


k8s/upgrade/src/upgrade_resources/upgrade.rs
```diff
@@ -39,7 +39,7 @@ pub enum DeleteResources {
 /// Delete Upgrade resource.
 #[derive(clap::Args, Debug)]
 pub struct DeleteUpgradeArgs {
-    /// If set then upgrade will skip the io-engine pods restart
+   /// Upgrade to the specified branch.
```
avishnu, line 42: Why do we have/need upgrade_to_branch as part of DeleteUpgradeArgs struct?

k8s/upgrade/src/upgrade_resources/upgrade.rs
```diff
@@ -97,6 +97,10 @@ pub struct UpgradeArgs {
     /// Upgrade to the specified branch.
     #[clap(global = true, hide = true, long, short = 'u')]
     pub upgrade_to_branch: Option<String>,
+
+    /// Upgrade tag for testing, contains timestamp.
+    #[clap(global = true, hide = true, long)]
+    pub upgrade_tag_for_test_envt: Option<String>,
```
avishnu, line 103: Is it a good idea to have a new variable for test environment inside production code? Is it normal?

sinhaashish: In the meeting it was discussed that we can introduce a flag , which this PR is doing. Since this flag is hidden, users cant use is unless they go through the code. The binary wont show this with help command.

k8s/upgrade/src/upgrade_resources/upgrade.rs
sinhaashish: `upgrade_to_branch` is used to get the name of the upgrade resources. It is appended as suffix to the upgrade resource names. This is only for testing by developers as they can test whether the upgrade to any feature branch can happen or not. This flag is hidden to normal users.
Upgrade jobs are like : 
`mayastor-upgrade-job-develop`
`mayastor-upgrade-job-thin-provisioning`


Since the branch name is appended as suffix , we need to the branch name while deleting as well. This is needed in getting the exact name of the upgrade resource.

blaisedias (approved):
Looks plausible.
But I don't think we should remove have `for_test_envt`  in the option.
Today this is required for testing - but maybe in the future not?

k8s/upgrade/src/upgrade_resources/upgrade.rs
```diff
@@ -97,6 +98,10 @@ pub struct UpgradeArgs {
...
```
niladrih, line 104: Redundant to have the 'upgrade' prefix in arguments meant for upgrade subcommands. Consider renaming the flags..

niladrih (commented):
left minor comments

k8s/upgrade/src/upgrade_resources/upgrade.rs
niladrih: Do we plan to feed the image tag timestamp suffix to the upgrade command, or the complete image tag... ?

niladrih: It seems like this flag could only possibly work with the 'upgrade' subcommand, and not the 'delete upgrade' or 'get upgrade-status' subcommands. The flag is maybe not a global flag...?

niladrih (commented):
left few more comments..

niladrih:
IMO, this feature could be a general image tag override. But in that case, can we adhere to the upgrade-path reliably?

k8s/upgrade/src/upgrade_resources/upgrade.rs
```diff
@@ -39,7 +39,8 @@ pub enum DeleteResources {
 /// Delete Upgrade resource.
 #[derive(clap::Args, Debug)]
 pub struct DeleteUpgradeArgs {
-    /// If set then upgrade will skip the io-engine pods restart
+   /// Used for fetching the name of the upgrade resources.
+   /// Pass the name of the branch to which upgrade has happened earlier.
```
niladrih, line 43: Looks like an indentation issue...
```suggestion
    /// Used for fetching the name of the upgrade resources.
    /// Pass the name of the branch to which upgrade has happened earlier.
```

niladrih (commented):
nit: indentation

k8s/upgrade/src/upgrade_resources/upgrade.rs
```diff
@@ -599,6 +607,7 @@ impl UpgradeResources {
                         Actions::Create,
                         skip_data_plane_restart,
                         upgrade_to_branch,
+                        upgrade_tag_for_test_envt,
```
tiagolobocastro, line 610: When is `upgrade_to_branch` used?

k8s/upgrade/src/upgrade_resources/upgrade.rs
tiagolobocastro: > Is it a good idea to have a new variable for test environment inside production code? Is it normal?
No, it's not usually a good idea... though not sure exactly here, may be no other way?

As an alternative, when we built these timestamped images, can't we just set the target to the timestampeped tag?

sinhaashish: modified the code . now the user can directly pass the tag , rather than appending the time stamp.

k8s/upgrade/src/upgrade_resources/upgrade.rs
sinhaashish: it is used to upgrade to a branch , for testing , so this flag is hidden too. Right now with 2.1.0 release you can only upgrade from 2.0.0 or 2.0.1 to 2.1.0 . but this feature is there keeping future options in mind.

k8s/upgrade/src/constant.rs
```diff
@@ -37,15 +37,20 @@ pub(crate) fn upgrade_obj_suffix() -> String {
 }
 
 /// Fetch the image tag for upgrade job's pod.
-pub(crate) fn get_image_version_tag() -> String {
-    let version = release_version();
-    match version {
-        Some(upgrade_job_image_tag) => upgrade_job_image_tag,
-        None => UPGRADE_JOB_IMAGE_TAG.to_string(),
+pub(crate) fn get_image_version_tag(image_tag: Option<&String>) -> String {
```
niladrih, line 40: nit: maybe rename function to 'job_image_tag'

k8s/upgrade/src/constant.rs
```diff
@@ -37,15 +37,20 @@ pub(crate) fn upgrade_obj_suffix() -> String {
...
+            match version {
+                Some(upgrade_job_image_tag) => upgrade_job_image_tag,
+                None => UPGRADE_JOB_IMAGE_TAG.to_string(),
+            }
+        }
     }
 }
 
-/// Returns the git tag version (if tag is found) or simply returns the commit hash (12 characters).
+/// Returns the git tag version (if tag is found)
```
niladrih, line 53: nit: Missing sentence's ending fullstop.

k8s/upgrade/src/upgrade_resources/upgrade.rs
```diff
@@ -97,6 +98,10 @@ pub struct UpgradeArgs {
     /// Upgrade to the specified branch.
     #[clap(global = true, hide = true, long, short = 'u')]
     pub upgrade_to_branch: Option<String>,
+
+    /// Upgrade tag for testing, contains timestamp.
```
niladrih, line 102: nit: The testing use-case is of no relevance to this codebase. The doc comment should talk about what this option does for this codebase, i.e. overrides the container image tag for the upgrade job pod.

niladrih (approved):

PR #229, opened 2023-05-11
feat(upgrade): add unsupported upgrade version file upgrade

sinhaashish:
This PR incorporates the below changes:
- add a file for unsupported version
- This file is read by both plugin and upgrade job for upgrade path validation 
- since the to version is already fixed, the from/source version is validated against the unsupported version list.
- In order to upgrade to develop , use the flag `--skip-upgrade-path-validation-for-unsupported-version`
- return the error code mapped to errors.

**Test case 1**
 Install version `2.0.0` and add `2.0.0` in unsupported_versions.yaml as shown below
```
unsupported_versions:
  - 2.0.0
  ```
Upgrade fails stating 
```
ashish@vm:~/code/rust/mayastor-extensions(upgradevalidation-using-file)$ ./target/debug/kubectl-mayastor  upgrade --dry-run

The upgrade path is not valid. The source version is in the list of unsupported versions:
 2.0.0
```
**Test case 2**
Using the `skip-upgrade-path-validation-for-unsupported-version ` so that upgrade against can be tested.
```
./target/debug/kubectl-mayastor  upgrade --skip-upgrade-path-validation-for-unsupported-version --dry-run

List of control plane pods which will be restarted during upgrade.
 mayastor-agent-core-56cc7fc757-zzjqv
mayastor-api-rest-6f96bbb988-tf8c2

List of data plane pods which will be restarted during upgrade.
 mayastor-io-engine-crpcx
mayastor-io-engine-k6h4n
mayastor-io-engine-ml5wn

Finally the cluster deployment will be upgraded to version : develop
```

**Test case3**
Upgrade to develop
Run `/target/debug/kubectl-mayastor upgrade --skip-upgrade-path-validation-for-unsupported-version`
Upgrade succeeds
```
ashish@vm:~/code/rust/mayastor-extensions(upgradevalidation-using-file)$ ./target/debug/kubectl-mayastor get upgrade-status
Upgrade From: 2.0.0
Upgrade To: 0.0.0
Upgrade Status: Successfully upgraded Mayastor
```

### CODE REVIEW

File: k8s/plugin/src/main.rs [hunk 1]
```diff
@@ -151,12 +151,13 @@ async fn execute(cli_args: CliArgs) {
             }
             Operations::Upgrade(resources) => {
                 let _ignore = preflight_validations::preflight_check(
-                    &cli_args.namespace,
+                    &cli_args.namespace.clone(),
```
tiagolobocastro, line 154: why do we need to clone if we're providing a ref?

k8s/upgrade-job/src/helm/upgrade.rs
```diff
@@ -134,18 +153,35 @@ impl HelmUpgradeBuilder {
...
         // Validating upgrade path.
+        let mut invalid_upgrade_path: PathBuf;
+
+        match self.upgrade_path_dir {
+            Some(up_path) => invalid_upgrade_path = up_path,
+            None => {
+                return FilePathNotPresent.fail();
+            }
+        }
+        invalid_upgrade_path.push("unsupported_versions.yaml");
```
tiagolobocastro, line 164:
```suggestion
        let invalid_upgrade_path = match self.upgrade_path_dir {
            Some(path) => Ok(path.join("unsupported_versions.yaml")),
            None => Err(Error::FilePathNotPresent),
        }?;
```
Also, should "unsupported_versions.yaml" be a const?

k8s/upgrade-job/src/opts.rs
```diff
@@ -41,10 +41,18 @@ pub(crate) struct CliArgs {
     )]
     core_chart_dir: Option<PathBuf>,
 
+    /// This is the Helm chart directory filepath for the core Helm chart variant.
+    #[arg(long, env = "UPGRADE_EXCEPTION_PATH", value_name = "DIR PATH")]
+    upgrade_exception_dir: Option<PathBuf>,
```
tiagolobocastro, line 46: why not specify the file path, rather than the dir? And have the default value here so it's clear?

k8s/upgrade-job/src/upgrade/path.rs
```diff
@@ -72,3 +65,27 @@ pub(crate) fn version_from_release_chart(chart_name: String) -> Result<Version>
         version_string: version.to_string(),
     })
 }
+
+#[derive(Debug, Serialize, Deserialize)]
+struct UnspportedVesrsions {
+    unsupported_versions: Vec<String>,
+}
+
+pub(crate) fn get_invalid_upgrade_path(path: PathBuf) -> Result<HashSet<Version>> {
```
tiagolobocastro, line 74:
```suggestion
pub(crate) fn invalid_upgrade_path(path: PathBuf) -> Result<HashSet<Version>> {
```

Missing doc comments

k8s/upgrade-job/src/upgrade/path.rs
```diff
@@ -72,3 +65,27 @@ pub(crate) fn version_from_release_chart(chart_name: String) -> Result<Version>
         version_string: version.to_string(),
     })
 }
+
+#[derive(Debug, Serialize, Deserialize)]
+struct UnspportedVesrsions {
```
tiagolobocastro, line 70:
```suggestion
struct UnsupportedVersions {
```

k8s/upgrade/src/constant.rs
```diff
@@ -74,9 +67,9 @@ pub(crate) const HELM_RELEASE_NAME_LABEL: &str = "openebs.io/release";
 
 pub(crate) const DEFAULT_IMAGE_REGISTRY: &str = "docker.io";
 /// The upgrade job will use the UPGRADE_JOB_IMAGE_NAME image (below) with this tag.
-pub(crate) const UPGRADE_JOB_IMAGE_TAG: &str = "develop";
+pub(crate) const UPGRADE_JOB_IMAGE_TAG: &str = "won";
 /// Upgrade job container image repository.
-pub(crate) const UPGRADE_JOB_IMAGE_REPO: &str = "openebs";
+pub(crate) const UPGRADE_JOB_IMAGE_REPO: &str = "sinhaashish";
```
tiagolobocastro, line 72: please revert these

nix/pkgs/images/default.nix
```diff
@@ -49,7 +50,7 @@ let
     chmod -w build/chart
     chmod -w build/chart/*.yaml
 
-    mkdir -p $out && cp -drf --preserve=mode build/chart $out/chart
+    mkdir -p $out && cp -drf --preserve=mode build/chart $out/chart && cp -drf ${upgrade_path}/* $out/k8s
```
tiagolobocastro, line 53: do we need to add "--preserve=mode" ?

k8s/upgrade-job/src/helm/upgrade.rs
sinhaashish: passing the complete file path now.

k8s/upgrade-job/src/opts.rs
sinhaashish: passing the complete file path now.

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -157,3 +164,62 @@ pub async fn is_rebuild_in_progress(client: &upgrade_lib::RestClient) -> Result<
...
+
+    if destination_version.contains("develop") {
+        console_logger::error("", user_prompt::UPGRADE_TO_UNSUPPORTED_VERSION);
+        return error::InvalidUpgradePath.fail();
+    }
+    Ok(())
+}
+
+/// Return the unsupported version by parsing the unsupported_versions.yaml file.
+pub async fn get_unsupported_versions(path: PathBuf) -> error::Result<HashSet<Version>> {
```
tiagolobocastro, line 208: doesn't need to be async?

k8s/upgrade-job/src/helm/upgrade.rs
```diff
@@ -134,18 +153,30 @@ impl HelmUpgradeBuilder {
         }
 
         // Validating upgrade path.
+        let invalid_upgrade_path = match self.upgrade_path_file {
```
tiagolobocastro, line 156: Can we not default it in the cli argument to our file?

k8s/upgrade/config/unsupported_versions.yaml
```diff
@@ -0,0 +1,3 @@
+unsupported_versions:
+  # add the list of unsupported versions as shown below
+  - 0.0.0
```
tiagolobocastro, line 3: Should we add v1 for completeness?
Do we support "wild cards" here btw?

tiagolobocastro (approved):

k8s/upgrade-job/src/helm/upgrade.rs
sinhaashish: done

k8s/upgrade/config/unsupported_versions.yaml
sinhaashish: no wild cards support , the version in chart.yal is 0.0.0 so v1 not added here

tiagolobocastro: might be worth adding that, even as a follow-up PR

k8s/upgrade-job/src/opts.rs
```diff
@@ -41,10 +41,22 @@ pub(crate) struct CliArgs {
     )]
     core_chart_dir: Option<PathBuf>,
 
+    /// This is the path to upgrade exception file.
+    #[arg(
+        long,
+        env = "UPGRADE_EXCEPTION_FILE_PATH",
+        default_value = "/k8s/upgrade/config/unsupported_versions.yaml"
```
niladrih, line 48: nit: default_value is left for backwards compatibility. Better to use default_value_t.

This value should be in the constants file... and the constant should be used here.

niladrih (approved):

sinhaashish:
bors merge

PR #231, opened 2023-05-15
feat(upgrade-job): support for upgrade to openebs-3.7.0

niladrih:
- remove support for upgrade for the openebs/openebs helm chart:
Previously the upgrade-job was outfitted with the tools to upgrade the
openebs/openebs helm chart. However, the upgrade assumed that the helm
chart only made use of mayastor and no other storage engine. This is a
convenient feature, until one of the other openebs/openebs subcharts
is upgraded. Then this openebs/openebs upgrade via this upgrade-job is
insufficient, as it only concerns itself with the changes of the mayastor
subchart. As an alternative, the upgrade documentation at openebs.io/docs
will furnish the instructions to enable a user to carry out the mayastor
upgrade alongside any other upgrade that the particular openebs/openebs
might require. The user is to upgrade the openebs/openebs helm chart, and
iff he has done so, this upgrade-job proceeds to carry out the 'data plane
upgrade', thus reaching an upgraded state for mayastor.
- add helm chart value 'set' flags for v2.1.0 and v2.2.0
- refactor upgrade to get the source helm chart version from the
REST API deployment label 'openebs.io/version' instead of the 'helm ls'
output
- remove the HashSet from the upgrade::path::UnsupportedVersions
and use the 'contains()' from the Vec instead.
- remove priorityClassName 'set' flags, as users with working global
priorityClassName should be able to just continue using it, instead of
being moved from the older default to the latest default"

### DIFF
--- a/k8s/upgrade-job/src/helm/upgrade.rs
+++ b/k8s/upgrade-job/src/helm/upgrade.rs
@@ -1,9 +1,10 @@
 use crate::{
     common::{
-        constants::{CORE_CHART_NAME, UMBRELLA_CHART_NAME},
+        constants::{CORE_CHART_NAME, TO_UMBRELLA_SEMVER, UMBRELLA_CHART_NAME},
         error::{
-            HelmUpgradeOptionsAbsent, InvalidUpgradePath, NoInputHelmChartDir, NotAKnownHelmChart,
-            RegexCompile, Result,
+            CoreChartUpgradeNoneChartDir, HelmUpgradeOptionsAbsent, InvalidHelmUpgrade,
+            InvalidUpgradePath, NoInputHelmChartDir, NotAKnownHelmChart, RegexCompile, Result,
+            UmbrellaChartNotUpgraded,
         },
     },
     helm::{client::HelmReleaseClient, values::generate_values_args},
@@ -30,14 +31,13 @@ pub(crate) enum HelmChart {
 pub(crate) struct HelmUpgradeBuilder {
     release_name: Option<String>,
     namespace: Option<String>,
-    umbrella_chart_dir: Option<PathBuf>,
     core_chart_dir: Option<PathBuf>,
     upgrade_path_file: PathBuf,
     skip_upgrade_path_validation: bool,
 }
 
 impl HelmUpgradeBuilder {
-    /// This is a builder option to add the Namespace of the helm chart to be upgrade.
+    /// This is a builder option to add the Namespace of the helm chart to be upgraded.
     #[must_use]
     pub(crate) fn with_namespace<J>(mut self, ns: J) -> Self
     where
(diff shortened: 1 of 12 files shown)


niladrih:
bors try

k8s/upgrade-job/src/helm/upgrade.rs
```diff
@@ -88,109 +81,108 @@ impl HelmUpgradeBuilder {
...
+
+        // The version of the Core chart which we are (maybe) going to.
+        let chart_dir: PathBuf = self.core_chart_dir.ok_or(
+            NoInputHelmChartDir {
+                chart_name: CORE_CHART_NAME.to_string(),
+            }
+            .build(),
+        )?;
+        let mut chart_yaml_path = chart_dir.clone();
+        chart_yaml_path.push("Chart.yaml");
```
tiagolobocastro, line 114:
```suggestion
        let chart_yaml_path = chart_dir.join("Chart.yaml");
```

tiagolobocastro: Also, why not make "core_chart_dir" "core_chart_path"? So we don't even have to append the Chart.yaml

k8s/upgrade-job/src/helm/upgrade.rs
```diff
@@ -88,109 +81,108 @@ impl HelmUpgradeBuilder {
...
             format!(r"^({CORE_CHART_NAME}-[0-9]+\.[0-9]+\.[0-9]+(-rc\.[0-9]+)?)$");
 
-        // Assign HelmChart variant and validate directory path input for the said
-        // variant's chart based on the 'chart' member of the HelmReleaseElement.
+        // Validate if already upgraded for Umbrella chart, and prepare for upgrade for Core chart.
         let chart_variant: HelmChart;
-        let chart_dir: PathBuf;
-        // Case: HelmChart::Umbrella.
-        if Regex::new(umbrella_chart_regex.as_str())
+        let mut already_upgraded: bool = false;
```
tiagolobocastro, line 126:
```suggestion
        let mut already_upgraded = false;
```

k8s/upgrade-job/src/helm/upgrade.rs
```diff
@@ -88,109 +81,108 @@ impl HelmUpgradeBuilder {
...
+                    InvalidUpgradePath
+                );
+            }
 
-        if !self.skip_upgrade_path_validation {
-            let upgrade_path_is_valid = upgrade::path::is_valid(
-                chart_variant.clone(),
+            // Generate args to pass to the `helm upgrade` command.
+            let mut values_yaml_path = chart_dir.clone();
+            values_yaml_path.push("values.yaml");
```
tiagolobocastro, line 164: Isn't set already set above, though we did None it?

k8s/upgrade-job/Cargo.toml
```diff
@@ -23,3 +23,4 @@ serde_json = "1.0.93"
 serde_yaml = "0.9.17"
 serde = { version = "1.0.152", features = ["derive"] }
 semver = { version="1.0.17", features = ["serde"] }
+const_format = "0.2.30"
```
tiagolobocastro, line 26: nit: I'm tempted to say just use normal format, we don't have any high performance runtime constraints here, so it's better to reduce compile times.

niladrih:
bors try

tiagolobocastro (approved):

k8s/upgrade-job/src/helm/upgrade.rs
niladrih: We're basically taking the path to the helm chart directory as input, and poking around in multiple files...

k8s/upgrade-job/src/helm/upgrade.rs
niladrih: No... it isn't. Didn't get you.

k8s/upgrade-job/Cargo.toml
niladrih: Blazingly fast 😎, with those two literals.
Okay, I'll get rid of them.

sinhaashish (approved):

sinhaashish (approved):

sinhaashish (approved):

niladrih:
bors merge

PR #232, opened 2023-05-16
feat(snapshots): add snapshot crds, snapshot-controller and external snapshotter side car containers.

Abhinandan-Purkait:
- Adds VolumeSnapshot, VolumeSnapshotContent and VolumeSnapshotClass crds.
- Adds external-snapshotter and snapshot-controller as sidecar containers to csi-controller.
- Adds resource limit and requirements for snapshot-controller container.


niladrih (commented):
Have we tested this in a cluster where these CRDs already exist? Will the helm install fail, or is it idempotent?

Abhinandan-Purkait:
> Have we tested this in a cluster where these CRDs already exist? Will the helm install fail, or is it idempotent?

Yes, have tested this with clusters where these CRDs already exists and helm install doesn't fail.

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #233, opened 2023-05-16
feat(upgrade): read unsupported versions yaml file at compile time

sinhaashish:
The `unsupported_versions.yaml` file was read using the std::current_dir(). This lead to the problem where in the file was not present in nix-build environment. 
This PR modifies the code to read the file at compile time only .

### DIFF
--- a/k8s/upgrade/src/preflight_validations.rs
+++ b/k8s/upgrade/src/preflight_validations.rs
@@ -1,15 +1,15 @@
 use crate::{
-    constant::{get_image_version_tag, SINGLE_REPLICA_VOLUME, UNSUPPORTED_VERSION_FILE},
+    constant::{get_image_version_tag, SINGLE_REPLICA_VOLUME, UPGRADE_TO_DEVELOP_BRANCH},
     error, upgrade_lib,
     upgrade_resources::upgrade::{get_pvc_from_uuid, get_source_version},
     user_prompt,
 };
 use openapi::models::CordonDrainState;
 use semver::Version;
-use serde::{Deserialize, Serialize};
+use serde::Deserialize;
 use serde_yaml;
 use snafu::ResultExt;
-use std::{collections::HashSet, fs::File, path::PathBuf};
+use std::{collections::HashSet, path::PathBuf};
 
 /// Validation to be done before applying upgrade.
 pub async fn preflight_check(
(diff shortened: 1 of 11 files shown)


k8s/upgrade/src/preflight_validations.rs
```diff
@@ -168,15 +168,29 @@ pub async fn is_rebuild_in_progress(client: &upgrade_lib::RestClient) -> error::
...
+    /// Returns an UnsupportedVersions object.
+    fn try_from(bytes: &[u8]) -> error::Result<Self> {
+        let mut path = PathBuf::new();
+        path.push(UNSUPPORTED_VERSION_FILE);
+        serde_yaml::from_reader(bytes).context(error::YamlParseFromFile { filepath: path })
+    }
+}
+
+pub async fn upgrade_path_validation(namespace: &str) -> error::Result<()> {
+    let unsupported_version_file = &std::include_bytes!("../config/unsupported_versions.yaml")[..];
```
niladrih, line 192: nit: This variable is misleading. This is not a std::fs::File. Maybe rename to unsupported_version_buf? As in... it is a bytes buffer.

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -185,7 +199,7 @@ pub async fn upgrade_path_validation(namespace: &str) -> error::Result<()> {
 
     if unsupported_versions.contains(&source) {
         let mut invalid_source_list: String = String::new();
```
niladrih, line 201: nit: I think Default::default() makes it a bit more obvious instead of String::new().

niladrih (commented):
Small nitpick comments..

k8s/upgrade/src/preflight_validations.rs
niladrih, line 1:
```rust

    if destination_version.contains("develop") {
        console_logger::error("", user_prompt::UPGRADE_TO_UNSUPPORTED_VERSION);
        return error::InvalidUpgradePath.fail();
    }
```

The "develop" should be a constant maybe?

niladrih (commented):
Left one comment

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -168,15 +168,29 @@ pub async fn is_rebuild_in_progress(client: &upgrade_lib::RestClient) -> error::
...
+}
+
+impl TryFrom<&[u8]> for UnsupportedVersions {
+    type Error = crate::error::Error;
 
-    let unsupported_versions = get_unsupported_versions(path)?;
+    /// Returns an UnsupportedVersions object.
+    fn try_from(bytes: &[u8]) -> error::Result<Self> {
+        let mut path = PathBuf::new();
+        path.push(UNSUPPORTED_VERSION_FILE);
```
tiagolobocastro, line 186: let path = Path::new(UNSUPPORTED_VERSION_FILE).to_path_buf();

tiagolobocastro: Actually, this is a bit misleading here, because bytes is just an array of bytes, I think the context should be added by the calling function!

tiagolobocastro (commented):
I think we also need to revert the previous nix changes?

k8s/upgrade/src/preflight_validations.rs
sinhaashish: IMHO, `serde_yaml::from_reader` is reading the bytes and any error occurs while parsing should be thrown. Since the parsing is of the unsupported_version.yaml file the error is thrown about the same.

tiagolobocastro: it would be thrown, but it would be the original error from serde_yaml... you need only attach the error you want with the path at the calling side. 
Let say we try to load a different file into UnsupportedVersions, the error message is going to report the wrong path!

k8s/upgrade/src/preflight_validations.rs
```diff
@@ -168,15 +168,29 @@ pub async fn is_rebuild_in_progress(client: &upgrade_lib::RestClient) -> error::
 /// Struct to deserialize the unsupported version yaml.
 #[derive(Debug, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
sinhaashish: done

sinhaashish:
> I think we also need to revert the previous nix changes?

changes made

sinhaashish:
bors try

niladrih (approved):

tiagolobocastro (approved):

sinhaashish:
bors merge

PR #235, opened 2023-05-17
fix(upgrade): update upgrade-job-image-tag

sinhaashish:


k8s/upgrade/src/constant.rs
```diff
@@ -67,7 +67,7 @@ pub(crate) const HELM_RELEASE_NAME_LABEL: &str = "openebs.io/release";
 
 pub(crate) const DEFAULT_IMAGE_REGISTRY: &str = "docker.io";
 /// The upgrade job will use the UPGRADE_JOB_IMAGE_NAME image (below) with this tag.
-pub(crate) const UPGRADE_JOB_IMAGE_TAG: &str = "develop";
+pub(crate) const UPGRADE_JOB_IMAGE_TAG: &str = "release-2.2";
```
tiagolobocastro, line 70: hmm what about when we have a proper release, do we need to change this to a tag?

niladrih: Yes.. please note the functions release_version() and get_image_version_tag() above..

niladrih: nit: maybe the constant variable name could have a DEFAULT_ prefix, so that it's clear that it's a fallback value.

tiagolobocastro: But why do we need these at all?

niladrih: Ok... here is the workflow for these values...
higher number in () means higher priority..
```
                                                                                   ┌───────────────────────────────┐
┌───────────────────┐                     ┌─────────────────────────┐              │                               │
│(10) Tagged commit ├────────────┐    ┌───┤ upgrade container image ├──────────────┤         upgrade Pod           │
└───────────────────┘            │    │   └─────────────────────────┘              │                               │
                                 │    │                                      ┌─────┤                               │
                                 │    │                                      │     └───────────────────────────────┘
                                 ├────┤                                      │
                                 │    │                                      │     ┌───────────────────────────────┐
┌────────────────────────┐       │    │    ┌────────────────────────────┐    │     │                               │
│(1) Not-a-tagged commit ├───────┘    └────┤upgrade k8s resources'suffix├────┤     │         upgrade RBAC          │
└────────────────────────┘                 └────────────────────────────┘    └─────┤                               │
                                                                                   │                               │
                                                                                   └───────────────────────────────┘
```

tiagolobocastro: I think this will go away if we start using semver throughout the project.. so for example, we'd be pushing io-engine:develop as io-engine:v0.0.0 or io-engine:v2.3.0-develop.
But meanwhile, why don't we just use make TARGET_BRANCH pub in dependencies/control-plane/utils/utils-lib/src/test_constants.rs and just use it, it will be correct for whichever branch you're in.

niladrih: Okay... Let me take a look..

sinhaashish: This PR wont be needed as now we are passing the `--tag`.

PR #237, opened 2023-05-17
fix(upgrade) : use custom tag for images

sinhaashish:


niladrih:
I don't think we should change the image tag from what is in the bundled helm chart.

Wdyt @avishnu @tiagolobocastro

tiagolobocastro:
> I don't think we should change the image tag from what is in the bundled helm chart.
> 
> Wdyt @avishnu @tiagolobocastro

Yes I think so too..

k8s/upgrade-job/src/helm/values.rs
```diff
@@ -107,7 +108,14 @@ pub(crate) fn generate_values_args(
...
 
-    let image_tag_arg = format!("{image_key}.{tag_key}={}", to_values.image_tag());
+    let image_tag_arg = match custom_image_tag {
+        Some(image_tag) => {
+            format!("{image_key}.{tag_key}={}", image_tag.as_str())
+        }
+        None => {
+            format!("{image_key}.{tag_key}={}", to_values.image_tag())
+        }
+    };
```
Abhinandan-Purkait, line 118:
```suggestion
   let image_tag_arg = format!(
    "{image_key}.{tag_key}={}",
    custom_image_tag.unwrap_or_else(|| to_values.image_tag().as_str())
);
```
Would this be simple?

Abhinandan-Purkait (approved):

niladrih:
This change is no longer needed.

PR #238, opened 2023-05-17
fix(upgrade-job): refactor version check

niladrih:
- add failure case for when the source version is newer than the target version
- remove redundant check for 'already_upgraded' case: currently the invalid versions are explicitly listed in an input YAML file, thus there need not be an additional exclusion case here, as we do not list the current version in the invalid-versions-file


tiagolobocastro (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

PR #242, opened 2023-05-22
fix(upgrade-job): upgrade path validation and set flags

niladrih:
- use `--set` flags to set thin provisioning flags for 2.0.x releases, instead of 2.1.x release.
- allow 'no-rollback' rule to be overridden by the `--skip-upgrade-path-validation` flag. This is not necessary for the openebs/openebs (Umbrella) chart because this upgrade-job does not support upgrade for the openebs chart, so might as well move that check to be a Core chart exclusive check.
- allow the `--set` flags for 2.0.x release to be used with future 2.0.x releases as well, and not just till 2.0.1.


sinhaashish:
why is this required `--app-tag`?

niladrih:
> why is this required `--app-tag`?

Oh, that's not supposted to be there..

niladrih:
bors try

sinhaashish:
is the code for set flag removed now ?

sinhaashish (approved):

tiagolobocastro (approved):

niladrih:
bors merge

PR #243, opened 2023-05-22
feat(call-home): add nats to helm

sinhaashish:


chart/values.yaml
```diff
@@ -587,6 +587,34 @@ loki-stack:
...
+      fileStorage:
+        enabled: false
+  cluster:
+    enabled: true
+    # Cluster name is required, by default will be release name.
+    name: "nats"
+    replicas: 3
+  # The nats box can be installed for debugging
+  natsbox:
+    enabled: true
```
datacore-vvarakantham, line 612: The Nats box need to be set to false.

chart/values.yaml
```diff
@@ -587,6 +587,34 @@ loki-stack:
               - __meta_kubernetes_pod_uid
               - __meta_kubernetes_pod_container_name
               target_label: __path__
+
+nats:
```
niladrih, line 591: Needs comment stating what this nats is for...

chart/values.yaml
```diff
@@ -587,6 +587,34 @@ loki-stack:
...
+        # Size of nats message is around 0.3 KB, so it can store around 10K messages.
+        size: "5Mi"
+      fileStorage:
+        enabled: false
+  cluster:
+    enabled: true
+    # Cluster name is required, by default will be release name.
+    name: "nats"
+    replicas: 3
+  # The nats box can be installed for debugging
```
niladrih, line 610: Why are we using clustering?

chart/values.yaml
tiagolobocastro: Should we have an `eventing` component which in turn enables or disables NATS ?

chart/values.yaml
```diff
@@ -587,6 +587,34 @@ loki-stack:
...
+      memStorage:
+        enabled: true
+        # Size of nats message is around 0.3 KB, so it can store around 10K messages.
+        size: "5Mi"
+      fileStorage:
+        enabled: false
+  cluster:
+    enabled: true
+    # Cluster name is required, by default will be release name.
+    name: "nats"
```
tiagolobocastro, line 608: can we leave it as default to use release name?

chart/values.yaml
tiagolobocastro: Probably because we're using in-memory, so if one pod restarts, the other ones still keep the messages? @datacore-vvarakantham

chart/values.yaml
```diff
@@ -587,6 +587,34 @@ loki-stack:
...
+    replicas: 3
+  # The nats box can be installed for debugging
+  natsbox:
+    enabled: true
+    image:
+      repository: natsio/nats-box
+      pullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
tiagolobocastro: I'd say let's leave it out entirely, it can always be set anyway..

chart/values.yaml
niladrih: Should be have 2 factor redundancy here? We're using the CR/configmap for persistent storage on the cluster etcd, and it's not mission critical..

chart/values.yaml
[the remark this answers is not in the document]
niladrih: It's [set to true by default](https://github.com/nats-io/k8s/blob/v0.13.1/helm/charts/nats/values.yaml), and not sure, but I think that's another running container if we leave it out and it defaults to the chart's default value.

chart/values.yaml
niladrih: Is the 'name' key used? I don't see the option here on the latest version of the 'nats' helm chart. Refer to: https://github.com/nats-io/k8s/blob/v0.13.1/helm/charts/nats/values.yaml#L336

datacore-vvarakantham: Yes. In cluster mode the servers connect to each other. And the cluster self heals if any of the server is down.

Also this is how its described in Nats docs for stream replication.

Replicas=1 - Cannot operate during an outage of the server servicing the stream. Highly performant.
Replicas=2 - No significant benefit at this time. We recommend using Replicas=3 instead.
Replicas=3 - Can tolerate loss of one server servicing the stream. An ideal balance between risk and performance.

chart/values.yaml
sinhaashish: The options under nats are those of nats subchart, so we cant rename this.

chart/values.yaml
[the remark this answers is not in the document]
sinhaashish: yes the nats-box is set to be enabled by default so overriding it in values.yaml

chart/values.yaml
```diff
@@ -587,6 +587,34 @@ loki-stack:
               - __meta_kubernetes_pod_uid
               - __meta_kubernetes_pod_container_name
               target_label: __path__
+
+nats:
+  enabled: true
+  nats:
+    image:
+      repository: nats
+      pullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
sinhaashish: not needed as it will be default and will be constant as version `0.19.14`

chart/values.yaml
sinhaashish: removed

tiagolobocastro (approved):

datacore-vvarakantham (approved):

sinhaashish:
bors merge

sinhaashish:
bors merge

PR #249, opened 2023-05-30
fix(callhome) : apply nats when callhome is enabled

sinhaashish:
A per the suggestion here `IMHO we should disable it if either callhome or "eventing" is disabled.` 
 I tried to tweak the condition like `condition: obs.callhome.enabled and nats.enabled` , but this doesn't work.

Helm does not support the `and ` in conditions . The max is does is an `OR` condition using comma  like `condition: obs.callhome.enabled, nats.enabled`.  Only the first valid path found in the list is evaluated .

Thus enabling nats only when the call home is enabled.

**Refer**
- [Link 1](https://stackoverflow.com/questions/75649506/how-to-add-multiple-conditions-check-in-global-charts-file-in-helm)
- [Link 2](https://helm.sh/docs/topics/charts/#tags-and-condition-fields-in-dependencies)

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -588,10 +588,13 @@ loki-stack:
               - __meta_kubernetes_pod_container_name
               target_label: __path__
 
-# Cloud Native Messaging System
-nats:
-  # -- Enable nats for transmitting events to subscribers
+# Eventing component which enables or disables eventing-related components.
+eventing:
   enabled: true
+
+# Configuration for the nats message-bus. This is an eventing component, and is enabled when
+# 'eventing.enabled' is set to 'true'.
+nats:
   nats:
     image:
       pullPolicy: IfNotPresent
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -39,5 +39,5 @@ dependencies:
   - name: nats
     version: 0.19.14
     repository: https://nats-io.github.io/k8s/helm/charts/
-    condition: nats.enabled
+    condition: eventing.enabled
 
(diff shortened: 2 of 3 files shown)


chriswldenyer (approved):
Looks reasonable

chriswldenyer:
> Should eventing be the one enabling nats? We might have eventing enabled and callhome disabled for example?

Agreed.

sinhaashish:
> Should eventing be the one enabling nats? We might have eventing enabled and callhome disabled for example?

Added eventing in values.yaml

chart/values.yaml
```diff
@@ -588,10 +588,12 @@ loki-stack:
               - __meta_kubernetes_pod_container_name
               target_label: __path__
 
+# -- Eventing component which enables or disables nats
```
tiagolobocastro, line 591: I would drop the "--" for now, not sure this would be useful for user to know nats is enabled or not, nats is an implementation detail.

tiagolobocastro (approved):

chart/values.yaml
```diff
@@ -588,10 +588,12 @@ loki-stack:
               - __meta_kubernetes_pod_container_name
               target_label: __path__
 
+# Eventing component which enables or disables nats
```
niladrih, line 591: nit:
```suggestion
# Eventing component which enables or disables eventing-related components.
```

chart/values.yaml
```diff
@@ -588,10 +588,12 @@ loki-stack:
...
+eventing:
+  enabled: true
+
 # Cloud Native Messaging System
 nats:
```
niladrih, line 596:
```suggestion
# Configuration for the nats message-bus. This is an eventing component, and is enabled when
# 'eventing.enabled' is set to 'true'.
nats:
```

niladrih (commented):
nit:

chart/values.yaml
sinhaashish: Thanks , done

niladrih (approved):
LGTM

datacore-vvarakantham:
LGTM

datacore-vvarakantham (approved):

chriswldenyer (approved):

sinhaashish:
bors merge

sinhaashish:
bors merge

PR #250, opened 2023-06-02
feat(upgrade): merge upgrade-job crate to upgrade

sinhaashish:


sinhaashish:
bors try

sinhaashish:
bors try

sinhaashish:
bors try

k8s/upgrade/src/upgrade_resources/mod.rs
tiagolobocastro, line 1: This file can be deleted now?

sinhaashish: yes, because `upgrade_resources` is not a folder any more. all the code has been removed to `upgrade_plugin`

niladrih (commented):
Why not put the upgrade-job in the /bin directory (ref: https://doc.rust-lang.org/cargo/guide/project-layout.html)?

Also, this change looks like it is moving the upgrade-job into the upgrade directory. Why is this necessary? Should this instead be a refactor, so that the upgrade lib contains all of the shared resources which can be used by upgrade-job and upgrade-plugin? Isn't this the same without this change? What am I missing here?

niladrih (commented):
Why not put the upgrade-job in the /bin directory (ref: https://doc.rust-lang.org/cargo/guide/project-layout.html)?

Also, this change looks like it is moving the upgrade-job into the upgrade directory. Why is this necessary? Should this instead be a refactor, so that the upgrade lib contains all of the shared resources which can be used by upgrade-job and upgrade-plugin? Isn't this the same without this change? What am I missing here?

sinhaashish:
> Why not put the upgrade-job in the /bin directory (ref: https://doc.rust-lang.org/cargo/guide/project-layout.html)?
> 
> Also, this change looks like it is moving the upgrade-job into the upgrade directory. Why is this necessary? Should this instead be a refactor, so that the upgrade lib contains all of the shared resources which can be used by upgrade-job and upgrade-plugin? Isn't this the same without this change? What am I missing here?

I had tried that in the previous version of my PR , wherein the errors and commons were in a common folder and were used by both upgrade job and plugin. That didn't worked , cargo clippy had errors stating that there were some unused code when i compiled the binary i.e. upgrade-job, and if the library were compiled then again there were errors of unused code from binary. 
Thats the reason, upon discussion with @tiagolobocastro  i just merged the two folders into one.
Upgrade functionality under one folder is an ideal situation which this PR is doing. Upgrade job and plugin both inside the upgrade folder.

k8s/upgrade/src/plugin/preflight_validations.rs
```diff
@@ -216,3 +216,26 @@ pub async fn upgrade_path_validation(namespace: &str) -> error::Result<()> {
...
+        &self.client
+    }
+}
+
+impl RestClient {
+    /// Create new Rest Client from the given `Configuration`.
+    pub fn new_with_config(config: Configuration) -> RestClient {
+        Self {
+            client: tower::ApiClient::new(config),
+        }
```
niladrih, line 239: Why is this change required?

niladrih (commented):
Left one comment.

k8s/upgrade/src/plugin/preflight_validations.rs
niladrih: Oh, okay

niladrih (approved):

sinhaashish:
bors merge

sinhaashish:
bors merge

PR #259, opened 2023-06-19
Adjust jenkins workspace path in case "@" in present in path

pchandra19:
This PR , modifies the workspace name in case "@" is present in the jenkins workspace dir.
I am not sure, if this will actually update the WORKSPACE env variable. I am hoping the next time it picks up the workspace it should point to the new one.
I did try to test this a few times, but since I was unable to reproduce the issue, It will have to be tested once this live and then adjusted if needed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -90,8 +90,9 @@ pipeline {
         printenv
         if [[ ${env.WORKSPACE} =~ "@" ]];
         then
-        newpath=$(echo ${env.WORKSPACE} | sed s/"@"/"__"/)
-        mv ${env.WORKSPACE} newpath
+          newpath=$(echo ${env.WORKSPACE} | sed s/"@"/"__"/)
+          mv ${env.WORKSPACE} newpath
+        fi
         printenv
         """
       }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -91,7 +91,7 @@ pipeline {
         if [[ "${env.WORKSPACE}" =~ "@" ]];
         then
             newpath=\$(echo ${env.WORKSPACE} | sed s/"@"/"__"/)
-            mv "${env.WORKSPACE}" $newpath
+            mv "${env.WORKSPACE}" "${newpath}"
             echo "Inside if"
         else
             echo "Not needed"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -82,6 +82,24 @@ pipeline {
...
+    stage('Rename jenkins workspace path') {
+      agent {label 'nixos-mayastor'}
+      steps{
+        sh """
+        printenv
+        if [[ "${env.WORKSPACE}" =~ "@" ]];
+        then
+            newpath=\$(echo ${env.WORKSPACE} | sed s/"@"/"__"/)
+            mv "${env.WORKSPACE}" newpath
+            echo "Inside if"
```
blaisedias, line 95: I think you should set `env.WORKSPACE`  to `newpath` for this to work properly

pchandra19: I had that doubt , whether I do that or change the dir name on disk. The problem is I wouldn't know unless the PR is merged.

Jenkinsfile
```diff
@@ -82,6 +82,24 @@ pipeline {
...
+
+    stage('Rename jenkins workspace path') {
+      agent {label 'nixos-mayastor'}
+      steps{
+        sh """
+        printenv
+        if [[ "${env.WORKSPACE}" =~ "@" ]];
+        then
+            newpath=\$(echo ${env.WORKSPACE} | sed s/"@"/"__"/)
+            mv "${env.WORKSPACE}" $newpath
```
tiagolobocastro, line 94: Seems a bit "dangerous" to do this, might override something?
Isn't it better to change container build itself (in this repo) to support "@" ?

pchandra19: yes , if that can be handled in build code nothing better. So should I close this PR then ?
Also, any idea who can fix it from a build code level. I will create a JIRA and assign to them ?

tiagolobocastro: It can also be you if you like nix :) otherwise maybe @niladrih would like to take it?

pchandra19:
Also, I tried to test by simulating a test in my sandbox and I think this approach will not work. I am closing the PR

PR #266, opened 2023-06-27
fix(upgrade-job): use '--install' flag with helm upgrade

niladrih:
The upgrade-job issues a helm upgrade command which in itself does not install CRDs. The changes for release v2.3.0 introduces the snapshot CRDs to the mayastor helm chart. Using the --install flag will add it into the upgrade template if it is missing from the cluster


niladrih:
bors try

w3aman (approved):
tried upgrade with this branch..it works fine...csi-controller pods came into running state.

avishnu (approved):
lgtm

niladrih:
bors merge

PR #273, opened 2023-06-29
feat(eventing): add transmission module

sinhaashish:


call-home/src/common/constants.rs
```diff
@@ -102,7 +111,7 @@ pub const RECEIVER_ENDPOINT: &str = "https://openebs.phonehome.datacore.com/open
 const CALL_HOME_FREQUENCY_IN_HOURS: i64 = 24;
 
 pub fn call_home_frequency() -> std::time::Duration {
-    chrono::Duration::hours(CALL_HOME_FREQUENCY_IN_HOURS)
+    chrono::Duration::seconds(CALL_HOME_FREQUENCY_IN_HOURS)
```
tiagolobocastro, line 114: I don't think we want to change this?

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -184,3 +215,93 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
...
+    pub fn new(data: Vec<Sample>) -> Self {
+        Self { record: data }
+    }
+
+    /// Maps the prometheus data to EventData
+    pub fn event_record(&self) -> errors::Result<EventData> {
+        let mut event_record = EventData::default();
+        let mut iterator = self.record.iter();
+        if let Some(rec) = iterator.next() {
+            let val = rec.labels.get("action").unwrap();
```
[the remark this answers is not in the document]
sinhaashish: ` pub fn event_record(&self) -> errors::Result<EventData>` actually populates `EventData`. which is used every where . Having separate function to get the required values will repeat the code. The better implementation is done in my code , wherein the rec is parsed and EventData is populated once from it.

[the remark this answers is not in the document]
tiagolobocastro: But you already have separate functions here to get action and component.
Having the new_type wrapper won't add a separate function, instead it will combine the impl of these in a neat place.
https://doc.rust-lang.org/rust-by-example/generics/new_types.html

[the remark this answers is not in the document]
sinhaashish: done

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -184,3 +199,160 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
...
+    }
+
+    /// Returns VolumeCreated.
+    pub fn volume_created(&self) -> VolumeCreated {
+        let mut vc = 0;
+        for rec in self.record.iter() {
+            let value = rec.labels.get(ACTION).unwrap_or_default();
+            if value == CREATED && rec.metric.as_str() == VOLUME {
+                vc = stats(rec.value.clone());
+                break;
```
tiagolobocastro, line 276: hmm don't think you can break here, what if other records are also volume?
I suggest something like this:
```rust
struct Record(extern_type);
impl Record {
   fn action(&self) -> ACTION {
     self.labels.get(ACTION).unwrap_or_default()
   }
   fn metric(&self) -> Option<stats::MetricType> {
     self.metric.as_str().try_into().ok()
   }
   fn gauge_value(&self) -> u32 { 
     match val {
        Value::Gauge(v) => v as u32,
        _ => 0,
     }
   }
}
impl TryFrom<Record> for VolumeCreated {
   type Error=();
   fn try_from(src: &Record) -> Result<Self, ()> {
      match src.metric() {
       Some(stats::Volume) => {
         Ok(VolumeCreated(src.gauge_value())
       }
       _ => None,
      }
   }
}
impl<T: TryFrom<Record>> From<Vec<Record>> for Vec<T> {
  fn from(src: &Vec<Record>) -> Vec<VolumeCreated> {
    self.iter().filter_map(|r| r.try_from().ok()).collect::Vec<_>()
  }
}

fn data_thingy() {
let volumes_created: Vec<VolumesCreated> = records.into();
let pools_created: Vec<PoolsCreated> = records.into();
}
```

call-home/src/bin/callhome/main.rs
```diff
@@ -91,9 +96,12 @@ async fn run() -> anyhow::Result<()> {
             k8s_cluster_id.clone(),
             namespace.clone(),
             version.clone(),
+            transmission_url.clone(),
         )
         .await;
 
+        tracing::info!(" report is {:?}", report);
```
tiagolobocastro, line 103:
```suggestion
```

GUess this was for debug only?

call-home/src/bin/callhome/main.rs
```diff
@@ -129,6 +138,19 @@ async fn generate_report(
...
+    let mut event_data = EventData::default();
+    match event_stats(transmission_url.clone()).await {
+        Ok(data) => {
+            event_data.volume_created = data.volume_created();
+            event_data.volume_deleted = data.volume_deleted();
+            event_data.pool_created = data.pool_created();
+            event_data.pool_deleted = data.pool_deleted();
+        }
+        Err(err) => {
+            error!("{:?}", err);
```
tiagolobocastro, line 150: Might want to qualify what went wrong specifically

sinhaashish: The same is just printed in this file for `match nodes` , `match pools`, ` match volumes` ., on the very same line its done in PR.

call-home/src/bin/callhome/collector/report_models.rs
sinhaashish: Thanks for the help @tiagolobocastro . addressed your comment

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -184,3 +199,217 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
     }
     pools_size_vector
 }
+
+/// EventData consists of pool and volume events.
+#[derive(Debug, Default, Clone)]
+pub struct EventData {
```
tiagolobocastro, line 205: This doesn't need to be pub right, pub(crate) at most?

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -184,3 +199,217 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
...
+    // Method to retrieve the value
+    fn value(&self) -> u32 {
+        self.0
+    }
+}
+
+/// Record of events populated from prometheus.
+#[derive(Debug, Default, Clone)]
+/// Wrapper for Stats fetched from prometheus data.
+pub struct EventsRecord {
```
tiagolobocastro, line 259: same heres

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -184,3 +199,217 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
...
+    pub record: Vec<Sample>,
+}
+
+/// Implements methods for wrapper.
+impl EventsRecord {
+    pub(crate) fn new(data: Vec<Sample>) -> Self {
```
tiagolobocastro, line 265: missing doc comment

call-home/src/bin/callhome/main.rs
```diff
@@ -29,6 +30,10 @@ struct CliArgs {
     /// The namespace we are supposed to operate in.
     #[arg(short, long, default_value = DEFAULT_NAMESPACE)]
     namespace: String,
+
+    /// The rest endpoint to connect to.
```
tiagolobocastro, line 34: wrong doc comment?

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -184,3 +199,217 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
...
+    }
+}
+
+impl<T: TryFrom<Record>> From<&EventsRecord> for Option<T> {
+    fn from(src: &EventsRecord) -> Option<T> {
+        src.record
+            .iter()
+            .find_map(|r| T::try_from(Record(r.clone())).ok())
+    }
+}
```
tiagolobocastro, line 415: I think we can simplify this a lot by using a macro to generate the counters.
Also let's add an enum for Action and perhaps rename Metrics to Category (I'm on the fence on rename so either is fine)? 

```suggestion
struct Record<'a>(&'a Sample);
impl Record<'_> {
    fn action(&self) -> Action {
        Action::from(self.0.labels.get(ACTION).unwrap_or(""))
    }
    fn category(&self) -> Category {
        Category::from(self.0.metric.as_str())
    }
    fn gauge_value(&self) -> u32 {
        match self.0.value {
            Value::Gauge(v) => v as u32,
            _ => 0,
        }
    }
    fn try_gauge(&self, category: Category, action: Action) -> Result<GaugeCounter, ()> {
        if self.category() == category && self.action() == action {
            return Ok(GaugeCounter(self.gauge_value()));
        }
        Err(())
    }
}

/// Metrics category.
#[derive(Debug, Default, Copy, Clone, Eq, PartialEq)]
pub enum Category {
    #[default]
    Unknown,
    Pool,
    Volume,
}
/// Action category.
#[derive(Debug, Default, Copy, Clone, Eq, PartialEq)]
pub enum Action {
    #[default]
    Unknown,
    Created,
    Deleted,
}

impl From<&str> for Category {
    fn from(s: &str) -> Self {
        match s {
            "pool" => Category::Pool,
            "volume" => Category::Volume,
            _ => Category::Unknown,
        }
    }
}
impl From<&str> for Action {
    fn from(s: &str) -> Self {
        match s {
            "created" => Action::Created,
            "deleted" => Action::Deleted,
            _ => Action::Unknown,
        }
    }
}

macro_rules! make_gauge {
    ($name:ident, $category:expr, $action:expr) => {
        #[derive(Debug, Default, Clone)]
        pub struct $name(GaugeCounter);
        impl std::ops::Deref for $name {
            type Target = GaugeCounter;

            fn deref(&self) -> &Self::Target {
                &self.0
            }
        }
        impl TryFrom<Record<'_>> for $name {
            type Error = ();
            fn try_from(record: Record<'_>) -> Result<Self, ()> {
                record.try_gauge($category, $action).map($name)
            }
        }
    };
}

/// A gauge represents a single numerical value that can arbitrarily go up and down.
#[derive(Debug, Default, Clone)]
pub struct GaugeCounter(u32);
impl GaugeCounter {
    /// Get the inner value.
    pub fn value(&self) -> u32 {
        self.0
    }
}

make_gauge!(VolumeCreated, Category::Volume, Action::Created);
make_gauge!(VolumeDeleted, Category::Volume, Action::Deleted);
make_gauge!(PoolCreated, Category::Pool, Action::Created);
make_gauge!(PoolDeleted, Category::Pool, Action::Deleted);

impl<'a, T: TryFrom<Record<'a>>> From<&'a EventsRecord> for Option<T> {
    fn from(src: &'a EventsRecord) -> Option<T> {
        src.record.iter().find_map(|s| T::try_from(Record(s)).ok())
    }
}
```

tiagolobocastro (commented):
I'm a bit confused on the counter/gauge, on our last call I thought gauge was ever incrementing, but documentation suggest that's a counter rather than gauge:
https://prometheus.io/docs/concepts/metric_types/
_A counter is a cumulative metric that represents a single [monotonically increasing counter](https://en.wikipedia.org/wiki/Monotonic_function) whose value can only increase or be reset to zero on restart._

sinhaashish:
> I'm a bit confused on the counter/gauge, on our last call I thought someone said gauge was ever incrementing, but documentation suggest that's a counter rather than gauge: https://prometheus.io/docs/concepts/metric_types/ _A counter is a cumulative metric that represents a single [monotonically increasing counter](https://en.wikipedia.org/wiki/Monotonic_function) whose value can only increase or be reset to zero on restart._

changes made as suggested

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -184,3 +215,93 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
     }
     pools_size_vector
 }
+
+/// Record of events populated from prometheus.
+#[derive(Debug, Default, Clone)]
```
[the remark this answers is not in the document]
tiagolobocastro: This is still in the middle

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -184,3 +199,164 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
...
+            fn try_from(record: Record<'_>) -> Result<Self, ()> {
+                record.try_counter($resource, $action).map($name)
+            }
+        }
+    };
+}
+
+/// A Counter represents a single numerical value that only increases.
+#[derive(Debug, Default, Clone)]
+pub struct CounterValue(u32);
```
tiagolobocastro, line 345: pub(crate) ?

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -184,3 +199,164 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
...
+        if self.resource() == resource && self.action() == action {
+            return Ok(CounterValue(self.counter_value()));
+        }
+        Err(())
+    }
+}
+
+/// Metrics Resource.
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq)]
+pub enum Resource {
```
tiagolobocastro, line 289: pub(crate) ?

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -184,3 +199,164 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
...
+    pub(crate) volume_created: VolumeCreated,
+    pub(crate) volume_deleted: VolumeDeleted,
+    pub(crate) pool_created: PoolCreated,
+    pub(crate) pool_deleted: PoolDeleted,
+}
+
+/// Record of events populated from prometheus.
+#[derive(Debug, Default, Clone)]
+/// Wrapper for Stats fetched from prometheus data.
+pub struct EventsRecord {
```
tiagolobocastro, line 215: pub(crate)

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -184,3 +199,164 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
...
+    pub record: Vec<Sample>,
```
tiagolobocastro, line 216: pub(crate) ? or maybe private, this doesn't need to be exposed outside this module?

call-home/src/bin/callhome/main.rs
```diff
@@ -29,6 +32,10 @@ struct CliArgs {
     /// The namespace we are supposed to operate in.
     #[arg(short, long, default_value = DEFAULT_NAMESPACE)]
     namespace: String,
+
+    /// The endpoint to fetch events stats.
+    #[clap(long, short, default_value = DEFAULT_TRANSMISSION_URL)]
+    transmission_url: Url,
```
tiagolobocastro, line 38: What does transmission mean?
should this be renamed to something more like `stats_aggregator` ?

tiagolobocastro (approved):

call-home/src/bin/callhome/collector/report_models.rs
sinhaashish: made record as private, cant make `EventsRecord` as pub(crate) as  it is being used here `errors::Result<EventsRecord>` .

call-home/src/bin/callhome/collector/report_models.rs
sinhaashish: Upon making the Counter type as `pub (crate)` it was failing as the `Deref `trait for the counter types inside the make_counter! macro, attempts to use CounterValue as the target type and since CounterValue is pub crate, it cannot be used as the target type in the Deref implementation. Thus leaving it as pub.

Abhinandan-Purkait (approved):

sinhaashish:
bors merge

PR #281, opened 2023-07-04
feat(upgrade-job): install new CRDs

niladrih:
The helm upgrade --install command doesn't actually install CRDs which are missing.

Changes:
- add CRD creation step before helm upgrade

### DIFF
--- a/k8s/upgrade/src/bin/upgrade-job/common/error.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/common/error.rs
@@ -582,6 +582,33 @@ pub(crate) enum Error {
         args: Vec<String>,
         std_err: String,
     },
+
+    /// Error for when we fail to read the entries of a directory.
+    #[snafu(display("Failed to read the contents of directory {}: {}", path.display(), source))]
+    ReadingDirectoryContents {
+        source: std::io::Error,
+        path: PathBuf,
+    },
+
+    /// Error for when the 'crds' directory inside a helm chart directory is either not a
+    /// directory, or it does not exist.
+    #[snafu(display("Helm chart 'crds' directory {} is invalid", path.display()))]
+    InvalidHelmChartCrdDir { path: PathBuf },
+
+    /// Error for when CRD creation fails.
+    #[snafu(display("Failed to create CustomResourceDefinition '{}': {}", name, source))]
+    CreateCrd { source: kube::Error, name: String },
+
+    /// Error for when unwraping of Result<DirEntry, std::io::Error> fails.
+    #[snafu(display(
+        "Failed to collect DirEntry list from read_dir() into a Vec<_> for directory {}: {}",
+        path.display(),
+        source
+    ))]
+    CollectDirEntries {
+        source: std::io::Error,
+        path: PathBuf,
+    },
 }
 
 /// A wrapper type to remove repeated Result<T, Error> returns.
--- a/k8s/upgrade/src/bin/upgrade-job/common/kube_client.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/common/kube_client.rs
@@ -1,7 +1,10 @@
 use crate::common::error::{K8sClientGeneration, KubeClientSetBuilderNs, Result};
-use k8s_openapi::api::{
-    apps::v1::Deployment,
-    core::v1::{Namespace, Pod},
+use k8s_openapi::{
+    api::{
+        apps::v1::Deployment,
+        core::v1::{Namespace, Pod},
+    },
+    apiextensions_apiserver::pkg::apis::apiextensions::v1::CustomResourceDefinition,
 };
 use kube::{api::Api, Client};
 use snafu::ResultExt;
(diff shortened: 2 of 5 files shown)


k8s/upgrade/src/bin/upgrade-job/common/kube_client.rs
```diff
@@ -46,12 +50,14 @@ pub(crate) struct KubeClientSet {
     pods_api: Api<Pod>,
     namespaces_api: Api<Namespace>,
     deployments_api: Api<Deployment>,
+    customresourcedefinition_api: Api<CustomResourceDefinition>,
```
tiagolobocastro, line 53: crd is well known, will be easier to read imho

```suggestion
    crd_api: Api<CustomResourceDefinition>,
```

k8s/upgrade/src/bin/upgrade-job/helm/client.rs
```diff
@@ -260,3 +272,61 @@ impl HelmReleaseClient {
...
+            let crd_yaml = fs::read(entry.as_path()).context(ReadingFile {
+                filepath: entry.clone(),
+            })?;
+
+            let crd: Crd = serde_yaml::from_slice(crd_yaml.as_slice())
+                .context(YamlParseFromFile { filepath: entry })?;
+
+            // Create CRDs, and ignore creation failures due to the CRD already
+            // existing in the cluster.
+            let _pp = PostParams::default();
```
tiagolobocastro, line 306: we use it right below..
```suggestion
            let pp = PostParams::default();
```

k8s/upgrade/src/bin/upgrade-job/helm/client.rs
```diff
@@ -260,3 +272,61 @@ impl HelmReleaseClient {
...
+            let creation_result = crd_client.create(&_pp, &crd).await;
+            if let Err(err) = creation_result {
+                match err {
+                    // Return early if creation has failed due to the CRD already existing in the
+                    // cluster. Ref: https://github.com/kubernetes/apimachinery/blob/v0.27.3/pkg/apis/meta/v1/types.go#L846
+                    kube::Error::Api(response) if response.reason.eq("AlreadyExists") => {
+                        info!(
```
tiagolobocastro, line 313: should we check if it's the version we expect? maybe add a warning message if it's not?

k8s/upgrade/src/bin/upgrade-job/helm/client.rs
```diff
@@ -260,3 +272,61 @@ impl HelmReleaseClient {
...
+                            crd.name_any()
+                        );
+                        continue;
+                    }
+                    _ => {
+                        return CreateCrd {
+                            source_err: err,
+                            name: crd.name_any(),
+                        }
+                        .fail()
```
tiagolobocastro, line 324: Should we try to upgrade the other ones before returning the error?

k8s/upgrade/src/bin/upgrade-job/common/error.rs
```diff
@@ -582,6 +582,36 @@ pub(crate) enum Error {
...
+
+    /// Error for when the 'crds' directory inside a helm chart directory is either not a
+    /// directory, or it does not exist.
+    #[snafu(display("Helm chart 'crds' directory {} is invalid", path.display()))]
+    InvalidHelmChartCrdDir { path: PathBuf },
+
+    /// Error for when CRD creation fails.
+    #[snafu(display("Failed to create CustomResourceDefinition '{}': {}", name, source_err))]
+    CreateCrd {
+        source_err: kube::Error,
```
tiagolobocastro, line 601: usually these are called just `source`

tiagolobocastro (approved):

k8s/upgrade/src/bin/upgrade-job/helm/client.rs
niladrih: That's an important one, thank you for noticing this. I'm not quite clear on how to do this at the moment. But it's going into a TODO, and I'll circle back to this.

k8s/upgrade/src/bin/upgrade-job/helm/client.rs
niladrih: No, I don't think we should. Missing CRD creation, shouldn't fail, and if it does, it might be required to exit prematurely because we don't know what it could mean for the rest of the upgrade?

tiagolobocastro: Yeah that makes sense.
What happens in this case, we'd keep the previous chart or we'd be in a broken state?

niladrih: Helm upgrade doesn't begin until this function returns, so we are, for-sure going to be in the source chart.

k8s/upgrade/src/bin/upgrade-job/common/error.rs
niladrih: I just wanted to use this as a hack because the source field isn't exposed when building it (maybe because it doesn't actually exist in the generated type). 'source' is only captured when using the context api, or so I thought. Turns out there's an IntoError trait which let's you feed the source error to the Snafu type.

Using source, instead of source_err.

Abhinandan-Purkait (approved):

niladrih:
bors merge

PR #293, opened 2023-07-10
chore(eventing): add argument to skip saving report to swarm

sinhaashish:


call-home/src/bin/callhome/main.rs
```diff
@@ -60,6 +64,10 @@ async fn run() -> anyhow::Result<()> {
     let version = release_version();
     let endpoint = args.endpoint;
     let aggregator_url = args.aggregator_url;
+    let save_report = args
+        .save_report
+        .parse()
+        .expect("Invalid boolean value for save_report");
```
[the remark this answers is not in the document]
sinhaashish: Had tried boolean , but it errored out stating 
```
root@ashish-dev:~/code/rust/mayastor-extensions(stop-pushing-to-swarm)$ k logs mayastor-obs-callhome-6b68bcbd4f-rstl6 -c obs-callhome -f
error: unexpected value 'true' for '--save-report' found; no more were expected

Usage: obs-callhome <--endpoint <ENDPOINT>|--namespace <NAMESPACE>|--save-report|--aggregator-url <AGGREGATOR_URL>>

For more information, try '--help'.
```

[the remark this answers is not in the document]
tiagolobocastro: You need to either change clap argument to do parsing from string to bool, or change helm to do and if, you can see other examples in existing helm templates

niladrih:
This should be default behaviour, I think there should be a flag to disable this. Also I don't think this should be a configurable option. The whole point of this container is to send the report, I think.

blaisedias:
> Save report should be default behaviour, I think there should be a flag to disable this. Also I don't think this should be a configurable option. The whole point of this container is to send the report, I think.

This should be a configurable option unless the statistics should include test clusters spun up by us for testing purposes. Normally callhome is disabled - but then call home would be untested.

call-home/src/bin/callhome/main.rs
```diff
@@ -33,6 +33,10 @@ struct CliArgs {
     #[arg(short, long, default_value = DEFAULT_NAMESPACE)]
     namespace: String,
 
+    /// Save the report.
+    #[clap(long, short)]
+    save_report: bool,
```
tiagolobocastro, line 38: Suggest renaming and documenting what it means, example:
```suggestion
    /// Sends the report to the remote collection endpoint.
    #[clap(long, short)]
    sendReport: bool,
```

tiagolobocastro (approved):

call-home/src/bin/callhome/main.rs
sinhaashish: done

Abhinandan-Purkait (approved):

sinhaashish:
bors merge

chart/values.yaml
```diff
@@ -627,6 +627,8 @@ obs:
     enabled: true
     # -- Log level for callhome
     logLevel: "info"
+    # -- Save the report
```
tiagolobocastro, line 630:
```suggestion
```

Don't think we should document this, makes no sense for an end-user?

sinhaashish: removed

sinhaashish:
bors merge

sinhaashish:
bors merge

PR #299, opened 2023-07-20
chore(upgrade): modify check for rebuilding

sinhaashish:


abhilashshetty04:
@sinhaashish, Lets add nexus state validation before cordoning the node. Not necessary to list all nexuses in cluster. Only nexuses of the node about to be cordoned is fine.

@tiagolobocastro @avishnu , During every node upgrade, wait_for_rebuild lists all volumes and checks rebuild_progress in all childs. This will not return if any child has rebuild_progress. I think we should wait even if child is Faulted as eventually same or different child will rebuilding and then we will have rebuild_pregress populated. It might happen on the next node upgrade. So wait until all child is Online if we do that validation?

sinhaashish:
Closing this PR as a consolidated  #314  is raised to resolve this issue.

PR #301, opened 2023-07-21
feat(tests): add script for creating minikube cluster

niladrih:
This PR adds testing tools which will be of use when running in-repo tests.

## Description

This change adds a new directory and 3 new files.
```console
tests/
├── minikube/
│   └── setup.sh
├── README.md
└── shell.nix
```
The shell script creates a single-node minikube kubernetes cluster. Optionally, it can also pull in (from GitHub) and set up the following dependencies as well:
- https://github.com/Mirantis/cri-dockerd: This is skipped if it already exists. Pulls in released binary and configures systemd service.
- https://github.com/containernetworking/plugins: This is always installed whenever the `--install-prerequisites` option is used.

This script uses kubernetes v1.25.10 by default.
Installs kubectl cli utility of the same version as the configured kubernetes version. Replaces existing kubectl, if present.
It is interactive, but it can be bypassed with the `--assume-yes` option.
The following dependencies are pulled in by the nix-shell -- awk, conntrack, jq, crictl, minikube, curl, sed, cacert
Requires docker.

Limitations:
1. Only works with systemd-based Linux x86_64
2. Minikube setup is not configurable (e.g. cpu, mem)

```console
Usage: setup.sh [OPTIONS]

Options:
  -h, --help                      Display this text.
  --kubernetes-version <version>  Specify the version of kubernetes.
  -i, --install-prerequisites     Install cri-dockerd and containernetworking plugins before creating minikube cluster.
  -y, --assume-yes                Assume the answer 'Y' for all interactive questions.
  -k, --skip-kube-context-switch  Skip switching kubectl cluster to the created minikube cluster's context.

Examples:
  setup.sh --kubernetes-version v1.25.11
```

Tested on an ubuntu-22.04 vm.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,58 @@
+# Tests
+
+## Set up Minikube
+Platform prerequisites:
+- Systemd-based linux-amd64 system
+- Docker (does not require containerd)
+- Nix
+
+Usage:
+```console
+Usage: setup.sh [OPTIONS]
+
+Options:
+  -h, --help                      Display this text.
+  --kubernetes-version <version>  Specify the version of kubernetes.
+  -i, --install-prerequisites     Install cri-dockerd and containernetworking plugins before creating minikube cluster.
+  -y, --assume-yes                Assume the answer 'Y' for all interactive questions.
+  -k, --skip-kube-context-switch  Skip switching kubectl cluster to the created minikube cluster's context.
+
+Examples:
+  setup.sh --kubernetes-version v1.25.11
+```
+
+Run the following command to set up a Minikube single-node kubernetes cluster:
+```console
+nix-channel --update
+sudo -E env "PATH=$PATH" nix-shell
+./minikube/setup.sh [OPTIONS]
+```
+
+Verify kubernetes control-plane pod statuses
+```console
+kubectl get pods -n kube-system
+```
+Sample output:
+```console
+$ kubectl get pods -n kube-system
+
+NAME                                       READY   STATUS    RESTARTS   AGE
+calico-kube-controllers-85578c44bf-g2tfx   1/1     Running   0          59s
+calico-node-sn6kk                          1/1     Running   0          59s
+coredns-5d78c9869d-q6r6q                   1/1     Running   0          59s
+etcd-niladri-vm                            1/1     Running   0          71s
+kube-apiserver-niladri-vm                  1/1     Running   0          71s
+kube-controller-manager-niladri-vm         1/1     Running   0          71s
+kube-proxy-dqb2v                           1/1     Running   0          59s
+kube-scheduler-niladri-vm                  1/1     Running   0          71s
+```
+
+## Tear down Minikube
+
+Tear down an existing minikube cluster using the following command while in the nix-shell:
+```console
+minikube stop
+# Exit the nix-shell using the 'exit' command.
+# exit
+```
+Alternatively, use `minikube delete` to delete the minikube profile and all files related to your minikube cluster.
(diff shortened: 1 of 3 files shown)


tests/setup.sh
```diff
@@ -0,0 +1,223 @@
...
+  ${CURL} -H "Accept: application/octet-stream" "${github_asset_url}" -o "${output_filepath}"
+  cd -
+}
+
+install_prerequisites() {
+  # Check if LINUX.
+  local os_error="not a GNU/Linux environment"
+  os=$(must_succeed_command "uname -o" "${os_error}" | tr -d " \t\r\n")
+  if [ "${os}" != "GNU/Linux" ]; then
+    die "${os_error}"
```
tiagolobocastro, line 113: Doesn't run on the m1's ?

tiagolobocastro (approved):
Should we move these things to a minikube or k8s subfolder?
Should we add a stop script as well?

tests/setup.sh
niladrih: No, it doesn't. For all practical uses with Mayastor, we'd have to stick to amd64 linux. But, it's a worthwhile enhancement for this script.  

I don't have a nix development environment on my m1, so I'd have to poke around to explore this :(

niladrih: The minikube part of the script will work on m1s, but not the prerequisites. There is one bug here, the kubectl cli tool's installation assumes it's an amd64. I'll fix it..

niladrih: Fixed it.

niladrih:
>Should we move these things to a minikube or k8s subfolder?

Should the shell.nix be outside of the minikube directory, or inside it?

tests/setup.sh
```diff
@@ -0,0 +1,296 @@
...
+    esac
+    ;;
+  *)
+    die "the 'install_prerequisites' option does not support the OS ${os}" 1
+    ;;
+  esac
+}
+
+# Installing minikube and kubectl from hosted binaries because the nix unstable channel
+# versions are usually outdated.
```
tiagolobocastro, line 206: This is not true, unstable has the latest versions...

tests/setup.sh
```diff
@@ -0,0 +1,296 @@
...
+    dl_link_arch_path="arm64"
+    ;;
+  *)
+    die "the 'install_kubectl' option does not support the arch ${arch}" 1
+    ;;
+  esac
+
+  # Pulling minikube and kubectl binaries.
+  cd "${workspace}"
+  ${CURL} "https://storage.googleapis.com/minikube/releases/latest/minikube-${dl_link_os_path}-${dl_link_arch_path}" -o minikube
```
tiagolobocastro, line 240: This seems to be downloading things everytime?

tiagolobocastro:
> > Should we move these things to a minikube or k8s subfolder?
> 
> @tiagolobocastro -- Should the shell.nix be outside of the minikube directory, or inside it?

hmm not sure, probably inside?

niladrih:
I've put the script inside the directory, I've kept the shell.nix and the README.md outside. The README.md is a general one. WDYT @tiagolobocastro ?

tests/setup.sh
niladrih: Oh. I messed up. I forgot to run `nix-channel --update`

niladrih: I'll move back to using nix.

tests/setup.sh
niladrih: Fixed it. Minikube is pulled in via nix, that makes it idempotent. Kubectl checks for existing binary with the required version, also idempotent.

sinhaashish (approved):
Tested & lgtm

niladrih:
bors merge

PR #302, opened 2023-07-25
fix(eventing): skip created and deleted events while serialising in case of default value

sinhaashish:


tiagolobocastro (approved):
Suggest renaming title and commit, this doesn't seem like a chore but a fix or refactor

datacore-vvarakantham (approved):

sinhaashish:
bors merge

PR #306, opened 2023-07-30
fix(eventing): remove some from created removed events

sinhaashish:
The reports were coming like `created: Some(0)` .
So replaced `Option::is_none` with `is_zero` .

### DIFF
--- a/call-home/src/bin/callhome/collector/report_models.rs
+++ b/call-home/src/bin/callhome/collector/report_models.rs
@@ -17,10 +17,10 @@ pub(crate) struct Volumes {
     mean_size_in_bytes: u64,
     max_size_in_bytes: u64,
     capacity_percentiles_in_bytes: Percentiles,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    created: Option<u32>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    deleted: Option<u32>,
+    #[serde(skip_serializing_if = "is_zero")]
+    created: u32,
+    #[serde(skip_serializing_if = "is_zero")]
+    deleted: u32,
 }
 impl Volumes {
     /// Receives a openapi::models::Volumes object and returns a new report_models::volume object by
@@ -48,10 +48,10 @@ pub(crate) struct Pools {
     min_size_in_bytes: u64,
     mean_size_in_bytes: u64,
     capacity_percentiles_in_bytes: Percentiles,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    created: Option<u32>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    deleted: Option<u32>,
+    #[serde(skip_serializing_if = "is_zero")]
+    created: u32,
+    #[serde(skip_serializing_if = "is_zero")]
+    deleted: u32,
 }
 impl Pools {
     /// Receives a vector of openapi::models::Pool and returns a new report_models::Pools object by
@@ -348,8 +348,8 @@ macro_rules! make_counter {
 pub struct CounterValue(u32);
 impl CounterValue {
     /// Get the inner value.
-    pub(crate) fn value(&self) -> Option<u32> {
-        Some(self.0)
+    pub(crate) fn value(&self) -> u32 {
+        self.0
     }
 }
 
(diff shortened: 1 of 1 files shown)


call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -363,3 +363,8 @@ impl<'a, T: TryFrom<Record<'a>>> From<&'a EventsRecord> for Option<T> {
         src.record.iter().find_map(|s| T::try_from(Record(s)).ok())
     }
 }
+
+// Define the `is_zero` function to determine if the field should be serialized.
+fn is_zero<T: PartialEq + Default>(value: &T) -> bool {
+    *value == T::default()
```
tiagolobocastro, line 369: This doesn't look like it's doing what you want, this is the same as is_none?

sinhaashish: No, 
 The reports were coming like  `created: Some(1), deleted: Some(1) ` . Ths Pr removes Option from the report. and in case its zero skip the serialisation . 
```
Report { k8s_cluster_id: "daec579d161164639ad8c9fa48bc2be6528e47000b5215ba13495944b7a99be6", k8s_node_count: 4, product_name: "Mayastor", product_version: "v2.2.0-rc.1", deploy_namespace: "4dbb4d6af483bf116d48dc9f40cbe4da19a03a5b36f1f1725ebe7aa9f796e393", storage_node_count: 3, pools: Pools { count: 0, max_size_in_bytes: 0, min_size_in_bytes: 0, mean_size_in_bytes: 0, capacity_percentiles_in_bytes: Percentiles { percentile_50: 0, percentile_75: 0, percentile_90: 0 }, created: Some(1), deleted: Some(1) }, volumes: Volumes { count: 0, min_size_in_bytes: 0, mean_size_in_bytes: 0, max_size_in_bytes: 0, capacity_percentiles_in_bytes: Percentiles { percentile_50: 0, percentile_75: 0, percentile_90: 0 }, created: Some(1), deleted: Some(1) }, replicas: Replicas { count: 0, count_per_volume_percentiles: Percentiles { percentile_50: 0, percentile_75: 0, percentile_90: 0 } }, versions: Versions { control_plane_version: "" } }
```

tiagolobocastro: hmm I don't see how this is skiping serialisation for zeros..
Do you have an example before and after this change?

sinhaashish: Not related to this code  , but the `is_zero` works like as shown here 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b452fb231a1d8a11fb867872615d6701

tiagolobocastro: I know how it works, what I'm saying is your code here doesn't seem like it does what you're claiming. It won't skip Some(0), it still only skips None?

sinhaashish: It it wont skip `Some(0)` , but upon serialisation it will give `created: 0` , which is not desired, In case of 0 we should not have this field that why adde `is_zero`.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=63b5438d1e1480e67ae9241cf3cc0957

tiagolobocastro: But that's not what your code is doing? With you code it will still show Some(0) as created: 0?

tiagolobocastro: Oh I see, you also removed option!

sinhaashish: will share the before and after report to make things more clear.

tiagolobocastro: > will share the before and after report to make things more clear.

No it's fine, I had missed the fact you changed type from Option<n> to n.

tiagolobocastro (approved):

call-home/src/bin/callhome/collector/report_models.rs
tiagolobocastro, line 369: Though suggest you change impl of this function as it's still not doing what's it's suggesting

tiagolobocastro: Something like this is clearer..

```rust
fn is_zero(value &i32) -> bool {
  value == &0
}
```

sinhaashish: done

datacore-vvarakantham (approved):

sinhaashish:
bors merge

PR #307, opened 2023-08-02
feat(upgrade): add set flag in upgrade

sinhaashish:
Helm doesn't validates the  value passed 
Example 
```
root@ashish-dev:~/code/rust/mayastor-extensions(set-flag)$ helm install mayastor mayastor/mayastor -n mayastor --create-namespace --version 2.3.0 --set='alpha=beta,gamma={30,101}'
NAME: mayastor
LAST DEPLOYED: Wed Aug  9 06:26:30 2023
NAMESPACE: mayastor
STATUS: deployed
REVISION: 1
NOTES:
OpenEBS Mayastor has been installed. Check its status by running:
$ kubectl get pods -n mayastor
```
Upon getting the values 
```
root@ashish-dev:~/code/rust/mayastor-extensions(set-flag)$ helm get values mayastor --all  -n mayastor  -oyaml > ~/test.yaml
```

The values present in `test.yaml` 
```
root@ashish-dev:~$ grep -i alpha  test.yaml
alpha: beta
root@ashish-dev:~$ grep -i -C 4 gamma  test.yaml
      limits: {}
      requests: {}
eventing:
  enabled: false
gamma:
- 30
- 101
image:
  pullPolicy: IfNotPresent
```

So , we don't need to validate the values passed, Its just that we need to validate the format.

This PR validates by running `helm upgrade --dry-run` command .
```
root@ashish-dev:~/code/rust/mayastor-extensions(set-flag)$ ./target/debug/kubectl-mayastor  upgrade --set obs.callhome.enabled=false,a={1,2  --skip-upgrade-path-validation-for-unsupported-version

ServiceAccount: mayastor-upgrade-service-account-bbb created in namespace: mayastor
Cluster Role: mayastor-upgrade-role-bbb in namespace mayastor created
ClusterRoleBinding: mayastor-upgrade-role-binding-bbb in namespace mayastor created
Job: mayastor-upgrade-bbb created in namespace: mayastor

 The validation failed for upgrade, hence deleting teh upgrade resources. Please re-run upgrade with valid values.
 {"fromVersion":"2.3.0","toVersion":"0.0.0","message":"Failed to upgrade: `helm upgrade` command return an error,\ncommand: helm,\nargs: [\"upgrade\", \"mayastor\", \"/chart\", \"-n\", \"mayastor\", \"--timeout\", \"15m\", \"-f\", \"/chart/.tmpGpv2QI\", \"--set\", \"obs.callhome.enabled=false,a={1,2\", \"--atomic\", \"--dry-run\"],\nstd_err: Error: failed parsing --set data: list must terminate with '}'\n"}
Job mayastor-upgrade-bbb in namespace mayastor deleted
ClusterRoleBinding mayastor-upgrade-role-binding-bbb in namespace mayastor deleted
ClusterRole mayastor-upgrade-role-bbb in namespace mayastor deleted
ServiceAccount mayastor-upgrade-service-account-bbb in namespace mayastor deleted
```

### DIFF
--- a/k8s/plugin/src/main.rs
+++ b/k8s/plugin/src/main.rs
@@ -172,10 +172,7 @@ async fn execute(cli_args: CliArgs) {
                     &cli_args.namespace,
                     cli_args.kube_config_path.clone(),
                     cli_args.timeout,
-                    resources.skip_single_replica_volume_validation,
-                    resources.skip_replica_rebuild,
-                    resources.skip_cordoned_node_validation,
-                    resources.skip_upgrade_path_validation_for_unsupported_version,
+                    &resources,
                 )
                 .await
                 .map_err(|error| {
@@ -186,7 +183,10 @@ async fn execute(cli_args: CliArgs) {
                 if resources.dry_run {
                     _ = resources.dummy_apply(&cli_args.namespace).await;
                 } else {
-                    resources.apply(&cli_args.namespace).await;
+                    _ = resources.apply(&cli_args.namespace).await.map_err(|error| {
+                        eprintln!("{error}");
+                        std::process::exit(error.into());
+                    });
                 }
             }
 
--- a/k8s/upgrade/src/bin/upgrade-job/common/error.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/common/error.rs
@@ -346,6 +346,10 @@ pub(crate) enum Error {
         std_err: String,
     },
 
+    /// Error for when a Helm upgrade command execution succeeds, but with an error.
+    #[snafu(display("`helm upgrade --dry-run` command return an error, error: {}", std_err,))]
+    HelmUpgradeDryRunCommand { std_err: String },
+
     /// Error for when a Helm get values command execution succeeds, but with an error.
     #[snafu(display(
         "`helm get values` command return an error,\ncommand: {},\nargs: {:?},\nstd_err: {}",
--- a/k8s/upgrade/src/bin/upgrade-job/events/event_recorder.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/events/event_recorder.rs
@@ -251,12 +251,17 @@ impl EventRecorder {
     }
 
     /// This method is intended for use when upgrade fails.
-    pub(crate) async fn publish_unrecoverable<Error>(&self, err: &Error)
+    pub(crate) async fn publish_unrecoverable<Error>(&self, err: &Error, validation_error: bool)
     where
         Error: Display,
     {
+        let action = if validation_error {
+            EventAction::ValidationFailed
+        } else {
+            EventAction::Failed
+        };
         let _ = self
-            .publish_warning(format!("Failed to upgrade: {err}"), "Failed")
+            .publish_warning(format!("Failed to upgrade: {err}"), action)
             .await
             .map_err(|error| error!(%error, "Failed to upgrade {PRODUCT}"));
     }
@@ -266,5 +271,38 @@ impl EventRecorder {
         let _ = self.event_sender.take();
 
         let _ = self.event_loop_handle.await;
+    }
+}
+
+/// current volume status
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize)]
+pub enum EventAction {
+    #[serde(rename = "Failed")]
+    Failed,
+    #[serde(rename = "Validation Failed")]
+    ValidationFailed,
+    #[serde(rename = "Upgrading control-plane")]
+    UpgradingCP,
+    #[serde(rename = "Upgraded control-plane")]
+    UpgradedCP,
+    #[serde(rename = "Upgrading data-plane")]
+    UpgradingDP,
+    #[serde(rename = "Upgraded data-plane")]
+    UpgradedDP,
+    #[serde(rename = "Successful")]
+    Successful,
+}
+
+impl ToString for EventAction {
+    fn to_string(&self) -> String {
+        match self {
+            Self::Failed => String::from("Failed"),
+            Self::ValidationFailed => String::from("Validation Failed"),
+            Self::UpgradingCP => String::from("Upgrading control-plane"),
+            Self::UpgradedCP => String::from("Upgraded control-plane"),
+            Self::UpgradingDP => String::from("Upgrading data-plane"),
+            Self::UpgradedDP => String::from("Upgraded data-plane"),
+            Self::Successful => String::from("Successful"),
+        }
     }
 }
--- a/k8s/upgrade/src/bin/upgrade-job/helm/client.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/helm/client.rs
@@ -68,7 +68,7 @@ impl HelmReleaseClientBuilder {
 /// releases.
 #[derive(Clone)]
 pub(crate) struct HelmReleaseClient {
-    namespace: String,
+    pub(crate) namespace: String,
 }
 
 impl HelmReleaseClient {
--- a/k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
@@ -2,9 +2,10 @@ use crate::{
     common::{
         constants::{CORE_CHART_NAME, TO_UMBRELLA_SEMVER, UMBRELLA_CHART_NAME},
         error::{
-            CoreChartUpgradeNoneChartDir, HelmUpgradeOptionsAbsent, InvalidHelmUpgrade,
-            InvalidUpgradePath, NoInputHelmChartDir, NotAKnownHelmChart, RegexCompile, Result,
-            RollbackForbidden, UmbrellaChartNotUpgraded,
+            CoreChartUpgradeNoneChartDir, HelmCommand, HelmUpgradeDryRunCommand,
+            HelmUpgradeOptionsAbsent, InvalidHelmUpgrade, InvalidUpgradePath, NoInputHelmChartDir,
+            NotAKnownHelmChart, RegexCompile, Result, RollbackForbidden, U8VectorToString,
+            UmbrellaChartNotUpgraded,
         },
     },
     helm::{client::HelmReleaseClient, values::generate_values_yaml_file},
@@ -12,15 +13,14 @@ use crate::{
 };
 use regex::Regex;
 use semver::Version;
-
 use snafu::{ensure, ResultExt};
-use std::path::PathBuf;
+use std::{path::PathBuf, process::Command, str};
 use tempfile::NamedTempFile as TempFile;
-use tracing::info;
+use tracing::{debug, info};
 
 /// This is the helm chart variant of the helm chart installed in the cluster.
 /// The PRODUCT may be installed using either of these options, but never both.
-#[derive(Clone, Default)]
+#[derive(Clone, Default, PartialEq, Eq)]
 pub(crate) enum HelmChart {
     #[default]
     Umbrella,
@@ -34,6 +34,7 @@ pub(crate) struct HelmUpgradeBuilder {
     namespace: Option<String>,
     core_chart_dir: Option<PathBuf>,
     skip_upgrade_path_validation: bool,
+    values: Option<String>,
 }
 
 impl HelmUpgradeBuilder {
@@ -74,6 +75,16 @@ impl HelmUpgradeBuilder {
         self
     }
 
+    /// This is a builder option to add set flags set during upgrade.
+    #[must_use]
+    pub(crate) fn with_values<J>(mut self, values: J) -> Self
+    where
+        J: ToString,
+    {
+        self.values = Some(values.to_string());
+        self
+    }
+
     /// This builds the HelmUpgrade object.
     pub(crate) async fn build(self) -> Result<HelmUpgrade> {
         ensure!(
@@ -82,6 +93,7 @@ impl HelmUpgradeBuilder {
         );
         let release_name = self.release_name.clone().unwrap();
         let namespace = self.namespace.clone().unwrap();
+        let values = self.values.clone().unwrap_or_default();
 
         // Generate HelmReleaseClient.
         let client = HelmReleaseClient::builder()
@@ -183,6 +195,8 @@ impl HelmUpgradeBuilder {
             core_chart_extra_args = Some(vec_to_strings![
                 "-f",
                 _upgrade_values_file.path().to_string_lossy(),
+                "--set",
+                values,
                 "--atomic"
             ]);
             upgrade_values_file = Some(_upgrade_values_file)
@@ -224,6 +238,69 @@ impl HelmUpgrade {
         HelmUpgradeBuilder::default()
     }
 
+    /// Use the HelmReleaseClient's upgrade dry runcommand to validate the upgrade command.
+    pub(crate) async fn dry_run(self) -> Result<()> {
+        let mut args_with_dry_run = self.core_chart_extra_args.clone();
+
+        if let Some(mut args) = args_with_dry_run {
+            args.push("--dry-run".to_string());
+            args_with_dry_run = Some(args);
+        } else {
+            args_with_dry_run = Some(vec!["--dry-run".to_string()]);
+        }
+
+        if self.chart_variant == HelmChart::Core && !self.already_upgraded {
+            // It should be impossible to hit this error.
+            let chart_dir = self
+                .core_chart_dir
+                .ok_or(CoreChartUpgradeNoneChartDir.build())?;
+
+            info!("Validating helm upgrade...");
+
+            let command: &str = "helm";
+            let mut args: Vec<String> = vec_to_strings![
+                "upgrade",
+                self.release_name,
+                chart_dir.to_string_lossy(),
+                "-n",
+                self.client.namespace,
+                "--timeout",
+                "15m"
+            ];
+
+            // Extra args
+            args.extend(
+                args_with_dry_run
+                    .unwrap_or_default()
+                    .iter()
+                    .map(ToString::to_string),
+            );
+
+            debug!(%command, ?args, "Helm upgrade dry-run command");
+            let output =
+                Command::new(command)
+                    .args(args.clone())
+                    .output()
+                    .context(HelmCommand {
+                        command: command.to_string(),
+                        args: args.clone(),
+                    })?;
+
+            let stdout_str = str::from_utf8(output.stdout.as_slice()).context(U8VectorToString)?;
+            let stderr_str = str::from_utf8(output.stderr.as_slice()).context(U8VectorToString)?;
+            debug!(stdout=%stdout_str, stderr=%stderr_str, "Helm upgrade dry-run command standard output and error");
+            ensure!(
+                output.stderr.is_empty(),
+                HelmUpgradeDryRunCommand {
+                    std_err: str::from_utf8(output.stderr.as_slice())
+                        .context(U8VectorToString)?
+                        .to_string()
+                }
+            );
+        }
+        Ok(())
+    }
+
     /// Use the HelmReleaseClient's upgrade method to upgrade the installed helm release.
     pub(crate) async fn run(mut self) -> Result<()> {
         match self.chart_variant {
@@ -253,7 +330,6 @@ impl HelmUpgrade {
                 self.client
                     .upgrade(self.release_name, chart_dir, self.core_chart_extra_args)
                     .await?;
-
                 // This file is no longer required after the upgrade command has been executed.
                 self.upgrade_values_file = None;
 
--- a/k8s/upgrade/src/bin/upgrade-job/opts.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/opts.rs
@@ -38,6 +38,11 @@ pub(crate) struct CliArgs {
     /// The name of the Kubernetes Job Pod. The Job object will be used to post upgrade event.
     #[arg(env = "POD_NAME")]
     pod_name: String,
+
+    /// The set values specified by the user for upgrade
+    /// (can specify multiple or separate values with commas: key1=val1,key2=val2).
+    #[arg(short, long)]
+    values: String,
 }
 
 impl CliArgs {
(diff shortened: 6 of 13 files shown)


k8s/upgrade/src/bin/upgrade-job/helm/values.rs
```diff
@@ -109,8 +110,21 @@ pub(crate) fn generate_values_yaml_file(
...
 
+    // Override all the values by the value passed in set flag
+    let parts: Vec<&str> = values.split(',').collect();
+    for tokens in parts {
+        let data: Vec<&str> = tokens.split('=').collect();
+        yq.set_string_value(
+            YamlKey::try_from(format!(".{}", data[0]).as_str())?,
+            data[1],
+            upgrade_values_file.path(),
+        )?;
```
tiagolobocastro, line 126: This will panic if index 1 or 0 don't exist.
You can get around that using slice pattern matching, example:
```rust
      let v = l.split("=").collect::<Vec<_>>(); 
      match v[..] {
        [ var, value ] => println!("{var}={value}"),
        _ => println!("Invalid!")
      }
```
But we have to consider here what kind of inputs we allow. What if we have ',' in the values, example: --set="v={1,2}" ?

sinhaashish: @tiagolobocastro  as discussed the helm upgrade dry-run command validates the values. 
PTAL

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
```diff
@@ -82,6 +93,7 @@ impl HelmUpgradeBuilder {
         );
         let release_name = self.release_name.clone().unwrap();
         let namespace = self.namespace.clone().unwrap();
+        let values = self.values.clone().unwrap();
```
tiagolobocastro, line 96: We should handle error instead of unwrap.
We could either:
1. use like builder-like pattern and return error here
2. not allow them to be optional by including their setup in the new fn
3. use type-state pattern and only expose build method if they're all Some

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -103,25 +103,66 @@ pub struct UpgradeArgs {
     /// Upgrade to an unsupported version.
     #[clap(global = true, hide = true, long, default_value_t = false)]
     pub skip_upgrade_path_validation_for_unsupported_version: bool,
+
+    /// Upgrade to an unsupported version.
```
tiagolobocastro, line 107:
```suggestion
    /// The set values specified by the user for upgrade.
    /// (can specify multiple or separate values with commas: key1=val1,key2=val2).
```

tiagolobocastro: Helm seems to have this:
```
      --set stringArray                            set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)
      --set-file stringArray                       set values from respective files specified via the command line (can specify multiple or separate values with commas: key1=path1,key2=path2)
      --set-json stringArray                       set JSON values on the command line (can specify multiple or separate values with commas: key1=jsonval1,key2=jsonval2)
      --set-string stringArray                     set STRING values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)
```
Should we support more of this? IMHO I think we should at least support set-file.
CC @niladrih

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -103,25 +103,66 @@ pub struct UpgradeArgs {
...
-        )
-        .await
-        .map_err(|error| {
-            std::process::exit(error.into());
-        });
-        console_logger::info(UPGRADE_JOB_STARTED, "");
+        UpgradeResources::create_upgrade_resources(namespace, self).await?;
+
+        for _i in 0 .. 6 {
+            tokio::time::sleep(std::time::Duration::from_secs(10)).await;
```
tiagolobocastro, line 142: Should we sleep at the end of the first error only? perhaps add a smaller sleep right before the loop

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -103,25 +103,66 @@ pub struct UpgradeArgs {
...
+            match gur.get_upgrade_events(release_name.clone(), false).await {
+                Ok(latest_event) => {
+                    if latest_event.type_ == Some("Warning".to_string()) {
```
tiagolobocastro, line 145: I suggest you make this a bit clearer, example:
```rust
for _ in 0..6 {
  let Ok(latest_event) = gur.get_upgrade_events(release_name.clone(), false).await else {
    continue;
  }
  ok code here...
  break
}
```

tiagolobocastro, line 145: Rather than checking for first error being a warning, should we have some specific details here about validation failed?

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -238,31 +297,40 @@ impl UpgradeEventClient {
...
+        Some(data) => {
+            let e: UpgradeEvent = serde_json::from_str(data.as_str())
+                .context(error::EventSerdeDeserialization { event: data })?;
+            println!("Upgrade From: {}", e.from_version);
+            println!("Upgrade To: {}", e.to_version);
+            println!("Upgrade Status: {}", e.message);
+        }
+        None => return error::MessageInEventNotPresent.fail(),
+    };
+    Ok(())
```
tiagolobocastro, line 331:
```suggestion
            Ok(())
        }
        None => error::MessageInEventNotPresent.fail(),
    }
```

k8s/upgrade/src/bin/upgrade-job/opts.rs
```diff
@@ -38,6 +38,10 @@ pub(crate) struct CliArgs {
     /// The name of the Kubernetes Job Pod. The Job object will be used to post upgrade event.
     #[arg(env = "POD_NAME")]
     pod_name: String,
+
+    /// This is the values passed during the helm upgrade.
```
tiagolobocastro, line 42:
```suggestion
    /// The set values specified by the user for upgrade
    /// (can specify multiple or separate values with commas: key1=val1,key2=val2).
```

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
sinhaashish: There is a line just above which handles the error in case of None.
```
 ensure!(
            self.release_name.is_some() && self.namespace.is_some(),
            HelmUpgradeOptionsAbsent
        );
```
And for set values i changed it to unwrap_or_default 
`let values = self.values.clone().unwrap_or_default();`

k8s/upgrade/src/plugin/upgrade.rs
sinhaashish: `set-file` can be taken as separate PR

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
tiagolobocastro: hmm a more idiomatic way of doing it would not be this way but rather something like this:
let release_name = self.release_name.ok_or(Err::HelmUpgradeOptionsAbsent{})?;
where now you have a variable which has be validated and ready to use.

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
```diff
@@ -224,6 +238,31 @@ impl HelmUpgrade {
...
+    /// Use the HelmReleaseClient's upgrade dry runcommand to validate the upgrade command.
+    pub(crate) async fn dry_run(self) -> Result<()> {
+        let mut args_with_dry_run = self.core_chart_extra_args.clone();
+
+        if args_with_dry_run.is_none() {
+            args_with_dry_run = Some(vec!["--dry-run".to_string()]);
+        } else if let Some(mut args) = args_with_dry_run {
+            args.push("--dry-run".to_string());
+            args_with_dry_run = Some(args);
+        }
```
tiagolobocastro, line 250:
```suggestion
        if let Some(mut args) = args_with_dry_run {
            args.push("--dry-run".to_string());
            args_with_dry_run = Some(args);
        } else {
            args_with_dry_run = Some(vec!["--dry-run".to_string()]);
        }
```

k8s/upgrade/src/plugin/upgrade.rs
sinhaashish: IMHO , no as the events publishing might take time so just waiting for 10 seconds initially

k8s/upgrade/src/plugin/upgrade.rs
sinhaashish:
```
 // Delete any previous upgrade events
 gur.delete_upgrade_events(release_name.clone()).await?;
```
The above code actually deletes any upgrade related events before creating the job , so there will not be any upgrade related events before teh job starts. And on the job side the event is published as action as Error or others. So made the code change where matching id cone of actions for Error. 
https://github.com/openebs/mayastor-extensions/blob/5c35c902f9d01df7bb7593334ff5bcacd5563ebf/k8s/upgrade/src/bin/upgrade-job/events/event_recorder.rs#L259

k8s/upgrade/src/bin/upgrade-job/common/error.rs
```diff
@@ -346,6 +346,13 @@ pub(crate) enum Error {
         std_err: String,
     },
 
+    /// Error for when a Helm upgrade command execution succeeds, but with an error.
+    #[snafu(display(
+        "`helm upgrade --dry-run` command return an error, std_err: {}",
```
tiagolobocastro, line 351: what does std_err mean?

k8s/upgrade/src/bin/upgrade-job/helm/client.rs
```diff
@@ -68,7 +68,7 @@ impl HelmReleaseClientBuilder {
 /// releases.
 #[derive(Clone)]
 pub(crate) struct HelmReleaseClient {
-    namespace: String,
+    pub namespace: String,
```
tiagolobocastro, line 71: Seems a bit odd to have a pub(crate) struct with a pub method.

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
```diff
@@ -224,6 +238,68 @@ impl HelmUpgrade {
...
+        let mut args_with_dry_run = self.core_chart_extra_args.clone();
+
+        if let Some(mut args) = args_with_dry_run {
+            args.push("--dry-run".to_string());
+            args_with_dry_run = Some(args);
+        } else {
+            args_with_dry_run = Some(vec!["--dry-run".to_string()]);
+        }
+
+        if self.chart_variant == HelmChart::Core && !self.already_upgraded {
```
tiagolobocastro, line 252: what does this chart variant mean?

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
```diff
@@ -224,6 +238,68 @@ impl HelmUpgrade {
...
+                Command::new(command)
+                    .args(args.clone())
+                    .output()
+                    .context(HelmCommand {
+                        command: command.to_string(),
+                        args: args.clone(),
+                    })?;
+
+            let stdout_str = str::from_utf8(output.stdout.as_slice()).context(U8VectorToString)?;
+            debug!(stdout=%stdout_str, "Helm upgrade dry-run command standard output");
```
tiagolobocastro, line 290: Ah now I see what std_err was...
Should we include std_err in the debug log here as well?

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -103,25 +104,80 @@ pub struct UpgradeArgs {
     /// Upgrade to an unsupported version.
     #[clap(global = true, hide = true, long, default_value_t = false)]
     pub skip_upgrade_path_validation_for_unsupported_version: bool,
+
+    /// The set values specified by the user for upgrade.
+    /// (can specify multiple or separate values with commas: key1=val1,key2=val2).
+    #[clap(global = true, long)]
+    pub set: Vec<String>,
```
tiagolobocastro, line 111:
```suggestion
    pub set_args: Vec<String>,
```
?

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -103,25 +104,80 @@ pub struct UpgradeArgs {
...
+            skip_replica_rebuild: false,
+            skip_cordoned_node_validation: false,
+            skip_upgrade_path_validation_for_unsupported_version: false,
+            set: Default::default(),
+        }
+    }
     ///  Upgrade the resources.
-    pub async fn apply(&self, namespace: &str) {
+    pub async fn apply(&self, namespace: &str) -> error::Result<()> {
+        let gur = UpgradeEventClient::new(namespace).await?;
```
tiagolobocastro, line 135: what does gur mean?

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -103,25 +104,80 @@ pub struct UpgradeArgs {
...
+
+    /// Handle the event and errors out for invalid helm command.
+    async fn handle_upgrade_event(
+        &self,
+        latest_event: Event,
+        namespace: &str,
+    ) -> error::Result<()> {
+        if let Some(action) = latest_event.action {
+            if action == "Failed" {
+                if let Some(data) = latest_event.message {
```
tiagolobocastro, line 164: I'm confused, this is checking the latest event, how do we know if this is the validation event?

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -220,8 +276,26 @@ impl UpgradeEventClient {
...
+        self.api_client()
+            .delete_collection(&DeleteParams::default(), &event_lp)
+            .await
+            .context(error::DeleteEventsWithFieldSelector { field: selector })?;
+        Ok(())
+    }
+
     /// Fetch upgrade events and print.
-    pub async fn get_and_log(&self, release_name: String) -> error::Result<()> {
+    pub async fn get_upgrade_events(
```
tiagolobocastro, line 294: this function name suggest plural events, but seems we return only one event?

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
sinhaashish: Its already there 
```
let stdout_str = str::from_utf8(output.stdout.as_slice()).context(U8VectorToString)?;
            debug!(stdout=%stdout_str, "Helm upgrade dry-run command standard output");
``

File: k8s/upgrade/src/bin/upgrade-job/helm/client.rs (hunk 10 above)
Comment (sinhaashish) [reply k8s/upgrade/src/bin/upgrade-job/helm/client.rs:71]: changed to pub(crate)

File: k8s/plugin/src/main.rs [hunk 17]
```diff
@@ -186,7 +183,10 @@ async fn execute(cli_args: CliArgs) {
                 if resources.dry_run {
                     _ = resources.dummy_apply(&cli_args.namespace).await;
                 } else {
-                    resources.apply(&cli_args.namespace).await;
+                    _ = resources.apply(&cli_args.namespace).await.map_err(|error| {
+                        println!(" Error {error:#?}");
```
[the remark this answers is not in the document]
sinhaashish: done

k8s/upgrade/src/bin/upgrade-job/common/error.rs
sinhaashish: replaced it with error

k8s/upgrade/src/plugin/upgrade.rs
sinhaashish: done

k8s/upgrade/src/plugin/upgrade.rs
sinhaashish: renamed it to `upgrade_event_client`

k8s/upgrade/src/plugin/upgrade.rs
sinhaashish: renamed it to `get_latest_upgrade_event`

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
sinhaashish: This is the helm chart variant of the helm chart installed in the cluster its either CORE or umbrella. @niladrih  would like to know your input for the difference between them, The dry run is only for the core chart in this PR.

tiagolobocastro: Ah I see...
And for umbrella we don't want to run this dry run?

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -238,31 +317,43 @@ impl UpgradeEventClient {
...
-                    println!("Upgrade From: {}", e.from_version);
-                    println!("Upgrade To: {}", e.to_version);
-                    println!("Upgrade Status: {}", e.message);
-                    Ok(())
-                }
-                None => error::MessageInEventNotPresent.fail(),
-            },
-            None => error::UpgradeEventNotPresent.fail(),
+        let latest_event = event_list
+            .get(0)
```
tiagolobocastro, line 321: hmm are we returning the latest event of the first?

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
tiagolobocastro: I mean this debug log only includes the stdout and not the stderr, should we include stderr as well?

k8s/upgrade/src/plugin/upgrade.rs
sinhaashish: There can be multiple events so sorting them and getting the first element gives the latest event .

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
sinhaashish: added

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
sinhaashish: Cant confirm on exact reason but here https://github.com/openebs/mayastor-extensions/blob/f134161734b725202199c7911e64dd4dd4abb17c/k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs#L236 in case of Umbrella chart the upgrade is not being done . So i have skipped that in dry run as well. May be @niladrih  can help on this.

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
```diff
@@ -224,6 +238,69 @@ impl HelmUpgrade {
...
+                    .args(args.clone())
+                    .output()
+                    .context(HelmCommand {
+                        command: command.to_string(),
+                        args: args.clone(),
+                    })?;
+
+            let stdout_str = str::from_utf8(output.stdout.as_slice()).context(U8VectorToString)?;
+            let stderr_str = str::from_utf8(output.stderr.as_slice()).context(U8VectorToString)?;
+            debug!(stdout=%stdout_str,  stderr=%stderr_str, "Helm upgrade dry-run command standard output and error");
```
tiagolobocastro, line 291:
```suggestion
            debug!(stdout=%stdout_str, stderr=%stderr_str, "Helm upgrade dry-run command standard output and error");
```

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
```diff
@@ -224,6 +238,69 @@ impl HelmUpgrade {
...
+            ensure!(
```
tiagolobocastro, line 292: Do we also need to check for the exit code?

k8s/upgrade/src/plugin/upgrade.rs
tiagolobocastro: Ah I would have thought sorting by timestamp would give 0 as the earliest rather than latest.
Btw when we get a validation error there's no chance of another event being added after that?

k8s/upgrade/src/bin/upgrade-job/events/event_recorder.rs
```diff
@@ -251,12 +251,17 @@ impl EventRecorder {
     }
 
     /// This method is intended for use when upgrade fails.
-    pub(crate) async fn publish_unrecoverable<Error>(&self, err: &Error)
+    pub(crate) async fn publish_unrecoverable<Error>(&self, err: &Error, validation_error: bool)
     where
         Error: Display,
     {
+        let action = if validation_error {
+            "Validation Failed"
```
tiagolobocastro, line 259: this should be an enum which both the job and the plugin can can use for checking what kind of error it is.

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
tiagolobocastro: hmm interesting then I wonder if we should suggest ppl to use the umbrella chart at all, seems better to just use mayastor chart directly so you can get upgrades like this.

tiagolobocastro (approved):

k8s/upgrade/src/plugin/upgrade.rs
sinhaashish: No , once we get a validation we error out and return. So no other upgrade related event can be publihsed
```
   if let Err(error) = helm_upgrade_dry_run.dry_run().await {
        event.publish_unrecoverable(&error, true).await;
        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
        return Err(error);
    }
```

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
sinhaashish: Hmm, if @niladrih  agrees to it. but that can be taken in another PR.

k8s/upgrade/src/plugin/upgrade.rs
tiagolobocastro: So the job doesn't get restarted again on failure?

k8s/upgrade/src/bin/upgrade-job/events/event_recorder.rs
sinhaashish: done

Abhinandan-Purkait (approved):

sinhaashish:
bors merge

sinhaashish:
bors merge

sinhaashish:
bors merge

PR #314, opened 2023-08-17
chore(upgrade): fix rebuild validation

sinhaashish:
After the addition of partial rebuild, the rebuild validations has been changed for upgrade.

The logic to check if any rebuild is in progress has been changed. 

## Description

The new rebuild validation is based on the logic below.
1.   _If all the volumes are online (across the cluster) then continue with upgrade_ 
2.  _In case of volumes which are degraded/faulted check for replica rebuild in progress_

      -  If replica rebuild in progress then wait for rebuild to complete 	
      - If no replica rebuild in progress then step wait for a period of 10+ minutes with an interval of 35 seconds. At every interval check for replica rebuild in progress, If rebuilding then wait , else after 10 minutes continue with the upgrade.

There can be scenario where the already checked volume becomes degraded later, so add a vector of unhealthy volumes, and wait till the size of unhealthy volumes.

The rebuild in pre-validation i.e. client side is not modified with the above logic as it just checks for rebuilding . And once the job is created the the rebuild is again verified. So all the above logic reside in Upgrade job

## Motivation and Context

The rebuild status would have been incorrect if we just check for `child.rebuild_progress.is_some()`

## Regression

No

## How Has This Been Tested?

Ran the manual upgrade.

## Types of changes

## Checklist:

### DIFF
--- a/k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
@@ -7,12 +7,67 @@ use crate::common::{
 };
 use k8s_openapi::api::core::v1::Pod;
 use kube::{api::ObjectList, ResourceExt};
+use openapi::models::{Volume, VolumeStatus};
 use semver::{Version, VersionReq};
 use snafu::ResultExt;
+use std::time::Duration;
 use tracing::{info, warn};
 
-/// Function to check for any volume rebuild in progress across the cluster
-pub(crate) async fn is_rebuilding(rest_client: &RestClientSet) -> Result<bool> {
+/// Contains the Rebuild Results.
+#[derive(Default)]
+pub(crate) struct RebuildResult {
+    pub(crate) rebuilding: bool,
+    pub(crate) discarded_volumes: Vec<Volume>,
+}
+
+/// Function to check for any volume rebuild in progress across the cluster.
+pub(crate) async fn rebuild_result(
+    rest_client: &RestClientSet,
+    stale_volumes: &mut Vec<Volume>,
+) -> Result<RebuildResult> {
+    loop {
+        let unhealthy_volumes = list_unhealthy_volumes(rest_client, stale_volumes).await?;
+        if unhealthy_volumes.is_empty() {
+            break;
+        }
+
+        for volume in unhealthy_volumes.iter() {
+            match replica_rebuild_count(volume.clone()).await {
+                0 => {
+                    for _i in 0 .. 11 {
+                        // wait for a minute for any rebuild to start
+                        tokio::time::sleep(Duration::from_secs(60_u64)).await;
+                        let count = replica_rebuild_count(volume.clone()).await;
+                        if count > 0 {
+                            return Ok(RebuildResult {
+                                rebuilding: true,
+                                discarded_volumes: stale_volumes.clone(),
+                            });
+                        }
+                    }
+                    stale_volumes.push(volume.clone());
+                }
+                _ => {
+                    return Ok(RebuildResult {
+                        rebuilding: true,
+                        discarded_volumes: stale_volumes.clone(),
+                    })
+                }
+            }
+        }
+    }
+    Ok(RebuildResult {
+        rebuilding: false,
+        discarded_volumes: stale_volumes.clone(),
+    })
+}
+
+/// Return the list of unhealthy volumes.
+pub(crate) async fn list_unhealthy_volumes(
+    rest_client: &RestClientSet,
+    discarded_volumes: &[Volume],
+) -> Result<Vec<Volume>> {
+    let mut unhealthy_volumes: Vec<Volume> = Vec::new();
     // The number of volumes to get per request.
     let max_entries = 200;
     let mut starting_token = Some(0_isize);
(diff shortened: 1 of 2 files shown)


k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -7,12 +7,42 @@ use crate::common::{
...
+        if unhealthy_volumes.is_empty() {
+            break;
+        }
+
+        for volume in unhealthy_volumes.iter() {
+            match replica_rebuild_count(volume.clone()).await {
+                0 => {
+                    for _i in 0 .. 11 {
+                        // wait for a minute for any rebuild to start
+                        tokio::time::sleep(Duration::from_secs(60_u64)).await;
```
abhilashshetty04, line 29: etcd cache period is by default 30s. So objects are update with that interval. So i think we should use lesser sleep . perhaps for eg 35s?

abhilashshetty04: We return here does the caller of this fn call this again until it gets Ok(false) ?

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -7,12 +7,42 @@ use crate::common::{
...
+                        let count = replica_rebuild_count(volume.clone()).await;
+                        if count > 0 {
```
abhilashshetty04, line 31: In the last attempt (10 mins passed) if we see there is no rebuild. We return Ok(true). We do no exclude this volume for which we already waited 10 mins? This volume will again be listed as rebuild didnt start and we wait again?

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -7,12 +7,42 @@ use crate::common::{
...
+use openapi::models::{Volume, VolumeStatus};
 use semver::{Version, VersionReq};
 use snafu::ResultExt;
+use std::time::Duration;
 use tracing::{info, warn};
 
 /// Function to check for any volume rebuild in progress across the cluster
 pub(crate) async fn is_rebuilding(rest_client: &RestClientSet) -> Result<bool> {
+    loop {
+        let unhealthy_volumes = list_unhealthy_volumes(rest_client).await?;
```
Abhinandan-Purkait, line 19: This will insert unhealthy volume again and again even after timed wait. Please rather maintain a map of timed wait completed volumes and exclude them gradually.

sinhaashish: done

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
sinhaashish: Addressed , changed it to 35seconds , but will need some more upvote on this. @tiagolobocastro

k8s/upgrade/src/plugin/preflight_validations.rs
```diff
@@ -136,32 +136,71 @@ pub(crate) async fn rebuild_in_progress_validation(client: &RestClient) -> error
...
+    let mut rebuild_count = 0;
+    if let Some(target) = &volume.state.target {
+        for child in target.children.iter() {
+            if child.rebuild_progress.is_some() {
+                rebuild_count += 1;
+            }
+        }
+    }
+    rebuild_count
 }
```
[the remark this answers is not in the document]
sinhaashish: The preflight validation is rebuild doesn't have all these condition, as it just a client side validation. It doesn't require to wait for the rebuilds to complete and all..
The upgrade job needs to wait for the rebuilds so the code for rebuild check is only added in the Upgrade job code.

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
tiagolobocastro: Problem is now we wait 5.8 minutes rather than 10 minutes..
Either change back to 60 or adjust loop to something like:
```rust
let start = std::time::Instant::now();
const rebuildMaxWait: Duration = Duration::from_seconds(60); // -> can put this together to other consts/statics.
while start.elapased() < rebuildMaxWait {
...
}
```

sinhaashish: reverted to 60 seconds interval.

Abhinandan-Purkait (approved):

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -28,27 +83,37 @@ pub(crate) async fn is_rebuilding(rest_client: &RestClientSet) -> Result<bool> {
...
             }
         }
     }
-    Ok(false)
+    unhealthy_volumes.retain(|v| !discarded_volumes.contains(v));
+    Ok(unhealthy_volumes)
+}
+
+/// Count of number of replica rebuilding.
+pub(crate) async fn replica_rebuild_count(volume: Volume) -> i32 {
```
tiagolobocastro, line 99: can take a ref and avoid cloning

tiagolobocastro (approved):

abhilashshetty04 (approved):

sinhaashish:
bors merge

sinhaashish:
bors merge

PR #315, opened 2023-08-21
chore(upgrade): update upgrade docs

sinhaashish:
Update upgrade docs.
## Description

Add `set-args` to upgrade docs and remove the _capital_ aliases for `skip-data-plane-restart` , `skip-single-replica-volume-validation`, `skip-replica-rebuild`, `skip-cordoned-node-validation` to make it consistent with helm command 
The helm command doesn't have any capital alias.
## Motivation and Context

In case the user sets any value using --set-args flag at the time of install. The upgrade might override those values. 
So in order to keep the user values intact the user have to pass the `--set-args ` while upgrade.
Like `kubectl-mayastor upgrade --set-args obs.callhome.enabled=false`

## Regression

No. This is just a docs update , The feature was in PR #307 

## How Has This Been Tested?

1. Install mayastor 
 ```
helm install mayastor mayastor/mayastor -n mayastor --set obs.callhome.enabled=false --create-namespace --version 2.3.0 
```
2. Run upgrade
```
 kubectl-mayastor  upgrade --set-args obs.callhome.enabled=true  --skip-upgrade-path-validation-for-unsupported-version
```
 
 

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/k8s/plugin/README.md
+++ b/k8s/plugin/README.md
@@ -344,27 +344,29 @@ kubectl mayastor dump system -d <output_directory> -n <mayastor_namespace>
 
   Options:
   -d, --dry-run
-        Display all the validations output but will not execute upgrade
+          Display all the validations output but will not execute upgrade
   -r, --rest <REST>
-        The rest endpoint to connect to
-  -D, --skip-data-plane-restart
-        If set then upgrade will skip the io-engine pods restart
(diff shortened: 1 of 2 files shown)


sinhaashish:
bors merge

PR #319, opened 2023-08-22
fix(upgrade-job): wait for event worker to exit for errors

niladrih:
The upgrade job main thread starts a worker thread which listens for events and publishes them (as kubernetes events). When the upgrade method exits, the 'EventRecorder' instance for it is dropped. This in turn drops the event channel (used to pass events to the event worker) and also the handle to event worker. The event worker does not have enough time after the channel is dropped, to publish its events and exit. 

This PR defers all of the event publishes and actual upgrade steps to an inner function. The caller of this function makes sure the event loop exits before returning.

### DIFF
--- a/k8s/upgrade/src/bin/upgrade-job/events/event_recorder.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/events/event_recorder.rs
@@ -71,35 +71,25 @@ impl EventRecorderBuilder {
         self
     }
 
-    /// This is a builder option add the from-version in upgrade.
-    #[must_use]
-    pub(crate) fn with_from_version(mut self, from: String) -> Self {
-        self.from_version = Some(from);
-        self
-    }
-
-    /// This is a builder option add the to-version in upgrade.
-    #[must_use]
-    pub(crate) fn with_to_version(mut self, to: String) -> Self {
-        self.to_version = Some(to);
-        self
-    }
-
     // TODO: Make the builder option validations error out at compile-time, using std::compile_error
     // or something similar.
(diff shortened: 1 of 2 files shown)


Abhinandan-Purkait (approved):

tiagolobocastro (commented):
Not sure if waiting on a drop is a good practice...
Could we achieve the same some other way?

tiagolobocastro:
> > Not sure if waiting on a drop is a good practice... Could we achieve the same some other way?
> 
> Sure, could you suggest a suitable pattern for fleshing out this destructor-like tool.

What about just waiting for the work to complete in `shutdown_worker`?

niladrih:
> > > Not sure if waiting on a drop is a good practice... Could we achieve the same some other way?
> > 
> > 
> > Sure, could you suggest a suitable pattern for fleshing out this destructor-like tool.
> 
> What about just waiting for the work to complete in `shutdown_worker`?

So, no change?

niladrih:
Just noticed that shutdown worker doesn't guarantee publish of error events. Only guarantees publish of successful upgrade. I think this is still required. What might be a good alternative to the drop thing, @tiagolobocastro ?

niladrih:
Updated code in this PR to not use the EventRecorder's Drop. Instead it moves the event recorder to one level up in the calling sequence. This let's the event recorder's shutdown_worker method wait for the loop to exit.

sinhaashish (approved):
Tested & LGTM

abhilashshetty04 (approved):

niladrih:
bors merge

PR #328, opened 2023-08-25
refactor(upgrade-job): change dry-run helm upgrade implementation 

niladrih:
This PR removes the duplicate implementation of the helm upgrade command to allow for a dry-run upgrade command flag inclusion.

## Description

The upgrade-job runs the 'helm upgrade --dry-run' command. The implementation for the helm commands api is largely constrained to the helm::client module. However, the dry-run upgrade command is implemented separately in the helm::upgrade module. The dry-run upgrade command is largely similar to the regular upgrade command, and can use the helm::client api. The requirement does not require a separate implementation of the command execution tools in the helm::upgrade module. 

This PR changes the implementation of helm upgrade dry-run to use the existing api. It also defers the execution of the helm upgrade command to another section of the code, by passing a closure for the same. This allows for error handling and allows us to fail fast when the dry-run fails.

## Motivation and Context

Just a refactor, makes it prettier. Doesn't have any behavioural changes.

N/A

## Regression

No

N/A

## How Has This Been Tested?

I've tested this on a kubeadm kubernetes cluster while upgrading from v2.3.0 to release-2.4 images on the develop branch helm chart.

I've tested on a kubernetes v1.25.10 kubeadm cluster, with the mayastor-2.3.0 helm chart installed. I've upgraded the mayastor deployment to the release-2.4 branch images.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/k8s/upgrade/src/bin/upgrade-job/helm/client.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/helm/client.rs
@@ -17,7 +17,12 @@ use kube::ResourceExt;
 use kube_client::{api::PostParams, Api};
 use serde::Deserialize;
 use snafu::{ensure, IntoError, ResultExt};
-use std::{fs, path::PathBuf, process::Command, str};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+    process::Command,
+    str,
+};
 use tracing::{debug, info};
 
 /// This struct is used to deserialize the output of `helm list -n <namespace> --deployed -o yaml`.
@@ -189,8 +194,9 @@ impl HelmReleaseClient {
     pub(crate) async fn upgrade<A, B>(
         &self,
         release_name: A,
-        chart_dir: PathBuf,
+        chart_dir: &Path,
         maybe_extra_args: Option<Vec<B>>,
+        install_crds: bool,
     ) -> Result<()>
     where
         A: ToString,
@@ -201,8 +207,11 @@ impl HelmReleaseClient {
             .with_namespace(self.namespace.as_str())
             .build()
             .await?;
-        // Ref: https://helm.sh/docs/chart_best_practices/custom_resource_definitions
-        install_missing_crds(k8s_client.crd_api(), chart_dir.join("crds")).await?;
+
+        if install_crds {
+            // Ref: https://helm.sh/docs/chart_best_practices/custom_resource_definitions
+            install_missing_crds(k8s_client.crd_api(), chart_dir.join("crds")).await?;
+        }
 
         let command: &str = "helm";
         let mut args: Vec<String> = vec_to_strings![
--- a/k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
@@ -2,10 +2,9 @@ use crate::{
     common::{
         constants::{CORE_CHART_NAME, TO_UMBRELLA_SEMVER, UMBRELLA_CHART_NAME},
         error::{
-            CoreChartUpgradeNoneChartDir, HelmCommand, HelmUpgradeDryRunCommand,
-            HelmUpgradeOptionsAbsent, InvalidHelmUpgrade, InvalidUpgradePath, NoInputHelmChartDir,
-            NotAKnownHelmChart, RegexCompile, Result, RollbackForbidden, U8VectorToString,
-            UmbrellaChartNotUpgraded,
+            CoreChartUpgradeNoneChartDir, HelmUpgradeOptionsAbsent, InvalidHelmUpgrade,
+            InvalidUpgradePath, NoInputHelmChartDir, NotAKnownHelmChart, RegexCompile, Result,
+            RollbackForbidden, UmbrellaChartNotUpgraded,
         },
     },
     helm::{client::HelmReleaseClient, values::generate_values_yaml_file},
@@ -14,9 +13,9 @@ use crate::{
 use regex::Regex;
 use semver::Version;
 use snafu::{ensure, ResultExt};
-use std::{path::PathBuf, process::Command, str};
+use std::{future::Future, path::PathBuf, pin::Pin};
 use tempfile::NamedTempFile as TempFile;
-use tracing::{debug, info};
+use tracing::info;
 
 /// This is the helm chart variant of the helm chart installed in the cluster.
 /// The PRODUCT may be installed using either of these options, but never both.
@@ -229,6 +228,7 @@ pub(crate) struct HelmUpgrade {
     core_chart_extra_args: Option<Vec<String>>,
     from_version: Version,
     to_version: Version,
+    #[allow(dead_code)]
     upgrade_values_file: Option<TempFile>,
 }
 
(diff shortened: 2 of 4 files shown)


niladrih:
Thanks to @tiagolobocastro for helping with the 'return an async closure' strategy.

sinhaashish:
@niladrih  Request you to please fill in the PR template. It might take 5 minutes of yours , but is helpful in opensoure world. Your pr says remove duplicate code . What is the code which is removed ?

niladrih:
bors try

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
```diff
@@ -238,109 +237,78 @@ impl HelmUpgrade {
...
+                            self.release_name,
+                            chart_dir.as_path(),
+                            self.core_chart_extra_args,
+                            true,
+                        )
+                        .await?;
+                    info!("Helm upgrade successful!");
+
+                    // This file is no longer required after the upgrade command has been executed.
+                    self.upgrade_values_file = None;
```
tiagolobocastro, line 305: self will be dropped, so we don't need this

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
```diff
@@ -351,3 +319,5 @@ impl HelmUpgrade {
         self.to_version.to_string()
     }
 }
+
+pub(crate) type HelmUpgradeRunner = Pin<Box<dyn Future<Output = Result<()>>>>;
```
tiagolobocastro, line 323: Worth a doc comment explaining how the run/dry-run works now

k8s/upgrade/src/bin/upgrade-job/upgrade.rs
```diff
@@ -37,20 +37,13 @@ pub(crate) async fn upgrade(opts: &CliArgs) -> Result<()> {
...
-
-    if let Err(error) = helm_upgrade_dry_run.dry_run().await {
+    // Dry-run helm upgrade.
+    let dry_run_result: Result<HelmUpgradeRunner> = helm_upgrade.dry_run().await;
+    if let Err(error) = dry_run_result {
         event.publish_unrecoverable(&error, true).await;
-        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
         return Err(error);
     }
+    let run_helm_upgrade = dry_run_result.unwrap();
```
tiagolobocastro, line 46:
```suggestion
    let run_helm_upgrade = match dry_run_result {
        Err(error) => {
            event.publish_unrecoverable(&error, true).await;
            Err(error)
        }
        Ok(run_helm_upgrade) => Ok(run_helm_upgrade)
    }?;
```

k8s/upgrade/src/plugin/constants.rs
```diff
@@ -71,7 +71,7 @@ pub(crate) const DEFAULT_IMAGE_REGISTRY: &str = "docker.io";
 pub(crate) const UPGRADE_JOB_IMAGE_TAG: &str = "develop";
 
 /// Upgrade job container image repository.
-pub(crate) const UPGRADE_JOB_IMAGE_REPO: &str = "openebs";
+pub(crate) const UPGRADE_JOB_IMAGE_REPO: &str = "niladrih";
```
tiagolobocastro: I've seen this many times, to avoid this ever getting checked in we probably should make this configurable via helm so ppl don't have to edit code..

```suggestion
pub(crate) const UPGRADE_JOB_IMAGE_REPO: &str = "openebs";
```

k8s/upgrade/src/bin/upgrade-job/upgrade.rs
```diff
@@ -67,9 +60,8 @@ pub(crate) async fn upgrade(opts: &CliArgs) -> Result<()> {
         .await?;
 
     // Control plane containers are updated in this step.
-    if let Err(error) = helm_upgrade.run().await {
+    if let Err(error) = run_helm_upgrade.await {
```
tiagolobocastro, line 63: I don't know what we had a sleep, but why is it now needed now?

k8s/upgrade/src/bin/upgrade-job/helm/client.rs
```diff
@@ -201,8 +207,11 @@ impl HelmReleaseClient {
             .with_namespace(self.namespace.as_str())
             .build()
             .await?;
-        // Ref: https://helm.sh/docs/chart_best_practices/custom_resource_definitions
-        install_missing_crds(k8s_client.crd_api(), chart_dir.join("crds")).await?;
+
+        if install_crds {
```
tiagolobocastro, line 211: oh seems like a fix? probably worth separate commit

k8s/upgrade/src/bin/upgrade-job/upgrade.rs
niladrih: This doesn't return the error. Just assigns it. We'd have to add a return?

k8s/upgrade/src/bin/upgrade-job/helm/client.rs
niladrih: This is a part for the refactor. The upgrade dry-run implementation doesn't do this step, this step needs to avoided when running the dry-run through the core upgrade method.

k8s/upgrade/src/bin/upgrade-job/upgrade.rs
tiagolobocastro: Yes it does, through the `?`

k8s/upgrade/src/bin/upgrade-job/upgrade.rs
niladrih: The sleeps were there to let the event recorder finish it's job, I've removed it because shutdown_worker will handle those cases. We didn't have sleeps there, those were introduced recently.

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
niladrih: Added a `#[allow(dead_code)]`.

tiagolobocastro: Why is that required?

niladrih: `self.upgrade_values_file` is a tempfile handle. it is used by the helm binary, but it is not used in the rust code. After upgrade method returns, we None the container, and so the tempfile is garbage-collected. Without this interaction, the file handle is never used. So it causes problems with the linter.

tiagolobocastro: Ah I see.. then simply add `#[allow(unused)]` on top of its declaration in the struct, rater than allowing deadcode for the entire struct:
struct {
a: bool,
#[allow(unused)]
b: bool,
}

tiagolobocastro (approved):

k8s/upgrade/src/bin/upgrade-job/helm/upgrade.rs
niladrih: Right, done.

sinhaashish (approved):

niladrih:
bors merge

PR #330, opened 2023-08-28
feat(upgrade): add set-file argument to upgrade

sinhaashish:
Add `set-file` argument to upgrade

## Description

On the line of `set-args` the `set-file` can take the file as input.

## Motivation and Context

You can use '--set-file' to set individual values from a file when the value itself is too long for the command line

## Regression

No

## How Has This Been Tested?

Install and upgrade using the `set-file` command . 

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/k8s/upgrade/src/bin/upgrade-job/opts.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/opts.rs
@@ -42,7 +42,12 @@ pub(crate) struct CliArgs {
     /// The set values specified by the user for upgrade
     /// (can specify multiple or separate values with commas: key1=val1,key2=val2).
     #[arg(short, long)]
-    values: String,
+    helm_args_set: String,
+
+    /// The set file values specified by the user for upgrade
+    /// (can specify multiple or separate values with commas: key1=path1,key2=path2).
+    #[arg(short, long)]
+    helm_args_set_file: String,
 }
 
 impl CliArgs {
@@ -83,7 +88,12 @@ impl CliArgs {
     }
 
     /// This returns the set values passed during upgrade.
-    pub(crate) fn values(&self) -> String {
-        self.values.clone()
+    pub(crate) fn helm_args_set(&self) -> String {
+        self.helm_args_set.clone()
+    }
+
+    /// This returns the set file passed during upgrade.
+    pub(crate) fn helm_args_set_file(&self) -> String {
+        self.helm_args_set_file.clone()
     }
 }
--- a/k8s/upgrade/src/plugin/objects.rs
+++ b/k8s/upgrade/src/plugin/objects.rs
@@ -253,10 +253,10 @@ pub(crate) fn upgrade_job(
     upgrade_image: String,
     release_name: String,
     args: &UpgradeArgs,
+    values: String,
     image_pull_secrets: Option<Vec<k8s_openapi::api::core::v1::LocalObjectReference>>,
     image_pull_policy: Option<String>,
 ) -> Job {
-    let values = args.set_args.join(",");
     let mut job_args: Vec<String> = vec![
         format!("--rest-endpoint=http://{release_name}-api-rest:8081"),
         format!("--namespace={namespace}"),
--- a/k8s/upgrade/src/plugin/objects.rs
+++ b/k8s/upgrade/src/plugin/objects.rs
@@ -1,7 +1,9 @@
 use crate::{
     plugin::constants::{
-        upgrade_name_concat, UPGRADE_BINARY_NAME, UPGRADE_JOB_CLUSTERROLEBINDING_NAME_SUFFIX,
-        UPGRADE_JOB_CLUSTERROLE_NAME_SUFFIX, UPGRADE_JOB_CONTAINER_NAME, UPGRADE_JOB_NAME_SUFFIX,
+        upgrade_name_concat, UPGRADE_BINARY_NAME, UPGRADE_CONFIG_MAP,
+        UPGRADE_CONFIG_MAP_MOUNT_PATH, UPGRADE_CONFIG_MAP_NAME_SUFFIX,
+        UPGRADE_JOB_CLUSTERROLEBINDING_NAME_SUFFIX, UPGRADE_JOB_CLUSTERROLE_NAME_SUFFIX,
+        UPGRADE_JOB_CONTAINER_NAME, UPGRADE_JOB_NAME_SUFFIX,
         UPGRADE_JOB_SERVICEACCOUNT_NAME_SUFFIX,
     },
     upgrade::UpgradeArgs,
@@ -11,11 +13,12 @@ use crate::{
 use k8s_openapi::api::{
     batch::v1::{Job, JobSpec},
     core::v1::{
-        Container, EnvVar, EnvVarSource, ExecAction, ObjectFieldSelector, PodSpec, PodTemplateSpec,
-        Probe, ServiceAccount,
+        ConfigMap, ConfigMapVolumeSource, Container, EnvVar, EnvVarSource, ExecAction,
+        ObjectFieldSelector, PodSpec, PodTemplateSpec, Probe, ServiceAccount, Volume, VolumeMount,
     },
     rbac::v1::{ClusterRole, ClusterRoleBinding, PolicyRule, RoleRef, Subject},
 };
+use std::collections::BTreeMap;
 
 use kube::core::ObjectMeta;
 use maplit::btreemap;
@@ -248,20 +251,43 @@ pub(crate) fn upgrade_job_cluster_role_binding(
     }
 }
 
+pub(crate) fn upgrade_configmap(
+    data: BTreeMap<String, String>,
+    namespace: &str,
+    release_name: String,
+) -> ConfigMap {
+    ConfigMap {
+        metadata: ObjectMeta {
+            labels: Some(upgrade_labels!()),
+            name: Some(upgrade_name_concat(
+                &release_name,
+                UPGRADE_CONFIG_MAP_NAME_SUFFIX,
+            )),
+            namespace: Some(namespace.to_string()),
+            ..Default::default()
+        },
+        data: Some(data),
+        immutable: Some(true),
+        ..Default::default()
+    }
+}
+
 pub(crate) fn upgrade_job(
     namespace: &str,
     upgrade_image: String,
     release_name: String,
     args: &UpgradeArgs,
-    values: String,
+    set_file: String,
     image_pull_secrets: Option<Vec<k8s_openapi::api::core::v1::LocalObjectReference>>,
     image_pull_policy: Option<String>,
 ) -> Job {
+    let helm_args_set = args.set.join(",");
     let mut job_args: Vec<String> = vec![
         format!("--rest-endpoint=http://{release_name}-api-rest:8081"),
         format!("--namespace={namespace}"),
         format!("--release-name={release_name}"),
-        format!("--values={values}"),
+        format!("--helm-args-set={helm_args_set}"),
+        format!("--helm-args-set-file={set_file}"),
     ];
     if args.skip_data_plane_restart {
         job_args.push("--skip-data-plane-restart".to_string());
@@ -325,12 +351,29 @@ pub(crate) fn upgrade_job(
                             period_seconds: Some(60),
                             ..Default::default()
                         }),
+                        volume_mounts: Some(vec![VolumeMount {
+                            read_only: Some(true),
+                            mount_path: UPGRADE_CONFIG_MAP_MOUNT_PATH.to_string(),
+                            name: UPGRADE_CONFIG_MAP.to_string(),
+                            ..Default::default()
+                        }]),
                         ..Default::default()
                     }],
                     service_account_name: Some(upgrade_name_concat(
                         &release_name,
                         UPGRADE_JOB_SERVICEACCOUNT_NAME_SUFFIX,
                     )),
+                    volumes: Some(vec![Volume {
+                        name: UPGRADE_CONFIG_MAP.to_string(),
+                        config_map: Some(ConfigMapVolumeSource {
+                            name: Some(upgrade_name_concat(
+                                &release_name,
+                                UPGRADE_CONFIG_MAP_NAME_SUFFIX,
+                            )),
+                            ..Default::default()
+                        }),
+                        ..Default::default()
+                    }]),
                     ..Default::default()
                 }),
             },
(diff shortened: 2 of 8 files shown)


niladrih (commented):
Left few comments.

niladrih:
This change should use a secret (not a configmap) to mount the file contents to the job pod, and should use the `--set-file` when running helm upgrade with the job. Interpreting helm's set-file API as set values might cause problems for us. We should let helm do its thing, as intended.

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -833,3 +841,21 @@ pub(crate) async fn get_source_version(ns: &str) -> error::Result<String> {
         .to_string();
     Ok(value.to_string())
 }
+
+/// Parse set-file and append set-args
+pub(crate) async fn parse_file_values(upgrade_args: &UpgradeArgs) -> error::Result<Vec<String>> {
+    let mut args: Vec<String> = Vec::new();
+    for file in &upgrade_args.set_file {
+        let data: Vec<_> = file.split('=').collect();
```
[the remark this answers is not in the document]
sinhaashish: Its because , you upgrade-job will run on a different node from where the file is present. Upgrade job will fail to find the file on that node. So reading the file and then passing it to the job.

sinhaashish:
@niladrih @tiagolobocastro  used config map in case of `--set-file` . PTAL

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -109,6 +109,11 @@ pub struct UpgradeArgs {
     /// (can specify multiple or separate values with commas: key1=val1,key2=val2).
     #[clap(global = true, long)]
     pub set_args: Vec<String>,
+
+    /// The set values from respective files specified via the command line
+    /// (can specify multiple or separate values with commas: key1=val1,key2=val2).
```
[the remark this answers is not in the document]
sinhaashish: Its inline with helm docs 
```
--set-file stringArray                       set values from respective files specified via the command line (can specify multiple or separate values with commas: key1=path1,key2=path2)
```

k8s/upgrade/src/plugin/objects.rs
```diff
@@ -253,10 +253,10 @@ pub(crate) fn upgrade_job(
     upgrade_image: String,
     release_name: String,
     args: &UpgradeArgs,
+    values: String,
```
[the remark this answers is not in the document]
sinhaashish: changed

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -833,3 +931,62 @@ pub(crate) async fn get_source_version(ns: &str) -> error::Result<String> {
...
+            .unwrap();
+        let cm_values =
+            fs::read_to_string(data[1]).context(error::ReadFromFile { filepath: data[1] })?;
+        let set_param = data[0].to_string();
+        // create config map key with key of set-file plus the file name
+        // this is done to make the key unique as /root/A.yaml and /root/data/A.yaml are different.
+        // Example : -- set-file jaeger-operator.tolerations=/root/tolerations.yaml
+        // create the key as jaeger-operator.tolerations_tolerations.yaml
+        // this is acts as a name of file under the mounted path inside pod.
+        let cm_key = format!("{set_param}_{file_name}");
```
niladrih, line 956: There should no more configmap data entries than the number of files. The same file should not create multiple configmap data entries.

niladrih (changes requested):
Left one comment

k8s/upgrade/src/plugin/upgrade.rs
sinhaashish: there can be scenarios:
1. --set-flag  jaeger-operator.tolerations=/root/tolerations.yaml --set-file ha.node.tolerations=/root/new/tolerations.yaml
Here the name of both the files are same but the files are different . So the `jaeger-operator.tolerations_tolerations.yaml`  and `ha.node.tolerations_tolerations.yaml` are two different files created in the pods.
2.  --set-flag  jaeger-operator.tolerations=/root/tolerations.yaml --set-file ha.node.tolerations=/root/tolerations.yaml. the files are passed are same. The PR still will create two file in pod. 
The problem here is we cant use the complete path `/root/tolerations.yaml` as the key as `/` is not allowed as the key for configmap. The key becomes the name of the file inside pod. Here the allowed values are `.`, `-`, `_`. So we had to parse `/root/tolerations.yaml` as `-root-tolerations.yaml` and then use it . This parsing is not ideal as the folders can have any seperators.

Solution to this problem : 
 a - just take the file name like `tolerations.yaml` as key . In this case both `/root/tolerations.yaml ` and `/root/new/tolerations.yaml` will have `tolerations.yaml`  as key , but the since data is a BtreeMap , the former gets overwritten by later , so even there are two different file , only one file gets created inside pod.
b- add a random number and append it to the file name to make it unique. In this case too two files will be generated. This has been done in the PR ,not by adding the random number, but appending the key `jaeger-operator.tolerations` to file name and thus making it unique.

IMHO this is the ideal approach where we dont have to do any parsing on the upgrade job . Its just running helm upgrade with `--set-flag  jaeger-operator.tolerations=/root/tolerations.yaml --set-file ha.node.tolerations=/root/new/tolerations.yaml` flag.

tiagolobocastro: Why can't we just use the key as the cm_key?

sinhaashish: we can use it , updated the code to use just key.

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -833,3 +930,43 @@ pub(crate) async fn get_source_version(ns: &str) -> error::Result<String> {
...
+    let mut upgrade_map = BTreeMap::new();
+    for file in &upgrade_args.set_file {
+        let data: Vec<_> = file.split('=').collect();
+        let cm_values =
+            fs::read_to_string(data[1]).context(error::ReadFromFile { filepath: data[1] })?;
+        // create config map key with key of set-file
+        // Example : -- set-file jaeger-operator.tolerations=/root/tolerations.yaml
+        // Key:  jaeger-operator.tolerations
+        // this is acts as a name of file under the mounted path inside pod.
+        let cm_key = data[0].to_string();
```
niladrih, line 947: I think we shouldn't have 1 file per key. This creates duplicate files when using the same file for several helm values (e.g. setting tolerations for multiple components). 

IMO, we should have 1 configmap entry per file. If the file names are the same, then the configmap key could be appended with random suffixes, or a '-' separated version of their absolute paths.

niladrih (commented):
Left one comment

k8s/upgrade/src/plugin/upgrade.rs
sinhaashish: The present implementation  does the below 
--set-args  jaeger-operator.tolerations=/home/ashish/tolerations.yaml
The key is `jaeger-operator.tolerations` , name of file is `jaeger-operator.tolerations` and the contents of file will be written in file`/upgrade-config-map/jaeger-operator.tolerations` inside pod.

Problem with having 1 configmap entry per file

Example `--set-file jaeger-operator.tolerations=/root/tolerations.yaml --set-file ha.node.tolerations=/root/temp/random/tolerations.yaml`

In this case if we work with file name as value then both the keys `jaeger-operator.tolerations` and `ha.node.tolerations` will have value as tolerations.yaml ( the file name) ; even though the files are different .

To solve this we need to keep a track of the files names and check if its same , in case its same then we add a random number to file name to make them unique.

All these , i.e. finding a file name. checking the file to the previous file name and then appending random number to file name adds to code debt. 

The solution to this problem is to create 1 config map per key. Yes agreed it will increase the number of config map but will save us from doing al the above hacks. 

There can be multiple cm for upgrade, its the same as the users are passing multiple set args in command. And these cm will be deleted once the upgrade resources are deleted. 

@tiagolobocastro  need your input here.

niladrih: I think I have covered this case in my comment above. file absolute paths are the same, we can avoid using the filename as the configmap key, and instead use the absolute path separated by '-' as the configmap key (which eventually turn into a file inside the container). It is impossible that two separate files will have the same absolute path.

avishnu: Can we not just use a tmp file name for the configmap? Usually tmp file names are reasonably unique and chances of collision is next to none. https://man7.org/linux/man-pages/man3/tmpfile.3.html

niladrih: Yes, the same logic applies here. However because we're not creating the file ourselves, and instead we're creating the configmap entry. We'd have to use a random filename/filename suffix instead. Yes, this will work.

sinhaashish: Just wanted to ask here , whats the issue with having one file per key?

avishnu: > Just wanted to ask here , whats the issue with having one file per key?

Yes, and the filename can be randomly generated using tmp file logic.

niladrih: > Just wanted to ask here , whats the issue with having one file per key?
@sinhaashish -- It is a copy of the same file...

> Yes, and the filename can be randomly generated using tmp file logic.
With <one-file-per-key> (this is not one-key-per-file), the file names are unique, as all helm values keys are unique. But, if multiple keys make use of data from the same file, then the file is copied for each of these keys. Because, one file per key, even if the file is the same file. The unique file name problem doesn't show up with one file per key.

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -563,12 +576,89 @@ impl UpgradeResources {
         Ok(())
     }
 
+    /// Create/Delete upgrade config map
+    pub async fn config_map_actions(
```
tiagolobocastro, line 580: Something like this?

```suggestion
    /// Create/Delete the upgrade config map used to store files specified by the set_file argument.
    pub async fn config_map_actions(
```

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -661,7 +757,10 @@ impl UpgradeResources {
         let args = &UpgradeArgs::default();
 
         // Delete the job
-        uo.job_actions(ns, Actions::Delete, args).await?;
+        _ = uo.job_actions(ns, Actions::Delete, args, None).await?;
```
tiagolobocastro, line 760: What is the return of job_actions?

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -833,3 +932,48 @@ pub(crate) async fn get_source_version(ns: &str) -> error::Result<String> {
...
+    upgrade_args: &UpgradeArgs,
+) -> error::Result<(BTreeMap<String, String>, HashMap<String, String>)> {
+    let mut data_map = BTreeMap::new();
+    let mut upgrade_map = HashMap::new();
+    let mut index = 1;
+    // for (index, file) in &upgrade_args.set_file .enumerate(){
+    for file in &upgrade_args.set_file {
+        let data: Vec<_> = file.split('=').collect();
+        let cm_values =
+            fs::read_to_string(data[1]).context(error::ReadFromFile { filepath: data[1] })?;
```
tiagolobocastro, line 947: this will panic if format incorrect..
suggest you do this at the top:
```rust
      let data: Vec<_> = file.split('=').collect();
      let [ _key, filepath ] = data[..] else {
          return Error .... not formatted properly;
      };
```

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -833,3 +932,48 @@ pub(crate) async fn get_source_version(ns: &str) -> error::Result<String> {
...
+/// Creat helm set file args.
+pub(crate) async fn create_helm_set_file_args(
+    upgrade_args: &UpgradeArgs,
+    set_file_map: Option<HashMap<String, String>>,
+) -> error::Result<Option<String>> {
+    if !&upgrade_args.set_file.is_empty() {
+        let mut helm_args_set_file = Vec::new();
+        for file in &upgrade_args.set_file {
+            let data: Vec<_> = file.split('=').collect();
+            let key = data[0].to_string();
```
tiagolobocastro, line 970: here to same as comment above..

k8s/upgrade/src/bin/upgrade-job/opts.rs
```diff
@@ -42,7 +42,12 @@ pub(crate) struct CliArgs {
     /// The set values specified by the user for upgrade
     /// (can specify multiple or separate values with commas: key1=val1,key2=val2).
     #[arg(short, long)]
-    values: String,
+    helm_args_set: String,
+
+    /// The set file values specified by the user for upgrade
+    /// (can specify multiple or separate values with commas: key1=path1,key2=path2).
+    #[arg(short, long)]
+    helm_args_set_file: String,
```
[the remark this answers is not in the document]
sinhaashish: done

k8s/upgrade/src/plugin/upgrade.rs
tiagolobocastro: You're still using data[1] rather than filepath, here and below

k8s/upgrade/src/plugin/upgrade.rs
```diff
@@ -833,3 +932,52 @@ pub(crate) async fn get_source_version(ns: &str) -> error::Result<String> {
...
+    if !&upgrade_args.set_file.is_empty() {
+        let mut helm_args_set_file = Vec::new();
+        for file in &upgrade_args.set_file {
+            let data: Vec<_> = file.split('=').collect();
+            let [ key, _filepath ] = data[..] else {
+                return error::InvalidSetFileArguments{ arguments : file.to_string()}.fail();
+            };
+            // Example : -- set-file jaeger-operator.tolerations=/root/tolerations.yaml
+            // gets converted to jaeger-operator.tolerations=/upgrade-config-map/1
+            let file = set_file_map.as_ref().unwrap().get(data[1]).unwrap();
```
tiagolobocastro, line 977: Let's handle the unwraps as well

tiagolobocastro (approved):

niladrih (approved):
LGTM

sinhaashish:
bors merge

PR #334, opened 2023-08-29
fix(upgrade-job): avoid quotes when setting yaml bools, null, int

niladrih:
This PR fixes a the case when yq sets a value other than a string on to a yaml file (i.e. a bool or a null). This removes the `"` character surrounding the set value when the value set is 'true', 'false', an int or 'null'.


avishnu (approved):
Looks ok. Do we need any better fix in the future?

niladrih:
> Looks ok. Do we need any better fix in the future?

I think we could make it look prettier, with better syntax. Functionally, we should be good. We could simplify by giving each of these types their own set function -- IMO this is unnecessary. WDYT @tiagolobocastro?

sinhaashish (approved):

niladrih:
bors merge

PR #336, opened 2023-09-01
chore(upgrade) : add volume node validation

sinhaashish:
Check for volumes which are somehow only affected by the particular node which are we are trying to upgrade.

## Description

This for the volume rebuild validation , while upgrading the  io-engine pod , do the volume validation for only those volumes which are present of the node currently being upgraded.

## Motivation and Context

This will make the upgrade faster , as it only checks for volume which are present of the io-engine node being upgraded .

## Regression

No

## Types of changes

## Checklist:

### DIFF
--- a/k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
@@ -10,7 +10,7 @@ use kube::{api::ObjectList, ResourceExt};
 use openapi::models::{Volume, VolumeStatus};
 use semver::{Version, VersionReq};
 use snafu::ResultExt;
-use std::time::Duration;
+use std::{collections::HashSet, time::Duration};
 use tracing::{info, warn};
 
 /// Contains the Rebuild Results.
@@ -24,6 +24,7 @@ pub(crate) struct RebuildResult {
 pub(crate) async fn rebuild_result(
     rest_client: &RestClientSet,
     stale_volumes: &mut Vec<Volume>,
+    node_name: &str,
 ) -> Result<RebuildResult> {
     loop {
         let unhealthy_volumes = list_unhealthy_volumes(rest_client, stale_volumes).await?;
@@ -32,33 +33,50 @@ pub(crate) async fn rebuild_result(
         }
 
         for volume in unhealthy_volumes.iter() {
-            match replica_rebuild_count(volume.clone()).await {
-                0 => {
-                    for _i in 0 .. 11 {
-                        // wait for a minute for any rebuild to start
-                        tokio::time::sleep(Duration::from_secs(60_u64)).await;
-                        let count = replica_rebuild_count(volume.clone()).await;
-                        if count > 0 {
-                            return Ok(RebuildResult {
-                                rebuilding: true,
-                                discarded_volumes: stale_volumes.clone(),
-                            });
-                        }
-                    }
-                    stale_volumes.push(volume.clone());
+            let target = if let Some(target) = volume.state.target.as_ref() {
+                target
+            } else {
+                continue;
+            };
+
+            let mut volume_over_nodes = HashSet::new();
+            volume_over_nodes.insert(target.node.as_str());
+
+            for (_, topology) in volume.state.replica_topology.iter() {
+                if let Some(node) = topology.node.as_ref() {
+                    volume_over_nodes.insert(node);
                 }
-                _ => {
-                    return Ok(RebuildResult {
-                        rebuilding: true,
-                        discarded_volumes: stale_volumes.clone(),
-                    })
+            }
+
+            if volume_over_nodes.contains(node_name) {
+                match replica_rebuild_count(volume) {
+                    0 => {
+                        for _i in 0 .. 11 {
+                            // wait for a minute for any rebuild to start
+                            tokio::time::sleep(Duration::from_secs(60_u64)).await;
+                            let count = replica_rebuild_count(volume);
+                            if count > 0 {
+                                return Ok(RebuildResult {
+                                    rebuilding: true,
+                                    discarded_volumes: stale_volumes.clone(),
+                                });
+                            }
+                        }
+                        stale_volumes.push(volume.clone());
+                    }
+                    _ => {
+                        return Ok(RebuildResult {
+                            rebuilding: true,
+                            discarded_volumes: stale_volumes.to_vec(),
+                        })
+                    }
                 }
             }
         }
     }
     Ok(RebuildResult {
         rebuilding: false,
-        discarded_volumes: stale_volumes.clone(),
+        discarded_volumes: stale_volumes.to_vec(),
     })
 }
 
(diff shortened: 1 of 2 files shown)


k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -24,6 +25,7 @@ pub(crate) struct RebuildResult {
 pub(crate) async fn rebuild_result(
     rest_client: &RestClientSet,
     stale_volumes: &mut Vec<Volume>,
+    node_name: &str,
```
niladrih, line 28: Is kubernetes node-name and mayastor node name always guaranteed to be the same? Are mayastor pools tied to the nodename or the `kubernetes.io/hostname` label value?
cc: @tiagolobocastro @Abhinandan-Purkait

niladrih (commented):
Left one comment..

niladrih (approved):

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -32,26 +34,41 @@ pub(crate) async fn rebuild_result(
...
+                .ok_or(
+                    EmptyVolumeSpecTarget {
+                        uuid: volume.spec.uuid.to_string(),
+                    }
+                    .build(),
+                )?
+                .node
+                .as_str();
+
+            if node_name == node {
```
dsharma-dc, line 50: Making sure if I understand correctly:

- Let's say we are upgrading node named `node3`. 
- A volume named `vol3` is on `node3` (has target there). `vol3` has replicas on `node3` and `node2`. 
- Now with the node name match check we'll be able to detect if e.g `vol3`'s replica on any other node e.g `node2` is rebuilding. `rebuilding` flag will be set `true`.
- In other scenario, if a volume `vol2` whose target is on e.g `node2` and has a rebuilding replica on `node3`, we won't be detecting that.

sinhaashish: > A volume named vol3 is on node3 (has target there). vol3 has replicas on node3 and node2.
You mean to say `vol3` has replicas on node1 and node2 ?

> Now with the node name match check we'll be able to detect if e.g vol3's replica on any other node e.g node2 is rebuilding. rebuilding flag will be set true.

Here the `rebuild_result` fetches the list of unhealthy volumes . For each unhealthy volume i check the node i.e. The node where front-end IO will be sent to , is the same as the node which is being upgraded.  If both are same then we find the `replica_rebuild_count` and wait for 10 mis  and all those logic . 


> In other scenario, if a volume vol2 whose target is on e.g node2 and has a rebuilding replica on node3, we won't be detecting that.
Yes

abhilashshetty04: We should probably check if there is any Healthy child part of rebuilding volume also in the node being upgraded. Otherwise we will end up Faulting one more child from same volume?

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -32,26 +34,41 @@ pub(crate) async fn rebuild_result(
...
+                .target
+                .as_ref()
+                .ok_or(
+                    EmptyVolumeSpecTarget {
+                        uuid: volume.spec.uuid.to_string(),
+                    }
+                    .build(),
+                )?
+                .node
+                .as_str();
```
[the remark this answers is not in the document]
sinhaashish: done

sinhaashish:
> We should check if the children are also located in the node to be drained, and not just the target node?

Made the change . PTAL

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -32,27 +33,42 @@ pub(crate) async fn rebuild_result(
...
+                    match replica_rebuild_count(volume.clone()).await {
+                        0 => {
+                            for _i in 0 .. 11 {
+                                // wait for a minute for any rebuild to start
+                                tokio::time::sleep(Duration::from_secs(60_u64)).await;
+                                let count = replica_rebuild_count(volume.clone()).await;
+                                if count > 0 {
+                                    return Ok(RebuildResult {
+                                        rebuilding: true,
+                                        discarded_volumes: stale_volumes.clone(),
```
tiagolobocastro, line 56: Do we need to clone here?

```suggestion
                                        discarded_volumes: stale_volumes,
```

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -32,27 +33,42 @@ pub(crate) async fn rebuild_result(
...
+                let mut volume_over_nodes = HashSet::new();
+                volume_over_nodes.insert(target.node.as_str());
+
+                for (_, topology) in volume.state.replica_topology.iter() {
+                    if let Some(node) = topology.node.as_ref() {
+                        volume_over_nodes.insert(node);
+                    }
+                }
+
+                if volume_over_nodes.contains(node_name) {
```
tiagolobocastro, line 46: Nice, so now we only consider volumes which "impact" this node :+1:

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -32,27 +33,42 @@ pub(crate) async fn rebuild_result(
...
-                    stale_volumes.push(volume.clone());
-                }
-                _ => {
-                    return Ok(RebuildResult {
-                        rebuilding: true,
-                        discarded_volumes: stale_volumes.clone(),
-                    })
                 }
+            } else {
+                continue;
```
tiagolobocastro, line 71: Reduce nesting by changing top to
```rust
let Some(target) = volume.state.target.as_ref() else {
  continue;
}
```

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -32,27 +33,42 @@ pub(crate) async fn rebuild_result(
...
+                    }
+                }
+
+                if volume_over_nodes.contains(node_name) {
+                    match replica_rebuild_count(volume.clone()).await {
+                        0 => {
+                            for _i in 0 .. 11 {
+                                // wait for a minute for any rebuild to start
+                                tokio::time::sleep(Duration::from_secs(60_u64)).await;
+                                let count = replica_rebuild_count(volume.clone()).await;
```
tiagolobocastro, line 52: Existing code but since we're here... seems `replica_rebuild_count` doesn't need the value, we could pass reference instead and also doesn't need to be async.

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
tiagolobocastro: Should we add volume to stale_volumes before returning?

tiagolobocastro (approved):

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
sinhaashish: done

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
sinhaashish: done

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
sinhaashish: no, the unhealthy volume  is added to  stale voume list only after the 10 mins.

sinhaashish:
bors merge

sinhaashish:
bors merge

PR #337, opened 2023-09-04
fix(upgrade-job): set defaults for absent helm values

niladrih:
The upgrade-job parses few specific values from the source mayastor helm chart's values.yaml. These are used to make conditional value 'sets'. Usually these values would always be available in the target helm chart, but it may be absent in the source helm chart (as it may have been added during later releases). It has to assume sane defaults for these values, so that the parse operation does not fail for the source helm chart.

## Description

This PR sets default values for 'csi.image.snapshotterTag', 'csi.image.snapshotControllerTag' and 'eventing.enabled'.

**N.B.:**
These default values are not used in the final values set used with the helm upgrade. They are simply used to conveniently make decisions about 'special' yq set operations (e.g. changed default values, etc.). Values yaml for upgrade is generated using yq.


avishnu (approved):

Abhinandan-Purkait (approved):
LGTM. Thanks

sinhaashish (approved):

PR #340, opened 2023-09-14
feat(snapshot, restore): prevent volume mode conversion for restore as default behaviour

Abhinandan-Purkait:
## Description
`--prevent-volume-mode-conversion`: Boolean that prevents an unauthorised user from modifying the volume mode when creating a PVC from an existing VolumeSnapshot. Only present as an alpha feature in v6.0.0 and above.

We enable the above flag on the csi-provisioner and the csi-snapshot-controller as a default behaviour.

## Motivation and Context
This change is needed so that the users cannot accidently change the volume mode, as changing mode from Block to Filesystem is destructive.

## Regression
No

## How Has This Been Tested?
This was tested manually. We would also add a bdd scenario to cover this in the e2e system tests.

## Types of changes

## Checklist:

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -293,6 +293,8 @@ csi:
     tolerations: []
     # -- Set PriorityClass, overrides global
     priorityClassName: ""
+    # -- Prevent modifying the volume mode when creating a PVC from an existing VolumeSnapshot
+    preventVolumeModeConversion: true
   node:
     logLevel: info
     topology:
--- a/chart/README.md
+++ b/chart/README.md
@@ -94,6 +94,7 @@ $ helm install my-release openebs/mayastor
 | base.&ZeroWidthSpace;logSilenceLevel | Silence specific module components | `nil` |
 | base.&ZeroWidthSpace;metrics.&ZeroWidthSpace;enabled | Enable the metrics exporter | `true` |
 | csi.&ZeroWidthSpace;controller.&ZeroWidthSpace;logLevel | Log level for the csi controller | `"info"` |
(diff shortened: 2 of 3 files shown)


chart/values.yaml
```diff
@@ -293,6 +293,8 @@ csi:
     tolerations: []
     # -- Set PriorityClass, overrides global
     priorityClassName: ""
+    # -- Set `--prevent-volume-mode-conversion` in csi-snapshot-controller and csi-provisioner to prevent modifying the volume mode when creating a PVC from an existing VolumeSnapshot
```
tiagolobocastro, line 296: How we do it is internal impl

```suggestion
    # -- Prevent modifying the volume mode when creating a PVC from an existing VolumeSnapshot
```

tiagolobocastro (approved):

chart/values.yaml
```diff
@@ -293,6 +293,8 @@ csi:
     tolerations: []
     # -- Set PriorityClass, overrides global
     priorityClassName: ""
+    # -- Prevent modifying the volume mode when creating a PVC from an existing VolumeSnapshot
+    preventVolumeModeConversion: true
```
niladrih, line 297: This option doesn't make it clear that this is a snapshot/snapshotController option. Should we put this under `csi.controller.snapshotController`? Also @tiagolobocastro wdyt?

niladrih (commented):
Left one comment.

chart/values.yaml
Abhinandan-Purkait: Actually it needs to put in two places for the functionality to work i.e in the external provisioner and the snapshot controller. That's why kept it a global csi controller value as both of them are a part of it.

niladrih: I missed that one. LGTM

niladrih (approved):

Abhinandan-Purkait:
bors merge

PR #343, opened 2023-09-20
fix: /var/local paths should be under release name

Abhinandan-Purkait:
## Description
Currently we use `/var/local/io-engine` and `/var/local/local-pv-hostpath/{{ .Release.Name }}`
This is a bit too spread we should bundle it all within a common `/var/local/{{ .Release.Name }}` example:-
`/var/local/{{ .Release.Name }}/{io-engine, localpv-hostpath}`

This PR changes the helm chart to include this.

## Regression
 No

## Types of changes

## Checklist:

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -386,7 +386,7 @@ etcd:
     # Name of etcd's localpv hostpath storage class.
     name: "mayastor-etcd-localpv"
     # -- Host path where local etcd data is stored in.
-    basePath: "/var/local/localpv-hostpath/{{ .Release.Name }}/etcd"
+    basePath: "/var/local/{{ .Release.Name }}/localpv-hostpath/etcd"
     # -- ReclaimPolicy of etcd's localpv hostpath storage class.
     reclaimPolicy: Delete
     # -- VolumeBindingMode of etcd's localpv hostpath storage class.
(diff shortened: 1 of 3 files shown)


chart/values.yaml
```diff
@@ -386,7 +386,7 @@ etcd:
     # Name of etcd's localpv hostpath storage class.
     name: "mayastor-etcd-localpv"
     # -- Host path where local etcd data is stored in.
-    basePath: "/var/local/localpv-hostpath/{{ .Release.Name }}/etcd"
+    basePath: "/var/local/{{ .Release.Name }}/localpv-hostpath/etcd"
```
tiagolobocastro, line 389: Presumably this doesn't have any issues for upgrades as the PV's were already created anyway, but new PV's will have this new path?
@niladrih we probably need to ensure this gets updated?

chart/values.yaml
```diff
@@ -486,7 +486,7 @@ loki-stack:
     # Name of loki's localpv hostpath storage class.
     name: "mayastor-loki-localpv"
     # -- Host path where local etcd data is stored in.
-    basePath: "/var/local/localpv-hostpath/{{ .Release.Name }}/loki"
+    basePath: "/var/local/{{ .Release.Name }}/localpv-hostpath//loki"
```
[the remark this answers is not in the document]
Abhinandan-Purkait: Damn. Thanks for pointing out. Done

chart/values.yaml
niladrih: Hmmm, so this should not get updated at upgrade-time, correct?

Abhinandan-Purkait: Yes, this can potentially patch the io-engine ds with the new path but we don't want that.

Abhinandan-Purkait: Right. But what about the mayastor-io-engine ds, if that gets patched with new path would it be a problem ?

tiagolobocastro: As things are today no I don't think so.

niladrih: Alright then, then I'm going with no-change because both no-change and change seem harmless.

niladrih (approved):
_“What's in a {{ .Release.Name }}? That which we call a rose by any other name would smell just as sweet.”_ - William Shakespeare

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #347, opened 2023-09-29
feat(tests/minikube): add cleanup script and refactor setup

niladrih:
Adds a cleanup script for the minikube. This cleanup.sh works only towards cleanup up after the most recent setup.sh run. Also, it won't work, if you haven't run setup.sh after this change (in case you check out by branch ;)).

## Description

Changes:
- use a yaml file to store cleanup configuration
- add an advisory locking framework so that setup and cleanup access
cleanup config exclusively
- add change so cni plugins are only installed if absent
- use 'which' instead of 'command -v' as 'command -v' output
may be out of date
- add helm, yq and which to nix-shell
- create a separate logging library
- use SCRIPT_DIR/bin instead of /usr/local/bin for prereq binaries
- enable prerequisite installation enabled by default
- moved all tests files into the tests/bdd directory
- other minor changes

## How Has This Been Tested?
Tested manually


tests/minikube/setup.sh
```diff
@@ -139,70 +196,108 @@ github_asset_binary() {
...
       # unlikely if cri-dockerd wasn't already installed, it is possible.
-      cri_dockerd_service_error=""
+      local cri_dockerd_service_error
       systemctl is-active --quiet cri-docker.socket || cri_dockerd_service_error=$?
       if [ -n "${cri_dockerd_service_error}" ]; then
         # Download systemd service files
-        cri_dockerd_version="v$(cri-dockerd --version 2>&1 | awk '{print $2}')"
-        systemd_service_files_url="https://raw.githubusercontent.com/Mirantis/cri-dockerd/${cri_dockerd_version}/packaging/systemd"
-        cd ${workspace}
+        local -r cri_dockerd_version="v$(cri-dockerd --version 2>&1 | awk '{print $2}')"
```
tiagolobocastro, line 252: Wondering if we're doing too much here, and simply set pre-requisite for user to set this up before hand..

but anyways, you might want to check first if the system is using systemd (might not be, depending on the linux distro)

tests/minikube/setup.sh
```diff
@@ -236,47 +331,25 @@ install_prerequisites() {
...
-    dl_link_arch_path="amd64"
-    ;;
-  "arm64" | "aarch64")
-    dl_link_arch_path="arm64"
-    ;;
-  *)
-    die "the 'install_kubectl' option does not support the arch ${arch}" 1
-    ;;
-  esac
+  ln -sf $(which minikube) "${bin_dir}/kubectl"
```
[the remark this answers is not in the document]
niladrih: There's `minikube kubectl` that minikube sets up([link](https://minikube.sigs.k8s.io/docs/handbook/kubectl/)), so it's basically kubectl that ONLY works with minikube. The solution to going back to your minikube might be to tear down the minikube using the cleanup. 

As a side note:
Our use-case -- running mayastor, requires privileged containers. And user-mode systemd can't give you that, unless the user is root. So, to work around all of this, I'm thinking of going with kvm-based minikube.  
This none driver minikube continues to exist, the binaries/symlinks don't go to /usr/local/bin, and nothing gets moved. PATH manipulation should be sufficient. This none minikube works well for pipelines.  
The local testing tool will be kvm minikube with disk support. I'll try to poke at this during the weekend.

tests/minikube/cleanup.sh
```diff
@@ -0,0 +1,76 @@
...
+die() {
+  local -r _return="${2:-1}"
+  test "${_PRINT_HELP:-no}" = yes && print_help >&2
+  echo_stderr "ERROR: $1"
+  exit "${_return}"
+}
+
+# Takes a lock by way of a local file with the name <cleanup-config-file-name>.lock.
+# The setup script honours this lock.
+lock_cleanup_config_file() {
```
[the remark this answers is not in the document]
niladrih: > why do we need a lock file?
So that the cleanup config is not altered by an on-going setup/cleanup while one is already underway.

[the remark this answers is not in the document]
niladrih: Moved this common code to a lib called 'advisory_lock' in scripts/utils.

tests/minikube/setup.sh
```diff
@@ -87,46 +143,47 @@ parse_args() {
...
-  must_succeed_command "command -v ${1}" "${error}"
+  local -r error="${2:-command ${1} not present in PATH}"
+  must_succeed_command "which ${1}" "${error}"
 }
 
 # Get latest release tag (not un-released) for a GitHub repo using GitHub's api.
 github_latest_version_tag() {
-  local github_org=${1}
-  local github_repo=${2}
+  local -r github_org=${1}
```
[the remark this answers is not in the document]
niladrih: `-r` makes the variable read-only.

tests/minikube/setup.sh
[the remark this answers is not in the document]
niladrih: I've reverted back to using a binary instead of`minikube kubectl`. Also, staying out of /usr, and using a local bin directory instead.

tests/minikube/setup.sh
niladrih: Added a check to see if 'systemctl' exists, which is a good sign that it's a systemd Linux system.
>Wondering if we're doing too much here, and simply set pre-requisite for user to set this up before hand
This might come in handy for CI runs with this script. I've made the installations enabled by default. Also, added a `-s` flag option to skip prerequiste installation.

I plan to add in a kvm-based profile for a personal-use scenario. Keep this in TODO, for now.

niladrih: > Added a check to see if 'systemctl' exists, which is a good sign that it's a systemd Linux system.

Removed it. There is already a check for this, I'd forgotten about it:
```
NIX_SHELL_PREREQUISITES=("conntrack" "minikube" "crictl" "curl" "jq" "yq" "awk" "systemctl" "docker")
```
```
      for bin in "${NIX_SHELL_PREREQUISITES[@]}"; do
        must_exists_in_path "${bin}" >/dev/null
      done
```

niladrih:
This script doesn't work with NixOS, as it tries to change files inside `/etc`. The Jenkins CI uses a NixOS VM, so it needs to be able work with that. This script needs support for kvm-mode minikube. This also lets this script spin up multi-node k8s clusters!  

Work in progress

niladrih:
I'll add in the kvm change in a separate PR.

.gitignore
```diff
@@ -17,4 +17,9 @@ __pycache__
 /tests/bdd/autogen/
 /rust-toolchain.toml
 /.rust-toolchain
-/kubectl-plugin
\ No newline at end of file
+/kubectl-plugin
```
tiagolobocastro, line 20: Is this where you download the kubectl-plugin binary?
I'd suggest keeping a directory for all downloaded binaries to keep it clean

tests/bdd/minikube/setup.sh
```diff
@@ -0,0 +1,460 @@
...
+        systemctl daemon-reload
+        systemctl enable --now --quiet cri-docker.socket
+        systemctl is-active --quiet cri-docker.socket || die "failed to set up cri-dockerd systemd service" 1
+
+        # Cleanup entry for cri-docker.socket systemd service
+        prepend_cleanup_script "# Disables and stops the systemd services
+systemctl disable --now --quiet cri-docker.socket
+systemctl disable --now --quiet cri-docker
+systemctl daemon-reload
+# Delete the service and configuration files for the systemd services
```
tiagolobocastro, line 247: all the bits from both scripts that deal with etc, maybe we should ensure they're not tried on read-only /etc/systemd directories?

tiagolobocastro (approved):

.gitignore
niladrih: I didn't put it there. It was there when I found it, the new-line at the end is making it seem like a new line. Not a binary download for minikube.

tests/bdd/minikube/setup.sh
niladrih: There's a `--skip-prerequisites` option to stop these things from starting. Apart from that, I plan to make the KVM-mode minikube the default one.

.gitignore
tiagolobocastro: hmm oh it's for the nix-build, if you don't specify arguments it puts them on current directory :D

tests/bdd/minikube/setup.sh
niladrih: Added a note for read-only /etc/systemd/system systems in the README.md

Abhinandan-Purkait (approved):

niladrih:
bors merge

PR #348, opened 2023-10-05
fix(workflows/main_chart): update GitHub token secret

niladrih:
Update GitHub token secret to enable pushes to openebs/mayastor-chart-donotuse


niladrih:
These changes are included in #351 .

PR #353, opened 2023-10-08
fix(upgrade-job): remove short args to resolve conflict over '-h' option

niladrih:
Two clap args introduced in https://github.com/openebs/mayastor-extensions/pull/330 make use the short form option argument. These two arguments -- 'helm_args_set' and 'helm_args_set_file' begin with 'h'. Both of these two args, as well the '--help' option try to make use of '-h'. This change removes the short form arg for these two args.

Remove short args for 'helm_args_set' and 'helm_args_set_file' CliArgs.

## Motivation and Context

`upgrade-job -h` fails.

## Regression

Yes.

 Fixes a regression introduced in  https://github.com/openebs/mayastor-extensions/pull/330
## How Has This Been Tested?

Trivial change, tested the successful execution of `upgrade-job -h`.

### DIFF
--- a/k8s/upgrade/src/bin/upgrade-job/opts.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/opts.rs
@@ -41,12 +41,12 @@ pub(crate) struct CliArgs {
 
     /// The set values specified by the user for upgrade
     /// (can specify multiple or separate values with commas: key1=val1,key2=val2).
-    #[arg(short, long)]
+    #[arg(long)]
     helm_args_set: String,
 
     /// The set file values specified by the user for upgrade
     /// (can specify multiple or separate values with commas: key1=path1,key2=path2).
-    #[arg(short, long)]
+    #[arg(long)]
     helm_args_set_file: String,
 }
 


sinhaashish (approved):

k8s/upgrade/src/bin/upgrade-job/opts.rs
dsharma-dc, line 1: Or perhaps you can retain short form with a different letter like structopt allows

```
    #[arg(short = "s", long)]
    helm_args_set: String,

    #[arg(short = "f", long)]
    helm_args_set_file: String,
```

dsharma-dc (approved):

tiagolobocastro (approved):
This would have been caught by an upgrade test in the repo... we're close to getting there :)

k8s/upgrade/src/bin/upgrade-job/opts.rs
niladrih: Yes, but we simply aren't using the shortform. Structopt is clap-derive API. Good one thought. Thanks.

niladrih:
> This would have been caught by an upgrade test in the repo... we're close to getting there :)

Yuss 💯

PR #360, opened 2023-10-24
chore(chart): bump snapshot and registrar csi sidecar versions

niladrih:
This bumps the csi sidecars for external-snapshotter and node-driver-registrar to minimum versions which support csi spec v1.8.0. The other sidecars are at par.


Abhinandan-Purkait:
Question: What is the version that we support at the csi level? Do we not need to update the protobufs?

tiagolobocastro:
Our Protobuf is 1.5 but 1.8 is not a breaking change so we're ok?
@niladrih could you please run e2e basic and snapshot tests and see if it's all good with these?

niladrih:
Hi folks, I've tested, and tests seem to pass with these version bumps.

Abhinandan-Purkait (approved):
LGTM

niladrih:
bors merge

PR #381, opened 2023-12-05
feat(upgrade): add special values options for 2.5.0 

niladrih:
In 2.5.0, the scrapeConfigs options is changed (ref: #344). This adds values changes for upgrade when going from a source version <2.5.0 to a target version >=2.5.0.

## Description

Changes:
- Sets .promtail.config.snippets.scrapeConfigs for upgrade from anywhere in the range 2.0.0-rc.1 - 2.4.0.
- Changes loki scrapeConfig to use single quotes, because [the code here](https://github.com/openebs/mayastor-extensions/blob/develop/k8s/upgrade/src/bin/upgrade-job/helm/yaml/yq.rs#L178) adds double quotes to strings, avoiding bools, null and ints. When double quotes are added to scrapeConfigs, it'll require the double quotes withing the scrapeConfigs to use escapes (\\").

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -588,7 +588,7 @@ loki-stack:
               - docker: {}
               - replace:
                   expression: '(\n)'
-                  replace: ""
+                  replace: ''
               - multiline:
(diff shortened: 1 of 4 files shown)


tiagolobocastro (approved):

abhilashshetty04 (approved):

niladrih:
bors merge

PR #391, opened 2023-12-12
fix: fix upgrade

niladrih:
Change:
- Use yq's strenv operator to use special characters with yq when updating the upgrade values.yaml.
- Fix check which forbids upgrade from a stable release to unstable version.
- Prefers the parameters of an existing default StorageClass over the chart defaults.

### DIFF
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -205,3 +205,30 @@ Usage:
         {{- end -}}
     {{- end -}}
 {{- end -}}
+
+{{/*
+    Generate the default StorageClass parameters.
+    This is required because StorageClass parameters cannot be patched after creation.
+    If the StorageClass already exists, the default StorageClass carries the parameters and values
+    of that StorageClass. Else, it carries the default parameters and values.
+*/}}
+{{- define "storageClass.parameters" -}}
+    {{- $scName := index . 0 -}}
+    {{- $valuesParams := index . 1 -}}
+
+    {{/* Check to see if a default StorageClass already exists */}}
+    {{- $sc := lookup "storage.k8s.io/v1" "StorageClass" "" $scName -}}
+
+    {{- if $sc -}}
+        {{/* Existing defaults */}}
+        {{ range $param, $val := $sc.parameters }}
+{{ $param | quote }}: {{ $val | quote }}
+        {{- end -}}
+
+    {{- else -}}
+        {{/* Current defaults */}}
+        {{ range $param, $val := $valuesParams }}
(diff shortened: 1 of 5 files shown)


chart/templates/_helpers.tpl
```diff
@@ -205,3 +205,28 @@ Usage:
...
+    {{- $sc := lookup "storage.k8s.io/v1" "StorageClass" "" $scName -}}
+    {{- if $sc -}}
+        {{ range $param, $val := $sc.parameters }}
+{{ $param | quote }}: {{ $val | quote }}
+        {{- end -}}
+    {{- else -}}
+{{/*
+    Latest defaults
+*/}}
+protocol: {{ $valuesProtocol | quote }}
```
tiagolobocastro, line 229: this we could take from values.yaml and avoid modifying the template again in the future

k8s/upgrade/src/plugin/preflight_validations.rs
```diff
@@ -186,6 +186,16 @@ impl TryFrom<&[u8]> for UnsupportedVersions {
...
 }
 
+/// Strips the prefix 'v' from a semver-like literal, e.g.: v1.2.3 -> 1.2.3.
+/// The Version crate doesn't work with the 'v' prefix.
+pub(crate) fn strip_v_prefix(version: &str) -> &str {
+    if version.starts_with('v') {
+        return version.strip_prefix('v').unwrap_or(version);
+    }
+
+    version
```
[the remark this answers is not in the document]
niladrih: Made this change.

chart/templates/_helpers.tpl
niladrih: Good one, made this change.

chart/templates/_helpers.tpl
```diff
@@ -205,3 +205,30 @@ Usage:
...
+    {{- if $sc -}}
+        {{/* Existing defaults */}}
+        {{ range $param, $val := $sc.parameters }}
+{{ $param | quote }}: {{ $val | quote }}
+        {{- end -}}
+
+    {{- else -}}
+        {{/* Current defaults */}}
+        {{ range $param, $val := $valuesParams }}
+{{ $param | quote }}: {{ $val | quote }}
```
tiagolobocastro, line 231: Does param need quote?

tiagolobocastro (approved):

chart/templates/_helpers.tpl
niladrih: Yes, I think so. It is a map[string]string.

w3aman (approved):
Tested this upgrade on k8s cluster

PR #395, opened 2023-12-21
chore(scripts/utils): remove set options

niladrih:
The utils scripts are meant to be imported by other scripts. Setting/unsetting
set options on these scripts may interfere with the set options on the calling
script. The set options aren't an essential part of the scripts, removing them.

### DIFF
--- a/scripts/utils/advisory_lock.sh
+++ b/scripts/utils/advisory_lock.sh
@@ -1,13 +1,6 @@
 #!/usr/bin/env bash
 
-# -o errexit: abort script if one command fails
-# -o errtrace: the ERR trap is inherited by shell functions
-# -o pipefail: entire command fails if pipe fails
-# -o history: record shell history
-# -o allexport: export all functions and variables to be available to subscripts
-set -o errexit -o errtrace -o pipefail -o history -o allexport
-
-# Takes a lock by way of a local file with the name <cleanup-config-file-name>.lock.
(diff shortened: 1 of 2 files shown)


hrudaya21 (approved):

scripts/utils/advisory_lock.sh
```diff
@@ -35,4 +35,6 @@ advisory_lock_remove() {
     rm -f "$lock_filepath"
     echo "Removed the lock on the file $filepath."
   fi
-}
\ No newline at end of file
+}
+
+set +o errexit +o errtrace +o pipefail +o history +o allexport
```
tiagolobocastro, line 40: I'm no shell expert but seems this also can cause issues, if calling script had any of these set it will unset them?

hrudaya21: I think it is possible. Possibly, We need to judge before setting the flag in the begining and unset it accordingly.

niladrih: If the import on the calling script is after their `set`s, then yes. Should we remove the sets on these utils scripts, so these have no options set, and the calling script controls all options?

niladrih: Removed all set options, this should work best when importing these.

tiagolobocastro (approved):
So long as the code here is not relying on any of the options its fine, otherwise we need another solution

niladrih:
bors merge

PR #397, opened 2023-12-22
fix(upgrade-job): avoid infinite loop when no unhealthy vols have a target

niladrih:
Fixes a bug where if, during data-plane upgrade, `rebuild_result` function finds unhealthy volumes, but none of the volumes have a target, then, the function continues to try to list unhealthy volumes constantly.

Adds a break from the infinite loop for an empty target accumulator.

## Description

Adds a break from the infinite loop for an empty target hashset.

## Types of changes

## Checklist:

### DIFF
--- a/k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
@@ -33,14 +33,13 @@ pub(crate) async fn rebuild_result(
             break;
         }
 
+        let mut volume_over_nodes = HashSet::new();
         for volume in unhealthy_volumes.iter() {
-            let target = if let Some(target) = volume.state.target.as_ref() {
-                target
-            } else {
-                continue;
+            let target = match volume.state.target.as_ref() {
+                Some(t) => t,
+                None => continue,
             };
 
-            let mut volume_over_nodes = HashSet::new();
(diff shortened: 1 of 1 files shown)


Abhinandan-Purkait (approved):

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
```diff
@@ -74,6 +73,9 @@ pub(crate) async fn rebuild_result(
                 }
             }
         }
+        if volume_over_nodes.is_empty() {
+            break 'outermost;
```
tiagolobocastro, line 77: why do we need to specify it, aren't we inside only outermost here?

niladrih: Just for readability. I can remove it.

niladrih: Removed loop label and break targets.

tiagolobocastro (approved):

niladrih:
bors merge

k8s/upgrade/src/bin/upgrade-job/upgrade/utils.rs
tiagolobocastro: It's fine to keep, just got confused why it was there.

PR #399, opened 2023-12-24
fix: upgrade-job fixes

niladrih:
Changes:
1. The localpv-provisioner helm chart was updated to v3.5.0 with the release of Mayastor v2.5.0. The yq merge would retain the release-version and image tags on the localpv helm values. They need to be manually overriden.
2. The scrapeConfigs change is only applied if the source configuration is the older default. It should move to newer default, only if the chart had the older default to begin with. This prevents overriding user-defined custom configuration.


tiagolobocastro (approved):

sinhaashish (approved):

niladrih:
bors merge

PR #396, opened 2023-12-22
feat(upgrade-job): add support for loki-stack helm chart upgrade

niladrih:
Requires #386.

Overview:
- adds new yq commands to delete objects, append to objects, and append to arrays.
- adds special upgrade values.yaml update for loki-stack changes from #386. Introduces migrations and set_literal_value updates to alter the existing yaml till it resembles the new one.
- replaces all `&Path` is the `helm::yaml::yq` module with `AsRef<Path>`.

Changes:
- removes .loki-stack.loki.config.ingester.lifecycler.ring.kvstore as it
does not exist in loki-stack v2.9.11
- adds set value for the grafana/loki container image tag
- removes .loki-stack.promtail.config.snippets.extraClientConfigs as it
does not exist in loki-stack v2.9.11
- removes .loki-stack.promtail.initContainer as it does not exist in loki-stack
v2.9.11
- migrates .loki-stack.promtail.config.lokiAddress to .loki-stack.promtail.config.clients
- adds yq command_output helper function
- add set value for .loki-stack.promtail.readinessProbe.httpGet.path
- migrate .loki-stack.promtail.config.snippets.extraClientConfigs to
.loki-stack.promtail.config.clients

PR #401, opened 2024-01-08
chore(scripts): add consolidated helm values generator

niladrih:
This adds a script which embeds the missing values.yaml bits from all of the dependent charts on to the values.yaml of the chart which imports it. This done so that the consolidated values may be deserialized (during upgrade) to real values for yaml bits which have been omitted in the higher values.yaml(s) for brevity.

### DIFF
--- a/scripts/helm/generate-consolidated-values.sh
+++ b/scripts/helm/generate-consolidated-values.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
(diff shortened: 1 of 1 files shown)


tiagolobocastro (commented):
How is this used?
Probably this is not `chore`, but maybe `ci(helm-chart/scripts)` or something like that

niladrih:
Closing this PR. These changes are included in PR #396 .

PR #414, opened 2024-02-06
refactor(chart): provide CRDs in a subchart

cmontemuino:
## Description

This changeset is mainly intended to skip Jaeger CRDs if it is not enabled. It is especially relevant when Jaeger tracing is installed separately from OpenEBS Mayastor.

Moving CRDs to a chart solves several issues: clean uninstall, possibility for upgrades, templating.

## Motivation and Context

Helm 3 does not manage CRDs (see https://helm.sh/docs/chart_best_practices/custom_resource_definitions/). `helm uninstall` won't remove CRDs, and `helm updgrade` won't upgrade them. Manual intervention is required with the current setup.
 - CRDs installed from `crds` folder **are not included into the helm release**.

Additionally, it is not possible to template CRDs in Helm 3 in the `crds` folder.

## Regression

No

## How Has This Been Tested?

Same results as before this MR:

```console
helm install --set crds.jaeger.enabled=true mayastor
```

Skip the installation of Jaeger CRDs:

```console
helm install mayastor
```

## Types of changes

  - https://github.com/openebs/mayastor/issues/1586

## Checklist:

- [ ] ~~I have added unit tests to cover my changes.~~
  - It seems like this project is not using `helm-unittest`

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -33,17 +33,28 @@ $ helm install mayastor . -n <mayastor-namespace>
 
 ### Installing the Chart via Helm Registry
 
-To install the chart with the release name `my-release`:
+To install the chart with the release name `mymayastor`:
 
 ```console
-$ helm repo add openebs https://openebs.github.io/mayastor-extensions/
-$ helm install my-release openebs/mayastor
+$ helm repo add mayastor https://openebs.github.io/mayastor-extensions/
+$ helm install mymayastor mayastor/mayastor
 ```
+
+### Uninstall Helm Chart
+
+```console
+$ helm uninstall [RELEASE_NAME] 
+```
+
+This removes all the Kubernetes components associated with the chart and deletes the release.
+
+*See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall/) for command documentation.*
 
 ## Chart Dependencies
 
 | Repository | Name | Version |
 |------------|------|---------|
+|  | crds | 0.0.0 |
 | https://charts.bitnami.com/bitnami | etcd | 8.6.0 |
 | https://grafana.github.io/helm-charts | loki-stack | 2.9.11 |
 | https://jaegertracing.github.io/helm-charts | jaeger-operator | 2.25.0 |
@@ -93,6 +104,8 @@ $ helm install my-release openebs/mayastor
 | base.&ZeroWidthSpace;jaeger.&ZeroWidthSpace;enabled | Enable jaeger tracing | `false` |
 | base.&ZeroWidthSpace;logSilenceLevel | Silence specific module components | `nil` |
 | base.&ZeroWidthSpace;metrics.&ZeroWidthSpace;enabled | Enable the metrics exporter | `true` |
+| crds.&ZeroWidthSpace;enabled | Install CRDs | `true` |
+| crds.&ZeroWidthSpace;jaeger.&ZeroWidthSpace;enabled | Install Jaeger CRDs | `false` |
 | csi.&ZeroWidthSpace;controller.&ZeroWidthSpace;logLevel | Log level for the csi controller | `"info"` |
 | csi.&ZeroWidthSpace;controller.&ZeroWidthSpace;preventVolumeModeConversion | Prevent modifying the volume mode when creating a PVC from an existing VolumeSnapshot | `true` |
 | csi.&ZeroWidthSpace;controller.&ZeroWidthSpace;priorityClassName | Set PriorityClass, overrides global | `""` |
--- a/chart/README.md
+++ b/chart/README.md
@@ -43,7 +43,7 @@ $ helm install mymayastor mayastor/mayastor
 ### Uninstall Helm Chart
 
 ```console
-$ helm uninstall [RELEASE_NAME] 
+$ helm uninstall [RELEASE_NAME]
 ```
 
 This removes all the Kubernetes components associated with the chart and deletes the release.
--- a/chart/charts/crds/templates/csi-volume-snapshot-class.yaml
+++ b/chart/charts/crds/templates/csi-volume-snapshot-class.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.volumeSnapshots.enabled -}}
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
@@ -145,3 +146,4 @@ status:
     plural: ""
   conditions: []
   storedVersions: []
+{{- end -}}
--- a/chart/charts/crds/values.yaml
+++ b/chart/charts/crds/values.yaml
@@ -1,0 +1,6 @@
+jaeger:
+  # -- Install Jaeger CRDs
+  enabled: true
+  # -- Install Volume Snapshots CRD
+  volumeSnapshots:
+    enabled: true
--- a/chart/charts/crds/values.yaml
+++ b/chart/charts/crds/values.yaml
@@ -1,6 +1,6 @@
 jaeger:
   # -- Install Jaeger CRDs
   enabled: true
-  # -- Install Volume Snapshots CRD
-  volumeSnapshots:
-    enabled: true
+# -- Install Volume Snapshots CRD
+volumeSnapshots:
+  enabled: true
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,9 @@
 /result*
 __pycache__
 /tests/bdd/openapi/
-/chart/charts/
+/chart/charts/*
+# charts/crds is valid chart for git
+!/chart/charts/crds/
 /chart/Chart.lock
 !/openapi/Cargo.toml
 !/openapi/build.rs
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -25,6 +25,9 @@ version: 0.0.0
 appVersion: "0.0.0"
 
 dependencies:
+  - name: crds
+    version: "0.0.0"
+    condition: crds.enabled
   - name: etcd
     repository: https://charts.bitnami.com/bitnami
     version: 8.6.0
--- a/chart/README.md.tmpl
+++ b/chart/README.md.tmpl
@@ -32,12 +32,22 @@ $ helm install mayastor . -n <mayastor-namespace>
 
 ### Installing the Chart via Helm Registry
 
-To install the chart with the release name `my-release`:
+To install the chart with the release name `mymayastor`:
 
 ```console
-$ helm repo add openebs https://openebs.github.io/mayastor-extensions/
-$ helm install my-release openebs/{{ template "chart.name" . }}
+$ helm repo add mayastor https://openebs.github.io/mayastor-extensions/
+$ helm install mymayastor mayastor/{{ template "chart.name" . }}
 ```
+
+### Uninstall Helm Chart
+
+```console
+$ helm uninstall [RELEASE_NAME] 
+```
+
+This removes all the Kubernetes components associated with the chart and deletes the release.
+
+*See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall/) for command documentation.*
 
 ## Chart Dependencies
 
--- a/chart/charts/crds/Chart.yaml
+++ b/chart/charts/crds/Chart.yaml
@@ -1,0 +1,7 @@
+apiVersion: v2
+name: crds
+version: 0.0.0
+description: |
+  A Helm chart that collects custom resource definitions (CRDs) from Mayastor. It also includes one CRD
+  from Jaeger tracing when enabled.
+
--- a/chart/charts/crds/README.md
+++ b/chart/charts/crds/README.md
@@ -1,0 +1,4 @@
+# Mayastor CRDs
+
+This chart brings Custom Resource Definitions (CRDs) used by Mayastor.
+
--- a/chart/charts/crds/templates/csi-volume-snapshot-content.yaml
+++ b/chart/charts/crds/templates/csi-volume-snapshot-content.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.volumeSnapshots.enabled -}}
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
@@ -483,3 +484,4 @@ status:
     plural: ""
   conditions: []
   storedVersions: []
+{{- end -}}
--- a/chart/charts/crds/templates/csi-volume-snapshot.yaml
+++ b/chart/charts/crds/templates/csi-volume-snapshot.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.volumeSnapshots.enabled -}}
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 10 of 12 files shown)


tiagolobocastro:
Seems the commit lint must have a bug as it's crashing :)
Anyhow, I think the fix here is to rename:
fixup! chore: allow chart/charts/crds
to
chore: allow chart/charts/crds
or perhaps
feat: allow chart/charts/crds

@cmontemuino any pros/cons to having the CRD files as a sub-chart, rather than just using a plain crd: true/false and checking for this in the yaml template?

@niladrih please take a look at this PR

chart/README.md
```diff
@@ -37,13 +37,24 @@ To install the chart with the release name `my-release`:
 
 ```console
 $ helm repo add openebs https://openebs.github.io/mayastor-extensions/
-$ helm install my-release openebs/mayastor
+$ helm install mymayastor
```
niladrih: Is this a valid command?

niladrih (commented):
@cmontemuino -- I understand if you want to move the csi-snapshot CRDs to the `chart/template` directory. What is the motivation for moving them into a subchart? Maybe a new directory inside the template directory would suffice for this use-case?

chart/README.md
cmontemuino: my bad. Thanks!!

cmontemuino: addressed in 6ff16f7

cmontemuino:
> Seems the commit lint must have a bug as it's crashing :) Anyhow, I think the fix here is to rename: fixup! chore: allow chart/charts/crds to chore: allow chart/charts/crds or perhaps feat: allow chart/charts/crds
> 
> @cmontemuino any pros/cons to having the CRD files as a sub-chart, rather than just using a plain crd: true/false and checking for this in the yaml template?
> 
> @niladrih please take a look at this PR

thanks for pointing this out. I've just squashed the offending commit

cmontemuino:
> @cmontemuino -- I understand if you want to move the csi-snapshot CRDs to the `chart/template` directory. What is the motivation for moving them into a subchart? Maybe a new directory inside the template directory would suffice for this use-case?

@niladrih do you mean adding the Jaeger CRD as a template?

cmontemuino:
> @cmontemuino any pros/cons to having the CRD files as a sub-chart, rather than just using a plain crd: true/false and checking for this in the yaml template?

@tiagolobocastro using Helm 3 approach to install CRDs forces you to upgrade CRDs manually (Helm won't upgrade them if you touch the files). That's problematic when doing GitOps.

That said, the main problem here is to avoid installing the Jaeger CRD. It is not possible to template CRDs in the `crd` folder.

A better approach here would be to publish CRDs in a separate chart. This PR is about applying the minimum required changes.

niladrih:
> > @cmontemuino -- I understand if you want to move the csi-snapshot CRDs to the `chart/template` directory. What is the motivation for moving them into a subchart? Maybe a new directory inside the template directory would suffice for this use-case?
> 
> @niladrih do you mean adding the Jaeger CRD as a template?

That's right. Something like this:
```yaml
{{- if and .Values.crds.enabled .Values.base.jaeger.enabled .Values.crds.jaeger-}}
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: jaegers.jaegertracing.io
...
{{- end -}}
```
```yaml
{{- if and .Values.crds.enabled .Values.crds.volumeSnapshots-}}
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: volumesnapshots.snapshot.storage.k8s.io
...
{{- end -}}
```

```sh
chart/
├── crds
└── templates/
    └── crds/
        ├── jaeger.yaml
        ├── volumesnapshots.yaml
        ├── volumesnapshotclasses.yaml
        └── volumesnapshotcontents.yaml
```

```yaml
crds:
  # Toggle installation for all CRDs. A 'false' disables installation for all.
  enabled: true
  # Toggle installation for Kubernetes CSI VolumeSnapshot CRDs.
  volumeSnapshots: true
  # Toggle installation for Jaeger CRD(s). 
  jaeger: true
```

WDYT @cmontemuino @tiagolobocastro? Is the `crds.enabled` a little too much?

cmontemuino:
> > > @cmontemuino -- I understand if you want to move the csi-snapshot CRDs to the `chart/template` directory. What is the motivation for moving them into a subchart? Maybe a new directory inside the template directory would suffice for this use-case?
> > 
> > 
> > @niladrih do you mean adding the Jaeger CRD as a template?
> 
> That's right. Something like this:
> 
> ```yaml
> {{- if and .Values.crds.enabled .Values.base.jaeger.enabled .Values.crds.jaeger-}}
> apiVersion: apiextensions.k8s.io/v1
> kind: CustomResourceDefinition
> metadata:
>   name: jaegers.jaegertracing.io
> ...
> {{- end -}}
> ```
> 
> ```yaml
> {{- if and .Values.crds.enabled .Values.crds.volumeSnapshots-}}
> apiVersion: apiextensions.k8s.io/v1
> kind: CustomResourceDefinition
> metadata:
>   name: volumesnapshots.snapshot.storage.k8s.io
> ...
> {{- end -}}
> ```
> 
> ```shell
> chart/
> ├── crds
> └── templates/
>     └── crds/
>         ├── jaeger.yaml
>         ├── volumesnapshots.yaml
>         ├── volumesnapshotclasses.yaml
>         └── volumesnapshotcontents.yaml
> ```
> 
> ```yaml
> crds:
>   # Toggle installation for all CRDs. A 'false' disables installation for all.
>   enabled: true
>   # Toggle installation for Kubernetes CSI VolumeSnapshot CRDs.
>   volumeSnapshots: true
>   # Toggle installation for Jaeger CRD(s). 
>   jaeger: true
> ```
> 
> WDYT @cmontemuino @tiagolobocastro? Is the `crds.enabled` a little too much?

Hi @niladrih! What you proposed could work for this repo.
I've checked the CRDs are being referenced from a ClusterRole only, that that's ok with Helm's installation order:
- https://github.com/openebs/mayastor-extensions/blob/develop/chart/templates/mayastor/rbac/rbac.yaml 
- https://github.com/helm/helm/blob/release-3.14/pkg/releaseutil/kind_sorter.go#L31

That said, I don't really know what could happen in the [operator]( https://github.com/openebs/mayastor-extensions/blob/develop/chart/templates/mayastor/operators/operator-diskpool-deployment.yaml) is consuming some of the CRDs, especially in the case of an upgrade when removing/adding APIs.

There is also the option to use chart hooks, but they don't come free of troubles during upgrades.

The following comes from helm best practices around CRDs:

https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you
> There is no support at this time for upgrading or deleting CRDs using Helm. This was an explicit decision after much community discussion due to the danger for unintentional data loss.

**This is why I've filed issue openebs/mayastor/issues/1586 ** 

https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts:

> Another way to do this is to put the CRD definition in one chart, and then put any resources that use that CRD in another chart.
>
> In this method, each chart must be installed separately. However, this workflow may be more useful for cluster operators who have admin access to a cluster

This PR is a convenient shortcut if you will. A completely separate chart for CRDs might make more sense.

niladrih:
@cmontemuino -- Hmm... let me try this out once. The doc seems convincing, I'll spin up a cluster later today.

Abhinandan-Purkait (approved):

niladrih:
@cmontemuino -- Changes look good to me, could you also add a switch for the volumeSnapshot CRDs?

```yaml
crds:
  # -- Install CRDs
  enabled: true
  jaeger:
    # -- Install Jaeger CRDs
    enabled: false
  volumeSnapshots:
    enabled: true
```

chart/charts/crds/values.yaml
```diff
@@ -0,0 +1,4 @@
+jaeger:
+  # -- Install Jaeger CRDs
+  enabled: false
```
niladrih: Should be `enabled: true` by default. The dependent chart should set it to false, as you have set it already.

chart/charts/crds/templates/csi-volume-snapshot-class.yaml
niladrih, line 1: Should have templating to enable/disable volumeSnapshot CRDs

chart/README.md
```diff
@@ -37,13 +37,24 @@ To install the chart with the release name `my-release`:
...
+
+This removes all the Kubernetes components associated with the chart and deletes the release.
+
+*See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall/) for command documentation.*
+
 ## Chart Dependencies
 
 | Repository | Name | Version |
 |------------|------|---------|
+|  | crds | 0.0.0 |
```
niladrih, line 57: ~Version should follow Mayastor chart version: `2.5.0`~

chart/README.md
```diff
@@ -37,13 +37,24 @@ To install the chart with the release name `my-release`:
 
 ```console
 $ helm repo add openebs https://openebs.github.io/mayastor-extensions/
-$ helm install my-release openebs/mayastor
+$ helm install mymayastor openebs/mayastor
```
niladrih:
```suggestion
$ helm install mayastor openebs/mayastor
```
... maybe change the repo name as well
```suggestion
$ helm repo add mayastor https://openebs.github.io/mayastor-extensions/
$ helm install mayastor mayastor/mayastor
```

niladrih (commented):
Left few comments.

chart/README.md
niladrih: This is not required.

cmontemuino:
Hi @niladrih, I've addressed all the comments already.
I'll fix the commit lint by squashing all commits into one. Is this ok?

chart/charts/crds/values.yaml
```diff
@@ -0,0 +1,7 @@
+jaeger:
+  # -- Install Jaeger CRDs
+  enabled: true
+  # -- Install Volume Snapshots CRD
+  volumeSnapshots:
+    enabled: true
```
niladrih:
```suggestion
jaeger:
  # -- Install Jaeger CRDs
  enabled: true
# -- Install Volume Snapshots CRD
volumeSnapshots:
  enabled: true
```

niladrih (commented):
Left one comment

niladrih:
@cmontemuino -- I see one issue. I've highlighted that above. Otherwise, changes look good to me. I think I'll approve and queue for a merge when you make the change. Thanks!

cmontemuino:
> @cmontemuino -- I see one issue. I've highlighted that above. Otherwise, changes look good to me. I think I'll approve and queue for a merge when you make the change. Thanks!

I see it now. Thanks!

cmontemuino:
> > @cmontemuino -- I see one issue. I've highlighted that above. Otherwise, changes look good to me. I think I'll approve and queue for a merge when you make the change. Thanks!
> 
> I see it now. Thanks!

@niladrih -- the PR is ready to go now

niladrih:
bors merge

niladrih (approved):

niladrih:
bors merge

cmontemuino:
@niladrih -- could you please issue `bors merge` one more time?

niladrih:
It failed because of an issue with the helm-doc script.

niladrih:
This requires https://github.com/openebs/mayastor-extensions/pull/435.

niladrih:
bors merge

niladrih:
@cmontemuino -- could you rebase your PR and run this helm-docs script against your changes:
```
nix-shell ./chart/shell.nix --run ./scripts/helm/generate-readme.sh 
```
Add the changes...
```
git status
git add chart/README.md
git commit ...
```
And then push to this branch..?

cmontemuino:
> @cmontemuino -- could you rebase your PR and run this helm-docs script against your changes:
> 
> ```
> nix-shell ./chart/shell.nix --run ./scripts/helm/generate-readme.sh 
> ```
> 
> Add the changes...
> 
> ```
> git status
> git add chart/README.md
> git commit ...
> ```
> 
> And then push to this branch..?

@niladrih -- all done

niladrih:
bors merge

tiagolobocastro:
@cmontemuino a side effect we're finding from this is that if we uninstall when there are CRs present then as the CRD is removed then the CRs are also removed... this essentially means you lose the snapshots!

I think we should add `helm.sh/resource-policy: keep` by default and provide an option to change it if required, but this should probably not be the default behaviour.
What do you think?

cmontemuino:
> @cmontemuino a side effect we're finding from this is that if we uninstall when there are CRs present then as the CRD is removed then the CRs are also removed... this essentially means you lose the snapshots!
> 
> I think we should add `helm.sh/resource-policy: keep` by default and provide an option to change it if required, but this should probably not be the default behaviour. What do you think?

I think you didn't notice the situation before because `helm uninstall` does not remove CRDs from `crds` directory.
What you suggest will work just fine. For people using Argo CD, another different annotation is required.

Would you like me to draft a PR to handle this scenario + documentation?
Basically, we need a new element, something like `crds.keep: true`.
Then, in the CRD templates, we add the annotation you've mentioned above.

Something like the following:

```yaml
# CRD code here ...
  metadata:
    {{- if .Values.crds.keep }}
    annotations:
      helm.sh/resource-policy: keep
      {{- if .Capabilities.APIVersions.Has "argoproj.io/v1alpha1" }}
      argocd.argoproj.io/sync-options: Prune=false
      {{- end }}
    {{- end }}
```

Of course it means tight-coupling to Argo CD (i.e., more maintenance, etc.), which might be non desirable....thus, something like the following might be a better fit:

```yaml
# CRD code here ...
  metadata:
    {{- if or (.Values.crds.keep) (.Values.crds.annotations }}
    annotations:
      {{- if .Values.crds.keep }}
      helm.sh/resource-policy: keep
      {{- end }}
      {{- range $key, $value := .Values.crds.annotations }}
        {{ $key }}: {{ $value | quote }}
      {{- end }}
    {{- end }}
```

With good documentation everything should be super clear.

Just let me know if you want me to do it and I'll prepare the PR this week.

tiagolobocastro:
That sounds good @cmontemuino, we'd like to take that PR if you can make it, otherwise we can look into adding it ourselves, let us know :)

Also, do we need to keep the crd presence check on the template? This seems to be required only if upgrading from v2.5.0 which packed the yaml in the crd folder. In case case we could simply document that one needs to use --set crds.enabled=false rather than keep the crd presence check?

cmontemuino:
> ```yaml
>       {{- range $key, $value := .Values.crds.annotations }}
>         {{ $key }}: {{ $value | quote }}
>       {{- end }}
> ```

Working on it already. Let me draft a PR and we discuss based on it.

cmontemuino:
> > ```yaml
> >       {{- range $key, $value := .Values.crds.annotations }}
> >         {{ $key }}: {{ $value | quote }}
> >       {{- end }}
> > ```
> 
> Working on it already. Let me draft a PR and we discuss based on it.

Can we move the conversation here? https://github.com/openebs/mayastor-extensions/pull/458

I need a bit of help to write the PR description too... not sure if this is a feature or regression for example.

PR #419, opened 2024-02-15
feat(json_fmt): helm changes to enable json support

abhilashshetty04:
base.loging.format= json|pretty|compact . 

ansi colour can be disabled also. Its enabled by default.

### DIFF
--- a/chart/templates/mayastor/agents/core/agent-core-deployment.yaml
+++ b/chart/templates/mayastor/agents/core/agent-core-deployment.yaml
@@ -59,14 +59,16 @@ spec:
             - "--events-url=nats://{{ .Release.Name }}-nats:4222"{{ end }}
             {{- if not .Values.agents.ha.enabled }}
             - "--disable-ha"{{ end }}
+            - "--fmt-style={{ include "logFormat" . }}"
+            - "--ansi-colors={{ .Values.base.logging.color }}"
           ports:
             - containerPort: 50051
           env:
             - name: RUST_LOG
               value: {{ .Values.agents.core.logLevel }}
-            {{- if default .Values.base.logSilenceLevel .Values.agents.core.logSilenceLevel }}
+            {{- if default .Values.base.logging.silenceLevel .Values.agents.core.logSilenceLevel }}
             - name: RUST_LOG_SILENCE
-              value: {{ default .Values.base.logSilenceLevel .Values.agents.core.logSilenceLevel }}
+              value: {{ default .Values.base.logging.silenceLevel .Values.agents.core.logSilenceLevel }}
             {{- end }}
             - name: MY_POD_NAME
               valueFrom:
@@ -92,14 +94,16 @@ spec:
             - "--core-grpc=https://{{ .Release.Name }}-agent-core:50051"{{ if .Values.base.jaeger.enabled }}
             - "--jaeger={{ .Values.base.jaeger.agent.name }}:{{ .Values.base.jaeger.agent.port }}"{{ end }}{{ if .Values.eventing.enabled }}
             - "--events-url=nats://{{ .Release.Name }}-nats:4222"{{ end }}
+            - "--ansi-colors={{ .Values.base.logging.color }}"
+            - "--fmt-style={{ include "logFormat" . }}"
           ports:
             - containerPort: 50052
           env:
             - name: RUST_LOG
               value: {{ .Values.agents.core.logLevel }}
-            {{- if default .Values.base.logSilenceLevel .Values.agents.core.logSilenceLevel }}
+            {{- if default .Values.base.logging.silenceLevel .Values.agents.core.logSilenceLevel }}
             - name: RUST_LOG_SILENCE
-              value: {{ default .Values.base.logSilenceLevel .Values.agents.core.logSilenceLevel }}
+              value: {{ default .Values.logging.silenceLevel .Values.agents.core.logSilenceLevel }}
             {{- end }}
             - name: MY_POD_NAME
               valueFrom:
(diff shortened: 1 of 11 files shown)


chart/values.yaml
```diff
@@ -51,6 +51,9 @@ base:
   default_req_timeout: 5s
   # -- Cache timeout for core agent & diskpool deployment
   cache_poll_period: 30s
+  logFormat:
+    jsonEnabled: false
+    ansiColorsEnabled: true
```
tiagolobocastro: how about an enum?
And maybe combine with other log stuff such as silence level

```suggestion
  logging:
    format: pretty|compact|json
    color: true
    silenceLevel:
```

abhilashshetty04: @tiagolobocastro , Initially thought about it. However, we use `FmtStyle::Compact ` only once while starting a csi node cmd line (freeze/unfreeze) service. Everywhere else its `FmtStyle::Pretty` by default. 

With my changes we keep the same behaviour. If someone wants to have json stdout they enable it or else its whatever style each service defaults to now.

tiagolobocastro: Yes but if we want to change it then it becomes a ambiguous:
jsonEnabled: true
compactEnabled: true

chart/values.yaml
```diff
@@ -51,6 +51,11 @@ base:
   default_req_timeout: 5s
   # -- Cache timeout for core agent & diskpool deployment
   cache_poll_period: 30s
+  logging:
+    # -- Valid values for format are pretty, json and compact
+    format: pretty
+    color: true
+    silenceLevel:
```
Abhinandan-Purkait, line 67: Where are we setting the silenceLevel? Did not see that on the template.

Abhinandan-Purkait (approved):

chart/values.yaml
abhilashshetty04: Not sure, Installation works post removing silenceLevel also. In develop,  `-  logSilenceLevel: `, Doesnt have any further key/val also. Seems like an incomplete config.

@niladrih , @tiagolobocastro any ideas?

tiagolobocastro: It's empty by default, meaning it's using container defaults. @abhilashshetty04 you need to modify the uses of the silenceLevel on every yaml, example: https://github.com/openebs/mayastor-extensions/blob/81ede2bbdbce5fd8ad64ad762390347529883733/chart/templates/mayastor/agents/core/agent-core-deployment.yaml#L67

abhilashshetty04: Thanks :0

chart/values.yaml
```diff
@@ -323,7 +328,7 @@ csi:
       keep_alive_tmo: ""
     # -- The kubeletDir directory for the csi-node plugin
     kubeletDir: /var/lib/kubelet
-    pluginMounthPath: /csi
+    pluginMountPath: /csi
```
tiagolobocastro: Please move this typo change to separate commit.
Also @niladrih we might need to plugin upgrade code to support this?

tiagolobocastro: I'd suggest to remove it from the PR and add both typo fix and plugin code in another PR

abhilashshetty04: Sure @tiagolobocastro , But I thought we dont need code change for this. Its just changes in values.yaml and its uses.

tiagolobocastro: Yes we don't need this change, let's remove it.

chart/templates/mayastor/agents/core/agent-core-deployment.yaml
```diff
@@ -59,14 +59,16 @@ spec:
             - "--events-url=nats://{{ .Release.Name }}-nats:4222"{{ end }}
             {{- if not .Values.agents.ha.enabled }}
             - "--disable-ha"{{ end }}
+            - "--fmt-style={{ include "logFormat" . }}"
```
tiagolobocastro, line 62: you seem to have undone the macro calling?

abhilashshetty04: @tiagolobocastro , What do you mean?

Btw this is how it will be passed. 
- "--ansi-colors=true"
- "--fmt-style=pretty"

tpl fn will invoke error if .Values.base.logging.format is not json | pretty | compact

chart/values.yaml
niladrih: This will need an upgrade change, yes.

abhilashshetty04: Reverted the changes

niladrih: I've raised PR with the upgrade changes required for both (1) the logging values changes, (2) the plugin mount path typo change.
https://github.com/openebs/mayastor-extensions/pull/443.

chart/templates/mayastor/agents/core/agent-core-deployment.yaml
tiagolobocastro: [17:15] Tiago Castro

I thought I had in the PR previously passing both json and color options using the macro

[17:16] Tiago Castro

        args:

           - "--fmt-style={{ include "logFormat" . }}"

           - "--ansi-colors={{ .Values.base.logging.color }}"

versus

        args:

           {{ include "logOptions" . }}

tiagolobocastro (approved):

chart/templates/mayastor/agents/core/agent-core-deployment.yaml
abhilashshetty04: This was the latest commit with macro call. its onlt for validating and passing fmt-style. ansi-colors is passed as usual assignment

abhilashshetty04:
bors merge

abhilashshetty04:
bors try

abhilashshetty04:
bors merge

PR #421, opened 2024-02-22
feat(eventing): add k8sClusterDomain default value to nats config

datacore-vvarakantham:
Set 'useFQDN' to 'false' in Nats config in values.yaml. Now NATS uses short names to refer to other pods within the same namespace - No domain suffix.

## Motivation and Context

To make it clear of the default value being used for K8sClusterDomain in NATS config.

## Types of changes

## Checklist:

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -710,6 +710,8 @@ nats:
   cluster:
     enabled: true
     replicas: 3
+  # Define if NATS is using FQDN name for clustering (i.e. nats-0.nats.default.svc.cluster.local) or short name (i.e. nats-0.nats.default).
+  useFQDN: false
   statefulSetPodLabels:
     app: nats
     openebs.io/logging: "true"


chart/values.yaml
```diff
@@ -675,6 +675,8 @@ nats:
   cluster:
     enabled: true
     replicas: 3
+  # Cluster Domain configured on the kubelets
+  k8sClusterDomain: cluster.local
```
tiagolobocastro: wondering if there's other places which would this need fixed, and if so can we have a global variable for this? Although not sure if that's possible in helm IIRC?

niladrih: This is not guaranteed to be `cluster.local` for all clusters.

datacore-vvarakantham: Yes it is not guaranteed to be cluster.local for all clusters. But can we add it for the user to know where to change the domain in Nats config?

datacore-vvarakantham: @tiagolobocastro Where can we add the global variable?

tiagolobocastro: I'm not sure, we could add this to like the base config, but then we can't use it here right? Because we can't reference variables here IIRC?

niladrih: Yes, it's been difficult to set values to the values.yaml using variables. May be not possible at the moment.

niladrih: I'm not sure how to go about doing this. I'm going to say I'm good with `custer.local` at the moment.

niladrih (approved):

chart/values.yaml
tiagolobocastro: Alright, I'm good with that.
I also want to check, what is the intent in adding this in the yaml, just as documentation? If so the comment should start with # --

chart/values.yaml
```diff
@@ -675,6 +675,8 @@ nats:
   cluster:
     enabled: true
     replicas: 3
+  # Cluster Domain configured on the kubelets
```
```suggestion
  # -- Kubernetes cluster domain name
```

chart/values.yaml
```diff
@@ -675,6 +675,8 @@ nats:
   cluster:
```
datacore-vvarakantham, line 710: hostNetwork: true
dnsPolicy: ClusterFirstWithHostNet

Instead can we add this config to Nats to get the domain by default? When we add this config here, it is added to the Nats pods spec by default. We are doing similarly for io-engine daemon set spec as well.

chart/values.yaml
datacore-vvarakantham: I removed k8sClusterDomain and set 'useFQDN' to 'false' now.

tiagolobocastro (approved):

tiagolobocastro:
@niladrih do we need upgrade job changes for this?
bors merge

niladrih:
@tiagolobocastro New values are file. No changes needed for upgrade.

PR #422, opened 2024-02-26
feat(metrics_exporter): remove polling interval

abhilashshetty04:


chart/values.yaml
```diff
@@ -85,9 +85,6 @@ base:
   metrics:
     # -- Enable the metrics exporter
     enabled: true
-    # metrics refresh time
-    # WARNING: Lowering pollingInterval value will affect performance adversely
-    pollingInterval: "5m"
```
niladrih, line 90: I'm not very clear on how this the exporter works today, but is this configuration now made available somewhere else on the values.yaml?

abhilashshetty04: No we dont need this at all. We have removed --polling-interval from metrics exporter container args as well.

tiagolobocastro (approved):

niladrih (approved):

abhilashshetty04:
bors merge

PR #424, opened 2024-02-29
Volume expansion changes

niladrih:
Changes:
- Adds `allowVolumeExpansion` to the default StorageClass. Value is set to true by default.
- Adds the resizer to the CSI controller Deployment. Leader-election is disabled, log verbosity is set to `v=2`, timeout and retry durations are left unchanged. Uses v1.9.3.
- Adds upgrade changes to prefer target helm chart's resizer version, instead of source helm chart's.


tiagolobocastro (approved):

dsharma-dc (approved):

niladrih:
bors merge

PR #432, opened 2024-03-08
feat(charts): add values to control etcd and loki hostpath sc creation

Abhinandan-Purkait:
## Description
Add `etcd.localpvScConfig.enabled` and `loki-stack.localpvScConfig.enabled` values so that we can control the creation of storage classes for etcd and loki, for cases when localpv is installed from the umbrella helm chart.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -422,6 +422,7 @@ io_engine:
 etcd:
   # Configuration for etcd's localpv hostpath storage class.
   localpvScConfig:
+    enabled: true
     # Name of etcd's localpv hostpath storage class.
     name: "mayastor-etcd-localpv"
     # -- Host path where local etcd data is stored in.
@@ -522,6 +523,7 @@ loki-stack:
   enabled: true
   # Configuration for loki's localpv hostpath storage class.
   localpvScConfig:
+    enabled: true
     # Name of loki's localpv hostpath storage class.
     name: "mayastor-loki-localpv"
     # -- Host path where local etcd data is stored in.
--- a/chart/templates/etcd/storage/localpv-storageclass.yaml
+++ b/chart/templates/etcd/storage/localpv-storageclass.yaml
@@ -1,4 +1,4 @@
-{{ if and (index .Values "localpv-provisioner" "enabled") .Values.etcd.persistence.enabled }}
+{{ if and .Values.etcd.localpvScConfig.enabled .Values.etcd.persistence.enabled }}
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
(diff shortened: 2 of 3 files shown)


chart/values.yaml
```diff
@@ -756,3 +756,7 @@ localpv-provisioner:
     enabled: false
   hostpathClass:
     enabled: false
+  # Create a hostpath storage class for mayastor-etcd.
+  createEtcdSc: true
```
tiagolobocastro: Do we need separate ones for etcd and loki?

tiagolobocastro (approved):

chart/values.yaml
```diff
@@ -756,3 +756,7 @@ localpv-provisioner:
...
+  # Create a hostpath storage class for Loki Stack.
+  createLokiSc: true
```
niladrih: These values are used by the mayastor chart and do not need to be passed down to the localpv-provisioner helm dependency.

niladrih (changes requested):

chart/values.yaml
niladrih: Could we maybe use something along the lines of `etcd.localpvScConfig.enabled: true` and `loki-stack.localpvScConfig.enabled: true`?

Abhinandan-Purkait: It's related to localpv-provisoner version of the mayastor chart so it probably makes sense to keep it here?

chart/values.yaml
Abhinandan-Purkait: Yes we do for case when someone wants to disable loki and might not want this storageclass.

chart/values.yaml
niladrih: It's not related to the localpv chart at all, I'd guess. StorageClasses are used by loki and etcd, localpv doesn't need the resources this enables/disables.

Abhinandan-Purkait: Okay. I will move it there. @tiagolobocastro @avishnu Does this sound good?

tiagolobocastro: Sure, sounds good to me, please update PR description as appropriate

chart/values.yaml
```diff
@@ -406,6 +406,7 @@ io_engine:
 etcd:
   # Configuration for etcd's localpv hostpath storage class.
   localpvScConfig:
+    enabled: true
```
niladrih, line 425: This needs `# -- ` style comment to generate documentation.

chart/values.yaml
```diff
@@ -506,6 +507,7 @@ loki-stack:
   enabled: true
   # Configuration for loki's localpv hostpath storage class.
   localpvScConfig:
+    enabled: true
```
niladrih, line 526: Same as above. Needs doc comment.

chart/values.yaml
Abhinandan-Purkait: Not sure if we want this to be a part of the doc.

niladrih (approved):
LGTM

chart/values.yaml
niladrih: Okay..

Abhinandan-Purkait:
bors merge

PR #434, opened 2024-03-13
feat(callhome): add volume replica and state, pool, spdk managed disks and storage media metrics

datacore-vvarakantham:
Added volume replica and status metrics to callhome report

```
{
  "volumes":
      {
           "volume_replica_counts": {
              "one_replica": 1,
              "two_replicas": 4,
              "three_replicas": 0,
              "four_replicas": 0,
              "five_or_more_replicas": 0
           },
           "volume_state_counts": {
                "unknown": 0,
                "online": 5,
                "degraded": 0,
                "faulted": 0,
                "shutdown": 0
            }
     }
}

```
Added pool metrics per node to callhome report

```
"storage_nodes": {
    "nodes": {
      "node-1-153835": {
        "pools": [
          {
            "replicas": 0,
            "replicas_capacity": 0
          }
        ],
        "mayastor_managed_disks_count": 1
      },
      "node-0-153835": {
        "pools": [
          {
            "replicas": 0,
            "replicas_capacity": 0
          }
        ],
        "mayastor_managed_disks_count": 1
      }
    }
  },
```

Added mayastor managed disks to callhome report
```
"mayastor_managed_disks": {
    "disk_types": {
      "SSD": {
        "count": 2,
        "capacity": 21474836480
      }
    }
    
```

Added storage media metrics to callhome report

```
"storage_media": {
    "count": 9,
    "total_capacity_in_bytes": 526959770112,
    "disk_types": {
      "SSD": {
        "count": 6,
        "capacity": 523738546176
      },
      "Optical Drive": {
        "count": 3,
        "capacity": 3221223936
      }
    }
  },
```

## Types of changes

## Checklist:

### DIFF
--- a/call-home/src/bin/callhome/collector/report_models.rs
+++ b/call-home/src/bin/callhome/collector/report_models.rs
@@ -1,11 +1,17 @@
-use obs::common::{constants::ACTION, errors};
-use openapi::models::{Volume, VolumeStatus};
+use obs::common::{
+    constants::{ACTION, BYTES_PER_SECTOR},
+    errors,
+};
+use openapi::models::{BlockDevice, Volume, VolumeStatus};
 use prometheus_parse::{Sample, Value};
 use reqwest_middleware::{ClientBuilder, ClientWithMiddleware};
 use reqwest_retry::{policies::ExponentialBackoff, RetryTransientMiddleware};
 use serde::{Deserialize, Serialize};
 use snafu::ResultExt;
-use std::convert::TryFrom;
+use std::{
+    collections::{HashMap, HashSet},
+    convert::TryFrom,
+};
 use url::Url;
 
 /// Volumes contains volume count, min, max, mean and capacity percentiles.
@@ -134,20 +140,22 @@ pub(crate) struct Pools {
     created: u32,
     #[serde(skip_serializing_if = "is_zero")]
     deleted: u32,
+    total_capacity_in_bytes: u64,
 }
 impl Pools {
     /// Receives a vector of openapi::models::Pool and returns a new report_models::Pools object by
     /// using the data provided.
     pub(crate) fn new(pools: Vec<openapi::models::Pool>, event_data: EventData) -> Self {
-        let pools_size_vector = get_pools_size_vector(pools);
+        let pools_size_vector = get_pools_size_vector(&pools);
         Self {
             count: pools_size_vector.len() as u64,
             max_size_in_bytes: get_max_value(pools_size_vector.clone()),
             min_size_in_bytes: get_min_value(pools_size_vector.clone()),
             mean_size_in_bytes: get_mean_value(pools_size_vector.clone()),
-            capacity_percentiles_in_bytes: Percentiles::new(pools_size_vector),
+            capacity_percentiles_in_bytes: Percentiles::new(pools_size_vector.clone()),
             created: event_data.pool_created.value(),
             deleted: event_data.pool_deleted.value(),
+            total_capacity_in_bytes: pools_size_vector.iter().sum(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -116,6 +200,166 @@ impl Nexus {
...
 }
 
+/// StorageMedia contains storage media devices count, total capacity of all the storage media
+/// devices and disk_types which contains capacity and count of storage media for each disk_type.
+#[derive(Serialize, Deserialize, Debug, Default)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct StorageMedia {
+    count: u64,
+    total_capacity_in_bytes: u64,
+    disk_types: HashMap<String, DiskTypeInfo>,
```
tiagolobocastro, line 216: what is the key here?

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -116,6 +200,166 @@ impl Nexus {
...
+
+        // obtain the details of each disk using the Block Device API.
+        for device in disks.clone() {
+            let disk_type = get_disk_type(device.clone());
+
+            // add the capacity of the disk to the total capacity for the disk type
+            disk_types
+                .entry(disk_type.clone())
+                .or_insert(DiskTypeInfo::default())
+                .capacity_in_bytes += device.size * 512;
```
tiagolobocastro: is size always in 512B blocks, even if the device is 4k?

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -116,6 +200,166 @@ impl Nexus {
...
+
+            // increase the total count for the disk type
+            disk_types
```
tiagolobocastro: hmm why not get the device above and then set both capacity and count?

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -116,6 +200,166 @@ impl Nexus {
...
+                .entry(disk_type)
+                .or_insert(DiskTypeInfo::default())
+                .count += 1;
+        }
+
+        let total_capacity: u64 = disks.iter().map(|disk| disk.size).sum();
+        Self {
+            count: disks.len() as u64,
+            total_capacity_in_bytes: total_capacity * 512,
```
tiagolobocastro: even if size is always in 512 blocks, let's add a constant here rather than use 512

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -116,6 +200,166 @@ impl Nexus {
...
+                    .or_insert(NodeInfo::default())
+                    .spdk_managed_disks_count += spec.disks.len() as u64;
+
+                let pool_info = PoolInfo::new(pool.id, replicas.clone());
+                storage_nodes
+                    .nodes
+                    .entry(spec.node)
+                    .or_insert(NodeInfo::default())
+                    .pools
+                    .push(pool_info)
```
tiagolobocastro: same thing here

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -116,6 +200,166 @@ impl Nexus {
...
+    }
+}
+
+/// NodeInfo contains PoolInfo (replicas capacity and replicas count) for each pool in a node and
+/// spdk managed disks count for the node.
+#[derive(Serialize, Deserialize, Debug, Default)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct NodeInfo {
+    pools: Vec<PoolInfo>,
+    spdk_managed_disks_count: u64,
```
tiagolobocastro: what does this mean?

call-home/src/bin/callhome/main.rs
```diff
@@ -208,12 +220,34 @@ async fn generate_report(
...
+                StorageNodes::new(replicas.clone().into_body(), pools.into_body())
+        }
+    }
+
+    // find valid disks to calculate storage media metrics
+    let valid_disks = disks
+        .into_iter()
+        .filter(|device| {
+            device.size > 0
+                && device.devtype != "partition"
```
tiagolobocastro, line 254: what happens if the mayastor pool is created on a partition?

call-home/src/bin/callhome/collector/report_models.rs
datacore-vvarakantham: The key here is disk type (ex: "SSD") that we get from get_disk_type() function. The value contains "DiskTypeInfo" from all nodes for the particular disk type.

call-home/src/bin/callhome/collector/report_models.rs
datacore-vvarakantham: Yes. Linux always considers sectors to be 512 bytes long independently of the devices real block size. According to this source.
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/types.h?id=v4.4-rc6#n121

call-home/src/bin/callhome/collector/report_models.rs
datacore-vvarakantham: I am finding the number of physical disks used by the pools in a node to get the value "spdk_managed_disks_count" for the node.

call-home/src/bin/callhome/main.rs
datacore-vvarakantham: I am excluding all partitions to focus on physical disk devices when calculating storage media metrics. 

Also, I am including physical disks utilized by pools to determine SPDK-managed disks. In cases where a pool is created with a partition, I am considering the parent physical disk to determine the count of SPDK-managed disks. When calculating the capacity of these SPDK-managed disks, I am taking into account the size of the partitions, if the disk is partitioned, rather than considering the full capacity of the physical disk.

call-home/src/bin/callhome/collector/report_models.rs
datacore-vvarakantham: Fixed

call-home/src/bin/callhome/collector/report_models.rs
datacore-vvarakantham: Using a constant now.

call-home/src/bin/callhome/collector/report_models.rs
datacore-vvarakantham: Done

call-home/src/common/constants.rs
```diff
@@ -67,6 +67,9 @@ pub const DEFAULT_MBUS_URL: &str = "nats://mayastor-nats:4222";
 /// Defines the default namespace.
 pub const DEFAULT_NAMESPACE: &str = "mayastor";
 
+// Number of bytes in a disk sector
```
tiagolobocastro:
```suggestion
/// Number of bytes in a disk sector.
```

call-home/src/bin/callhome/main.rs
```diff
@@ -177,18 +177,34 @@ async fn generate_report(
         }
     };
 
+    // List of disks on each node.
+    let mut node_disks = HashMap::new();
     let nodes = http_client.nodes_api().get_nodes(None).await;
     match nodes {
-        Ok(nodes) => report.storage_node_count = nodes.into_body().len() as u8,
+        Ok(nodes) => {
+            for node in nodes.clone().into_body() {
```
tiagolobocastro:
```suggestion
            let nodes = nodes.into_body();
            for node in &nodes {
```

call-home/src/bin/callhome/main.rs
```diff
@@ -177,18 +177,34 @@ async fn generate_report(
...
+                    .get_node_block_devices(&node.id, Some(true))
+                    .await
+                    .map(|b_devs_result| {
+                        node_disks
+                            .entry(node.id)
+                            .or_insert(Vec::new())
+                            .extend(b_devs_result.into_body());
+                    });
+            }
+            report.storage_node_count = nodes.into_body().len() as u8
```
tiagolobocastro:
```suggestion
            report.storage_node_count = nodes.len() as u8
```

call-home/src/bin/callhome/main.rs
```diff
@@ -177,18 +177,34 @@ async fn generate_report(
...
     };
 
+    // List of disks on each node.
+    let mut node_disks = HashMap::new();
     let nodes = http_client.nodes_api().get_nodes(None).await;
     match nodes {
-        Ok(nodes) => report.storage_node_count = nodes.into_body().len() as u8,
+        Ok(nodes) => {
+            for node in nodes.clone().into_body() {
+                let _ = http_client
```
tiagolobocastro: if let Ok() = ....

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -116,6 +206,232 @@ impl Nexus {
...
+#[serde(rename_all = "camelCase")]
+pub(crate) struct PoolInfo {
+    replicas: u64,
+    replicas_capacity_in_bytes: u64,
+}
+
+impl PoolInfo {
+    /// Receives pool_id and a vector of openapi::models::Replica and returns a new
+    /// report_models::PoolInfo object by using the data provided.
+    pub(crate) fn new(pool_id: String, replicas: Vec<openapi::models::Replica>) -> Self {
```
tiagolobocastro: Pass by reference instead, and avoid cloning

```suggestion
    pub(crate) fn new(pool_id: String, replicas: &Vec<openapi::models::Replica>) -> Self {
```

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -116,6 +206,232 @@ impl Nexus {
...
+        pools: Vec<openapi::models::Pool>,
+        node_disks: HashMap<String, Vec<BlockDevice>>,
+    ) -> Self {
+        let mut spdk_disks = Self::default();
+
+        // HashMap to keep track of counted node disks.
+        let mut counted_node_disks = HashMap::new();
+
+        // Iterate over each pool specification.
+        for pool_spec in pools.iter().filter_map(|pool| pool.spec.clone()) {
```
tiagolobocastro:
```suggestion
        for pool_spec in pools.into_iter().filter_map(|pool| pool.spec) {
```

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -230,6 +549,42 @@ fn get_pools_size_vector(pools: Vec<openapi::models::Pool>) -> Vec<u64> {
     pools_size_vector
 }
 
+// Gets disk type from BlockDevice data.
+fn get_disk_type(device: BlockDevice) -> String {
```
tiagolobocastro: do we need to consider loop devices here?

call-home/src/bin/callhome/collector/report_models.rs
orville-wright: @tiagolobocastro  - This Feature was request by PM for deeper understanding of users storage infra

call-home/src/bin/callhome/main.rs
orville-wright: @tiagolobocastro  - This Feature was request by PM for deeper understanding of users storage infra
What we need to understand is more info about the disk media allocated to the cluster itself. How Mayastor utilizes that disk media is important (full disk or GPT partitioned);  but this doesn't help us understand how a users storage capacity is evolving within the cluster. It actually confuses our understanding of it. Looking at the entire disk give us an exact understanding.

call-home/src/bin/callhome/collector/report_models.rs
orville-wright: @tiagolobocastro  @datacore-vvarakantham  - Yes , LOOP device support would be a good addition. I forgot about that storage class. Great catch.

orville-wright (approved):
Reviewed use-case and feature completeness.
LGTM

call-home/src/common/constants.rs
datacore-vvarakantham: Fixed

call-home/src/bin/callhome/main.rs
datacore-vvarakantham: Fixed

call-home/src/bin/callhome/main.rs
datacore-vvarakantham: Fixed

call-home/src/bin/callhome/collector/report_models.rs
datacore-vvarakantham: Fixed

call-home/src/bin/callhome/collector/report_models.rs
datacore-vvarakantham: Fixed

call-home/src/bin/callhome/collector/report_models.rs
datacore-vvarakantham: Added "Loop Device" type. I added "Network Block Device" and "RAM Disk" as well.

call-home/src/bin/callhome/main.rs
datacore-vvarakantham: Fixed

call-home/src/bin/callhome/collector/report_models.rs
tiagolobocastro: What about files, ex: `aio:///var/local/file`
and pcie access, example: `pcie:///0000:01:00.0`

datacore-vvarakantham: Is this a physical disk and not partition? If it is a physical disk can we keep it in the unknown disk type? Or should I get the disk type to be file?

tiagolobocastro: the first is just a file, so maybe we don't need to include that?
The second is pci access to nvme device, so can be classified as `NVMe SSD` I think.

datacore-vvarakantham: ("pci", Some(true)) => "PCIe HDD", // Explicitly handle PCIe HDDs
("pci", Some(false)) => "PCIe SSD", // General PCIe SSD

Should add these checks for PCIe devices?

tiagolobocastro: There's no such thing as pcie hdd.
It should be "NVMe SSD" as per your existing nomenclature as it's basically and nvme device, but detached from the linux kernel.

tiagolobocastro: You also won't get any information from mayastor block devices api for this device

datacore-vvarakantham: I added scheme check on the disk uri to find the pcie devices.

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -116,6 +206,251 @@ impl Nexus {
...
+    }
+}
+
+/// NodeInfo contains PoolInfo (replicas capacity and replicas count) for each pool in a node and
+/// spdk managed disks count for the node.
+#[derive(Serialize, Deserialize, Debug, Default)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct NodeInfo {
+    pools: Vec<PoolInfo>,
+    spdk_managed_disks_count: u64,
```
tiagolobocastro: Suggest we leave spdk out, it's mayastor

```suggestion
    mayastor_managed_disks_count: u64,
```

tiagolobocastro (approved):

call-home/src/bin/callhome/collector/report_models.rs
datacore-vvarakantham: Fixed

chriswldenyer (approved):
Looks reasonable.

datacore-vvarakantham:
bors merge

PR #437, opened 2024-03-14
fix(kubectl/dump/etcd): use correct namespace

tiagolobocastro:
Etcd dump failing since an etcd bugfix, which had etcd return all data rather than keys matching a prefix.

## Description

Namespace for etcd is not available in k8s env variables as the plugin runs outside the cluster.
A better solution is perhaps to simply ensure we create clients with the correct namespace as default?

## Types of changes

## Checklist:


niladrih (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #438, opened 2024-03-15
Chart changes

niladrih:
- add README.md helm-docs template for crds subchart
- add values to toggle install of all CRDs
- coreutils to nix-shells
- add value option to toggle installation of snapshot CRDs
- refactor generate-readme.sh script to work with CRDs subchart
- add templates to crds chart to avoid installation if CRDs already exists, because otherwise install/upgrade fails because these template-style CRDs carry a release-name label.
- remove CRD-install functionality during upgrade, as the above templating takes care of that at install/upgrade time.

### DIFF
--- a/chart/charts/crds/templates/_helpers.tpl
+++ b/chart/charts/crds/templates/_helpers.tpl
@@ -1,0 +1,19 @@
+{{/* vim: set filetype=mustache: */}}
+
+{{/*
+    This returns a "1" if the CRD is absent in the cluster
+    Usage:
+      {{- if (include "crdIsAbsent" (list <crd-name>)) -}}
+      # CRD Yaml
+      {{- end -}}
+*/}}
+{{- define "crdIsAbsent" -}}
+    {{- $crdName := index . 0 -}}
+    {{- $crd := lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" $crdName -}}
+    {{- $output := "1" -}}
+    {{- if $crd -}}
+        {{- $output = "" -}}
+    {{- end -}}
+
+    {{- $output -}}
+{{- end -}}
--- a/shell.nix
+++ b/shell.nix
@@ -23,6 +23,7 @@ mkShell {
     cargo-udeps
     clang
     commitlint
+    coreutils
     cowsay
     git
     helm-docs
--- a/chart/README.md
+++ b/chart/README.md
@@ -64,7 +64,7 @@ This removes all the Kubernetes components associated with the chart and deletes
 ## Values
 
 | Key | Description | Default |
-|-----|-------------|:-------:|
+|:----|:------------|:--------|
 | agents.&ZeroWidthSpace;core.&ZeroWidthSpace;capacity.&ZeroWidthSpace;thin.&ZeroWidthSpace;poolCommitment | The allowed pool commitment limit when dealing with thin provisioned volumes. Example: If the commitment is 250 and the pool is 10GiB we can overcommit the pool up to 25GiB (create 2 10GiB and 1 5GiB volume) but no further. | `"250%"` |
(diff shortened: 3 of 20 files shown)


chart/charts/crds/templates/_helpers.tpl
```diff
@@ -0,0 +1,19 @@
...
+{{/*
+    This returns a "1" if the CRD is absent in the cluster
+    Usage:
+      {{- if (include "crdIsAbsent" (list <crd-name>)) -}}
+      # CRD Yaml
+      {{- end -}}
+*/}}
+{{- define "crdIsAbsent" -}}
+    {{- $crdName := index . 0 -}}
+    {{- $crd := lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" $crdName -}}
```
tiagolobocastro, line 12: what happens when there's a v2?

shell.nix
```diff
@@ -23,6 +23,7 @@ mkShell {
     cargo-udeps
     clang
     commitlint
+    coreutils
```
tiagolobocastro, line 26: what is this used for?

chart/Chart.yaml
```diff
@@ -27,7 +27,6 @@ appVersion: "0.0.0"
 dependencies:
   - name: crds
     version: "0.0.0"
-    condition: crds.enabled
```
tiagolobocastro: this might still be useful if we want to disable the chart itself?

chart/Chart.yaml
```diff
@@ -27,7 +27,6 @@ appVersion: "0.0.0"
 dependencies:
   - name: crds
     version: "0.0.0"
```
tiagolobocastro: btw do we need to change the publish script to also set the version for the crd chart?

chart/charts/crds/templates/_helpers.tpl
niladrih: I haven't thought of that.
Maybe when the apiVersion changes, we check for the v2 apiVersion on the function instead of v1.
Typically Kubernetes will have v1 and v2 both on some of their release versions before v1 is removed and v2 is standard. Just like how it happened with apiextensions.k8s.io/v1beta1.

Also this doesn't cover different API versions for the same CRD.
These are complex behaviours, and depends on if there are breaking changes between the APIs.

shell.nix
niladrih: `realpath`. For the alpine-type folks.

chart/Chart.yaml
niladrih: Hmm, okay, it's fine to leave it in I guess.

chart/Chart.yaml
niladrih: This change is already in the PR.

tiagolobocastro: Yeah, this file is the first so hadn't gotten there :D

chart/charts/crds/templates/_helpers.tpl
tiagolobocastro: What if we need to keep a mix of v1 and v2? Example let's say csi with v2 and jaeger with v1?

niladrih: Two tpl functions, maybe? crdIsAbsentv1 crdIsAbsentv2

tiagolobocastro: Yeah guess so. We can't do that dynamically by parsing the CR yaml?

niladrih: I've been poking around, and I couldn't find a way to do that @tiagolobocastro.

niladrih: Do you think if the template function also took the apiVersion as input, it would be better?

tiagolobocastro: If you could find a way around it and parse yaml then I don't think we need to barcode the API Version on the template

tiagolobocastro (approved):

chart/charts/crds/templates/_helpers.tpl
niladrih: I'm sorry, I meant to write 'I couldn't find a way', I made a typo.

abhilashshetty04 (approved):

niladrih:
bors merge

niladrih:
bors cancel

niladrih:
bors merge

PR #441, opened 2024-03-15
Data plane upgrade binary

niladrih:
Changes:
- Refactors upgrade-job, moves most of the code from the bin directory to the outer upgrade library.
- Adds the restart-io-engines binary.

The restart-io-engines should go through all io-engine Pods restarting them, and then circling back and restarting them again, all while adhering to the 'upgrade' way of things.


tiagolobocastro (changes requested):
This seems like a huge change to be adding near release, I'd like to understand this a bit better before.

niladrih:
> This seems like a huge change to be adding near release, I'd like to understand this a bit better before.


That makes sense. I think I too want this to sit for now.

sinhaashish:
IMHO using a flag to skip the validation will be musch simpler change to do , rather than refactoring

dsavitskiy (approved):

niladrih:
This is no longer needed. Closing this PR.

tiagolobocastro:
> This is no longer needed. Closing this PR.

I think it's still useful to have this sort of thing @niladrih.
Example: someone manages their cluster with Argo. After an upgrade the pods still need restarting. 

We'd need some sort of upgrade-less restart or even have the kubectl-mayastor plugin itself do it.

niladrih:
Alright, I'd have set this up again.. hmm.

niladrih:
I've filed this issue: https://github.com/openebs/mayastor/issues/1674

PR #454, opened 2024-03-22
fix(upgrade-job): look for crds in charts dir

niladrih:
The upgrade-job validates the chart that it's got by checking if all the right directories are where they're supposed to be. Recent changes to the CRD installation method (ref: #414) has resulted in the upgrade-job failing, because things aren't how it expects. This PR changes the CRDs directory path to that of the subchart which encapsulates the CRDs now.


Abhinandan-Purkait (approved):

tiagolobocastro (approved):

niladrih:
bors merge

PR #463, opened 2024-04-01
feat(upgrade-job): use 'crds' subchart toggles to disable existing CRDs

niladrih:
Changes:
- adds a CRDs API client
- uses the crds subchart's enable toggles to disable CRDs which already exist in the cluster.

### DIFF
--- a/k8s/upgrade/src/bin/upgrade-job/common/error.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/common/error.rs
@@ -384,10 +384,6 @@ pub(crate) enum Error {
     ))]
     NotAKnownHelmChart { chart_name: String },
 
-    /// Error for when namespace option is not set when building KubeClientSet.
-    #[snafu(display("Mandatory KubeClientSetBuilder option 'namespace' not set"))]
-    KubeClientSetBuilderNs,
-
     /// Error for when mandatory options for an EventRecorder are missing when building.
     #[snafu(display("Mandatory options for EventRecorder were not given"))]
     EventRecorderOptionsAbsent,
@@ -695,6 +691,9 @@ pub(crate) enum Error {
         args: Vec<String>,
         std_err: String,
     },
+
+    #[snafu(display("failed to list CustomResourceDefinitions: {source}"))]
+    ListCrds { source: kube::Error },
 }
 
 /// A wrapper type to remove repeated Result<T, Error> returns.
--- a/k8s/upgrade/src/bin/upgrade-job/common/kube_client.rs
+++ b/k8s/upgrade/src/bin/upgrade-job/common/kube_client.rs
@@ -1,82 +1,40 @@
-use crate::common::error::{K8sClientGeneration, KubeClientSetBuilderNs, Result};
-use k8s_openapi::api::{
-    apps::v1::Deployment,
-    core::v1::{Namespace, Node, Pod},
+use crate::common::error::{K8sClientGeneration, Result};
+use k8s_openapi::{
+    api::{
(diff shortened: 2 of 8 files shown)


k8s/upgrade/src/bin/upgrade-job/common/kube_client.rs
```diff
@@ -48,9 +59,28 @@ pub(crate) struct KubeClientSet {
     pods_api: Api<Pod>,
     namespaces_api: Api<Namespace>,
     deployments_api: Api<Deployment>,
+    crds_api: Api<CustomResourceDefinition>,
 }
 
 impl KubeClientSet {
+    pub(crate) async fn get() -> Self {
+        KUBE_CLIENT_SET_BUILDER
```
tiagolobocastro: Why do this?

niladrih: Because, the kube-client set has clients for namespaced kubernetes resources.

Previously we have been creating a client whenever one is required. Because of the namespaced resources, functions which need a client but not necessarily the namespaced resources, have to accept the namespace as an argument.

With a global client builder, the client set builder is always available, and it always knows about the namespace. And this frees up these functions from having to know about the namespace.

tiagolobocastro: From what I can tell the namespace is available on most places where it's used?
Most times I've used these singleton patterns I later came to regret it. I suggest you rename get to more thing that makes it clearer to the caller that the namespace is pre-configured from something else. Or have the namespace itself as a singleton.

niladrih: Makes sense to me. I've tossed the whole thing out for a simpler client module.

k8s/upgrade/src/bin/upgrade-job/common/error.rs
```diff
@@ -695,6 +691,9 @@ pub(crate) enum Error {
         args: Vec<String>,
         std_err: String,
     },
+
+    #[snafu(display("failed to list CustomResourceDefinitions: {}", source))]
```
tiagolobocastro:
```suggestion
    #[snafu(display("failed to list CustomResourceDefinitions: {source}"))]
```

k8s/upgrade/src/bin/upgrade-job/common/kube_client.rs
```diff
@@ -1,82 +1,44 @@
...
-impl KubeClientSetBuilder {
-    /// Build Kubernetes API clients for a specific namespace (for namespaced object only).
-    #[must_use]
-    pub(crate) fn with_namespace<T>(mut self, namespace: T) -> Self
-    where
-        T: ToString,
-    {
-        self.namespace = Some(namespace.to_string());
-        self
+impl KubeClient {
```
tiagolobocastro: should this just be a module rather than a struct?

tiagolobocastro (approved):

k8s/upgrade/src/bin/upgrade-job/common/kube_client.rs
niladrih: Yeah, I guess so.

sinhaashish (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

niladrih:
bors cancel

niladrih:
bors merge

tiagolobocastro:
bors merge

PR #483, opened 2024-04-12
Add failure for when partial rebuild is enabled for source charts >=2.2,<=2.5

niladrih:


k8s/upgrade/src/bin/upgrade-job/upgrade.rs
```diff
@@ -132,3 +164,30 @@ async fn upgrade_product(opts: &CliArgs, event: &mut EventRecorder) -> Result<()
...
+        .filter_map(|pod| pod.labels().get(CHART_VERSION_LABEL_KEY))
+        .any(|v| {
+            let version =
+                Version::parse(v).expect("failed to parse version from io-engine Pod label");
+            version.ge(&PARTIAL_REBUILD_DISABLE_EXTENTS.0)
+                & version.le(&PARTIAL_REBUILD_DISABLE_EXTENTS.1)
+        });
+
+    if partial_rebuild_disable_required && partial_rebuild_is_enabled {
+        error!("Partial rebuild must be disabled for upgrades from {CORE_CHART_NAME} chart versions >= {}, <= {}", PARTIAL_REBUILD_DISABLE_EXTENTS.0, PARTIAL_REBUILD_DISABLE_EXTENTS.1);
```
niladrih, line 181: @tiagolobocastro @Abhinandan-Purkait @tiagolobocastro -- does this error message seem ok to you?

Abhinandan-Purkait: LGTM. Maybe can we specify the exact value as well if that adds to a better UX?

Abhinandan-Purkait (approved):

k8s/upgrade/src/bin/upgrade-job/upgrade.rs
niladrih: Two different charts to cover, do I do it for both?

tiagolobocastro (approved):

k8s/upgrade/src/bin/upgrade-job/upgrade.rs
tiagolobocastro: We can just do the mayastor specific version, then the user can always lookup the docs since they will have to anyway to figure out how to disable partial rebuild?

niladrih: I'm not adding the helm command here, as this plugin doesn't support the modification via itself. The helm command for this change should be picked up from the docs IMO.

niladrih:
bors merge

PR #494, opened 2024-04-19
fix(upgrade-job): deserialize default when localpv chart is disabled

niladrih:
Local PV Hostpath will be installed as a chart alongside Mayastor, and not as a dependency of Mayastor in `openebs/openebs` v4.0.1. The deserialize fails when the chart is installed with helm v3.13+ because disabled dependency values are absent in the values set for those helm versions.


Abhinandan-Purkait (approved):

avishnu (approved):

niladrih:
bors merge

PR #498, opened 2024-05-09
CallHome: add pagination to k8s & api-rest requests

niladrih:


call-home/src/bin/callhome/collector/k8s_client.rs
```diff
@@ -17,15 +22,33 @@ impl K8sClient {
...
+        loop {
+            let node_list = nodes.list_metadata(&list_params).await?;
+
+            let continue_ = node_list.metadata.continue_.clone();
+
+            nodes_count += node_list.items.len();
+
+            match continue_ {
+                Some(token) => {
+                    list_params = list_params.continue_token(token.as_str());
```
tiagolobocastro: Seems pointless to clone only to use reference later on

```suggestion
            nodes_count += node_list.items.len();       

            match node_list.metadata.continue_.as_str() {
                Some(token) => {
                    list_params = list_params.continue_token(token);
```

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -25,8 +25,8 @@ pub(crate) struct Volumes {
 impl Volumes {
     /// Receives a openapi::models::Volumes object and returns a new report_models::volume object by
     /// using the data provided.
-    pub(crate) fn new(volumes: openapi::models::Volumes, event_data: EventData) -> Self {
-        let volumes_size_vector = get_volumes_size_vector(volumes.entries);
+    pub(crate) fn new(volumes: Vec<Volume>, event_data: EventData) -> Self {
```
tiagolobocastro, line 40: We should test this for memory use to see if we have enough on the memory resources limit but I think the limit seems enough

call-home/src/bin/callhome/collector/storage_rest.rs
```diff
@@ -0,0 +1,29 @@
...
+
+    // The last paginated request will set the `starting_token` to `None`.
+    while starting_token.is_some() {
+        let vols = client
+            .volumes_api()
+            .get_volumes(STORAGE_API_PAGE_SIZE, None, starting_token)
+            .await
+            .map(|response| response.into_body())?;
+
+        volumes = volumes.into_iter().chain(vols.entries).collect();
```
tiagolobocastro:
```suggestion
        volumes.extend(vols.entries);
```

call-home/src/bin/callhome/main.rs
```diff
@@ -193,14 +194,10 @@ async fn generate_report(
...
-    let volumes = match volumes {
-        Ok(volumes) => Some(volumes.into_body()),
-        Err(err) => {
-            error!("{:?}", err);
-            None
-        }
-    };
+    let volumes = list_all_volumes(&http_client)
+        .await
+        .map_err(|error| error!("{error:?}"))
```
tiagolobocastro: This bit was also a bit confusing from the logs, maybe include what the error was, ex:
```suggestion
        .map_err(|error| error!("Failed to list all volumes: {error:?}"))
```

tiagolobocastro (commented):
I wonder if getting all replicas can also fail?
What do we do with this data btw? How much data is sent to the callhome endpoint, does that need to be split up too?

niladrih:
> I wonder if getting all replicas can also fail? What do we do with this data btw? How much data is sent to the callhome endpoint, does that need to be split up too?

This is what we serialize and send: https://github.com/openebs/mayastor-extensions/blob/develop/call-home/src/bin/callhome/collector/report_models.rs#L151

Here are links to the types that this embeds: 
- https://github.com/openebs/mayastor-extensions/blob/develop/call-home/src/bin/callhome/collector/report_models.rs#L45
- https://github.com/openebs/mayastor-extensions/blob/develop/call-home/src/bin/callhome/collector/report_models.rs#L14
- https://github.com/openebs/mayastor-extensions/blob/develop/call-home/src/bin/callhome/collector/report_models.rs#L76
- https://github.com/openebs/mayastor-extensions/blob/develop/call-home/src/bin/callhome/collector/report_models.rs#L97
- https://github.com/openebs/mayastor-extensions/blob/develop/call-home/src/bin/callhome/collector/report_models.rs#L122
- https://github.com/openebs/mayastor-extensions/blob/develop/call-home/src/bin/callhome/collector/report_models.rs#L128

Overall, it's like 30-40 u64s, and <10 Strings... like 10 UUIDs. I'd say it's small, nothing huge.

call-home/src/bin/callhome/collector/report_models.rs
niladrih: It's 16 MiB to 32 MiB request and limit at the moment.

I'll test! This is my first time using valgrind in a 'real world' setting outside of cs107! I'll mock 10,000 Volumes across 1000 Nodes.

Abhinandan-Purkait (approved):

call-home/src/bin/callhome/collector/report_models.rs
niladrih: I've tested this and there's about <41 MiB of heap usage.

I've tested with 10,000 volumes, 0 nodes, 0 pools (scale of the vols seemed to cover most typical nodes and pool counts as far as memory usage is concerned).

I'll bump the limit to 64 MiB.

niladrih: For some iterators, the elements are copied. Fixing those allocations with iterator level manipulations should eliminate a lot of this memory usage.

tiagolobocastro: That should be plenty then, though if we reduce those copies then maybe we can keep it at 32

tiagolobocastro (approved):

call-home/src/bin/callhome/collector/report_models.rs
niladrih: The request is set to 16 MiB. So it always starts out at 16 MiB.

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -64,33 +66,35 @@ impl VolumeReplicaCounts {
...
     fn new(volumes: &[Volume]) -> Self {
+        let (one_replica, two_replicas, three_replicas, four_replicas, five_or_more_replicas): (
+            u32,
+            u32,
+            u32,
+            u32,
+            u32,
+        ) = volumes.iter().fold(
+            (0, 0, 0, 0, 0),
+            |(one, two, three, four, five_or_more), vol| match vol.spec.num_replicas {
```
tiagolobocastro: how does this reduce memory usage?
Though it's probably more probably more efficient time wise

call-home/src/bin/callhome/collector/report_models.rs
```diff
@@ -33,25 +34,26 @@ pub(crate) struct Volumes {
...
         Self {
-            count: volumes_size_vector.len() as u64,
-            max_size_in_bytes: get_max_value(volumes_size_vector.clone()),
-            min_size_in_bytes: get_min_value(volumes_size_vector.clone()),
-            mean_size_in_bytes: get_mean_value(volumes_size_vector.clone()),
-            capacity_percentiles_in_bytes: Percentiles::new(volumes_size_vector),
+            count: volume_sizes.len() as u64,
+            max_size_in_bytes: get_max_value(&volume_sizes),
+            min_size_in_bytes: get_min_value(&volume_sizes),
+            mean_size_in_bytes: get_mean_value(volume_sizes.clone()),
```
tiagolobocastro: can't we pass the ref here? or perhaps even just the volumes?

call-home/src/common/iterator.rs
```diff
@@ -0,0 +1,69 @@
...
+    NonAscending,
+}
+
+#[derive(Debug, Clone)]
+pub struct SortedExactSizeIter<T> {
+    data: Vec<T>,
+    sort_order: SortOrder,
+}
+
+impl<T: Ord> SortedExactSizeIter<T> {
```
tiagolobocastro: Why do we need this?

call-home/src/common/iterator.rs
```diff
@@ -0,0 +1,69 @@
...
+    // Create a new SortedExactSizeIter from an iterator and a sort order.
+    pub fn from_iter<I>(iter: I, sort_order: SortOrder) -> Self
+    where
+        I: Iterator<Item = T>,
+    {
+        let mut data: Vec<T> = iter.collect();
+        match sort_order {
+            SortOrder::NonDescending => data.sort_unstable(),
```
tiagolobocastro: unstable?

call-home/src/common/iterator.rs
```diff
@@ -0,0 +1,69 @@
+#[derive(Debug, Clone)]
+pub enum SortOrder {
+    NonDescending,
+    NonAscending,
+}
+
+#[derive(Debug, Clone)]
+pub struct SortedExactSizeIter<T> {
+    data: Vec<T>,
```
tiagolobocastro: This is taking ownership so again more copies?

call-home/src/bin/callhome/collector/report_models.rs
niladrih: Yeah, doesn't reduce any memory usage. Makes things a little faster.

call-home/src/common/iterator.rs
niladrih: This type guarantees that it's sorted. That takes makes max, min and reduces the serveral sorts in `get_percentile`, while making sure its input is sorted with this type.

call-home/src/common/iterator.rs
niladrih: It's faster: https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable
Stable is not necessary for us. It would make sense to use stable sort, if the move operations were very costly. But they are constant time and cheap. https://doc.rust-lang.org/std/primitive.slice.html#method.sort

Stable here talks about if it will reorder elements or not. Not an indicator of the stability of the code.

>When applicable, unstable sorting is preferred because it is generally faster than stable sorting and it doesn’t allocate auxiliary memory. See [sort_unstable](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable).

call-home/src/common/iterator.rs
niladrih: Yes.. I'll have to get around this one. For this application, sorting the volumes and pools should be fine, instead of collecting their sizes into separate iterators. That'd save some space I think.

call-home/src/bin/callhome/collector/report_models.rs
niladrih: The sum method consumes the iterator. But that is something I can work around. I'll avoid the clone here.

tiagolobocastro: Can't see that code here now, but might be able to just move it down to the last init of the struct

call-home/src/common/iterator.rs
tiagolobocastro: Why can't we just sort it inline? It's trivial and would avoid adding this whole file.

call-home/src/bin/callhome/collector/report_models.rs
niladrih: Yeah, I've gotten rid of this clone.

niladrih:
@tiagolobocastro -- I've made a note of the review comments.

The context of this PR was pagination. I've removed the memory efficiency changes at the moment. I'll take these changes in a separate PR.

TODO: Link new PR here.

niladrih:
bors merge

PR #499, opened 2024-05-10
Make image prefix configurable

niladrih:


default.nix
```diff
@@ -1,9 +1,9 @@
-{ system ? null, allInOne ? true, incremental ? false, static ? false, img_tag ? "", tag ? "", img_org ? "" }:
+{ system ? null, allInOne ? true, incremental ? false, static ? false, img_tag ? "", tag ? "", img_org ? "", product_prefix ? "" }:
```
tiagolobocastro, line 1:
```suggestion
{ system ? null, allInOne ? true, incremental ? false, static ? false, img_tag ? "", tag ? "", img_org ? "", product_prefix ? "mayastor" }:
```

nix/overlay.nix
```diff
@@ -1,7 +1,10 @@
-{ allInOne ? true, incremental ? false, static ? false, img_tag ? "", tag ? "", img_org ? "" }:
+{ allInOne ? true, incremental ? false, static ? false, img_tag ? "", tag ? "", img_org ? "", product_prefix ? "" }:
```
tiagolobocastro, line 1:
```suggestion
{ allInOne ? true, incremental ? false, static ? false, img_tag ? "", tag ? "", img_org ? "", product_prefix ? "mayastor" }:
```

nix/pkgs/images/default.nix
```diff
@@ -2,7 +2,7 @@
 # avoid dependency on docker tool chain. Though the maturity of OCI
 # builder in nixpkgs is questionable which is why we postpone this step.
 
-{ pkgs, dockerTools, lib, extensions, busybox, gnupg, kubernetes-helm-wrapped, semver-tool, yq-go, runCommand, sourcer, img_tag ? "", img_org ? "" }:
+{ pkgs, dockerTools, lib, extensions, busybox, gnupg, kubernetes-helm-wrapped, semver-tool, yq-go, runCommand, sourcer, img_tag ? "", img_org ? "", img_prefix ? "" }:
```
tiagolobocastro:
```suggestion
{ pkgs, dockerTools, lib, extensions, busybox, gnupg, kubernetes-helm-wrapped, semver-tool, yq-go, runCommand, sourcer, img_tag ? "", img_org ? "", img_prefix }:
```

tiagolobocastro (commented):
Or perhaps even read defaults from some config file?

nix/pkgs/images/default.nix
niladrih: I'm using a config fix to store default value for product_prefix.

default.nix
```diff
@@ -1,9 +1,18 @@
...
```
tiagolobocastro, line 1: maybe null? just incase we ever want to build without a prefix and below we compare against null rather than ""

```suggestion
{ system ? null, allInOne ? true, incremental ? false, static ? false, img_tag ? "", tag ? "", img_org ? "", product_prefix ? null }:
```

nix/config.nix
```diff
@@ -0,0 +1,8 @@
+/*
+  This file contains default values for parameters used in nix derivations.
+  Set a value here if you want to share a nix value across all of the repos
+  which use this repo as a git submodule.
```
tiagolobocastro: Add back once we move this to common submodule
```suggestion
```

scripts/release.sh
```diff
@@ -285,7 +285,7 @@ if [ -n "$BUILD_BINARIES" ]; then
 fi
 
 for name in $IMAGES; do
-  image_basename=$($NIX_EVAL -f . images.$BUILD_TYPE.$name.imageName | xargs)
+  image_basename=$($NIX_EVAL -f . images.$BUILD_TYPE.$name.imageName --argstr product_prefix "$PRODUCT_PREFIX" | xargs)
```
tiagolobocastro, line 289: This seems unbound variable access?
Also if env var not set maybe we should not append --argstr at all

niladrih: We use the `set -u` option on the script. It's not going to be simple to add behaviour to a 'env variable not set' clause.

default.nix
niladrih: Sure, I think that makes sense.

nix/config.nix
niladrih: Oh, I forgot about the comment. Removed the last 2 lines.

avishnu:
[sad]   Vishnu Govind Attur reacted to your message:
________________________________

scripts/release.sh
tiagolobocastro: Did you test this? I suspect this script will fail as it stands due to the -u...

You can do something like this:
```
if ! [ -z ${PRODUCT_PREFIX+x} ]; then
  PRODUCT_PREFIX=${PRODUCT_PREFIX:-"\"\""}
  PRODUCT_PREFIX="--argstr product_prefix $PRODUCT_PREFIX"
else
  PRODUCT_PREFIX=""
fi
```

niladrih: Yeah the typical -z $PRODUCT_PREFIX check doesn't work. I did test it.

niladrih: Yes I did test the nix-eval too.

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

PR #505, opened 2024-05-16
chore: mod path of image pull secret

chriswldenyer:
Move the image pull secret to be under image, not base.

## Description

Move the image pull secret to be under image, not base.
This is the more logical location.

## Regression
No

## Types of changes

## Checklist:

### DIFF
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -77,8 +77,20 @@ Usage:
 {{ include "base_pull_secrets" . }}
 */}}
 {{- define "base_pull_secrets" -}}
-    {{- if .Values.base.imagePullSecrets.enabled }}
-    {{- include "render" (dict "value" .Values.base.imagePullSecrets.secrets "context" $) | nindent 8 }}
+    {{- if (not (empty .Values.image.pullSecrets)) }}
+        {{- range .Values.image.pullSecrets | uniq -}}
+            {{ nindent 8 "- name:" }} {{ . }}
+        {{- end }}
+    {{- else -}}
+        {{- if .Values.base.imagePullSecrets }}
+            {{- if .Values.base.imagePullSecrets.enabled }}
+                {{- if (empty .Values.base.imagePullSecrets.secrets) }}
+                    {{ nindent 8 "- name: login" }}
+                {{- else -}}
+                    {{- include "render" (dict "value" .Values.base.imagePullSecrets.secrets "context" $) | nindent 8 }}
+                {{- end}}
+            {{- end }}
+        {{- end }}
     {{- end }}
 {{- end -}}
 
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -26,6 +26,9 @@ image:
   # -- ImagePullPolicy for our images
   pullPolicy: Always
 
+  # -- docker-secrets required to pull images if the container registry from image.registry is protected
+  pullSecrets: []
+
 # -- Node labels for pod assignment
 # ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
 # Note that if multi-arch images support 'kubernetes.io/arch: amd64'
@@ -87,13 +90,6 @@ base:
       - name: etcd-probe
         image: busybox:latest
         command: ['sh', '-c', 'trap "exit 1" TERM; until nc -vzw 5 {{ .Release.Name }}-etcd {{.Values.etcd.service.port}}; do date; echo "Waiting for etcd..."; sleep 1; done;']
-  # docker-secrets required to pull images if the container registry from image.Registry is protected
-  imagePullSecrets:
-    # -- Enable imagePullSecrets for pulling our container images
-    enabled: false
-    # Name of the imagePullSecret in the installed namespace
-    secrets:
-      - name: login
 
   metrics:
     # -- Enable the metrics exporter
--- a/chart/README.md
+++ b/chart/README.md
@@ -16,7 +16,7 @@ Mayastor Helm chart for Kubernetes
    ```
  - Create secret if downloading the container images from a private repo.
    ```sh
-   kubectl create secret docker-registry <same-as-base.imagePullSecrets.secrets>  --docker-server="https://index.docker.io/v1/" --docker-username="<user-name>" --docker-password="<password>" --docker-email="<user-email>" -n <mayastor-namespace>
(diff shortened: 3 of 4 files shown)


chart/templates/_helpers.tpl
```diff
@@ -77,8 +77,16 @@ Usage:
...
 {{- define "base_pull_secrets" -}}
-    {{- if .Values.base.imagePullSecrets.enabled }}
-    {{- include "render" (dict "value" .Values.base.imagePullSecrets.secrets "context" $) | nindent 8 }}
+    {{- if .Values.base.imagePullSecrets }}
+        {{- if .Values.base.imagePullSecrets.enabled }}
+        {{- include "render" (dict "value" .Values.base.imagePullSecrets.secrets "context" $) | nindent 8 }}
+        {{- end }}
+    {{- else -}}
+        {{- if .Values.image.pullSecrets }}
+            {{- if .Values.image.pullSecrets.enabled }}
```
tiagolobocastro: I think this would work?
```suggestion
        {{- if (.Values.image.pullSecrets).enabled }}
```

tiagolobocastro (approved):
Please run helm upgrade manually (previous install with secrets enabled) to ensure it's working

chart/values.yaml
```diff
@@ -26,6 +26,14 @@ image:
   # -- ImagePullPolicy for our images
   pullPolicy: Always
 
+  # docker-secrets required to pull images if the container registry from image.Registry is protected
+  pullSecrets:
+    # -- Enable pullSecrets for pulling our container images
+    enabled: false
+    # -- Name of the pullSecret in the installed namespace
+    secrets:
+      - name: login
```
dsharma-dc: This (secrets)name is not going to change in this repo I'd believe?

tiagolobocastro: Looking at bitnami, they simply pass an array of secrets, presumably meaning its existence implies the enabled to true.
So I suggest we simply do the same:

```suggestion
  pullSecrets: []
```

chriswldenyer:
bors try

chart/templates/_helpers.tpl
chriswldenyer: Helm doesn't seem to like that shortcut:
helm template mayastor .
Error: template: mayastor/templates/mayastor/agents/ha/ha-node-daemonset.yaml:33:12: executing "mayastor/templates/mayastor/agents/ha/ha-node-daemonset.yaml" at <include "base_pull_secrets" .>: error calling include: template: mayastor/templates/_helpers.tpl:80:18: executing "base_pull_secrets" at <.Values.base.imagePullSecrets.enabled>: nil pointer evaluating interface {}.enabled

chart/values.yaml
chriswldenyer: fixed

chart/values.yaml
```diff
@@ -26,6 +26,9 @@ image:
   # -- ImagePullPolicy for our images
   pullPolicy: Always
 
+  # docker-secrets required to pull images if the container registry from image.Registry is protected
```
```suggestion
  # -- docker-secrets required to pull images if the container registry from image.registry is protected
```

tiagolobocastro (approved):

chart/values.yaml
chriswldenyer: Actually there is no default name any more in the latest change.

chart/values.yaml
chriswldenyer: fixed

chriswldenyer:
bors try

Abhinandan-Purkait (approved):

chart/templates/_helpers.tpl
```diff
@@ -77,8 +77,12 @@ Usage:
...
 */}}
 {{- define "base_pull_secrets" -}}
-    {{- if .Values.base.imagePullSecrets.enabled }}
-    {{- include "render" (dict "value" .Values.base.imagePullSecrets.secrets "context" $) | nindent 8 }}
+    {{- if .Values.base.imagePullSecrets }}
+        {{- if .Values.base.imagePullSecrets.enabled }}
+        {{- include "render" (dict "value" .Values.base.imagePullSecrets.secrets "context" $) | nindent 8 }}
+        {{- end }}
+    {{- else -}}
+    {{- include "render" (dict "value" .Values.image.pullSecrets "context" $) | nindent 8 }}
```
tiagolobocastro: As discussed, this needs changing

chriswldenyer: Updated, as suggested.

niladrih (approved):
Configuration lives through helm upgrade, no upgrade changes required.

chriswldenyer:
Re-requested reviews because of a change in the logic of the pull-secrets helper function.

tiagolobocastro (approved):

dsharma-dc (approved):

chriswldenyer:
One further change - use default secret name of "login" if the legacy method is used to enable the secret and the name is not specified.

chart/templates/_helpers.tpl
```diff
@@ -77,8 +77,20 @@ Usage:
...
-    {{- include "render" (dict "value" .Values.base.imagePullSecrets.secrets "context" $) | nindent 8 }}
+    {{- if (not (empty .Values.image.pullSecrets)) }}
+        {{- range .Values.image.pullSecrets | uniq -}}
+            {{ nindent 8 "- name:" }} {{ . }}
+        {{- end }}
+    {{- else -}}
+        {{- if .Values.base.imagePullSecrets }}
+            {{- if .Values.base.imagePullSecrets.enabled }}
+                {{- if (empty .Values.base.imagePullSecrets.secrets) }}
+                    {{ nindent 8 "- name: login" }}
```
niladrih, line 88: Is this really necessary? If neither the new nor the old configuration keys are being used, should we assume configuration for a user? This feels like it is a good candidate for a failure case.

niladrih (commented):
Left a comment.

chart/templates/_helpers.tpl
chriswldenyer: I was thinking of the case where the user is specifying --set base.imagePullSecrets=true but relying of the default name of "login", and having defined a secret called login in the cluster. In which case the name would no longer be defined in the updated values.yaml file.
I couldn't say how likely this scenario would be for users, but is possible AFAICT.

niladrih: Okay, yeah that case makes sense to me. Good idea to leave this in I guess.

niladrih (approved):

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

chriswldenyer:
bors merge

PR #514, opened 2024-06-12
fix(call-home): use exclusive method to calculate percentile

niladrih:
Changes:
1. The `get_percentile` function uses inclusive method of calculating percentiles. This is not the preferred way of calculating percentiles for discrete data sets. This PR changes the implementation to use the exclusive method. This method is more reliable for calculating percentiles between 10 and 90 percentile.
2. `get_percentile` returns the generated percentile as an integer without rounding it to the nearest interger. It uses a type cast for the f64, and simply drops the franctional part.
3. `get_percentile` returns a 0 for empty data sets. Percentiles for empty data sets are undefined.

While this PR introduces these changes. It doesn't break the existing APIs and doesn't change data types. This needs to be taken up in a different PR with adequate changes in the backed server.

### DIFF
--- a/call-home/src/math.rs
+++ b/call-home/src/math.rs
@@ -1,0 +1,69 @@
+/// Get percentile value for a sorted array of u64s sorted in non-descending order, using the
+/// exclusive method (but also accounts for 0th and 100th percentile). The exclusive method is the
+/// preferred way of calculating percentile for discrete data points. Sizes of volumes, pools and
+/// replica counts are discrete. This is more reliable for percentile values between 10 and 90.
+pub fn percentile_exclusive(sorted_data: &[u64], percentile: f64) -> Option<f64> {
+    match sorted_data.len() {
+        // Percentile for an empty data set is not defined.
+        0 => None,
+        // Only one data point. This is the only possible percentile value.
+        1 => Some(sorted_data[0] as f64),
+        len => {
+            // Position of the required percentile value.
+            let pos = (len as f64 + 1.0) * percentile / 100.0;
+
+            // Positions before the first data point don't actually exist.
+            if pos <= 1.0 {
+                return Some(sorted_data[0] as f64);
+            // Positions beyond the length of the data set don't actually exist.
+            } else if pos >= len as f64 {
+                return Some(sorted_data[len - 1] as f64);
+            }
+
+            /* Interpolate the value at fractional (decimal) position, because our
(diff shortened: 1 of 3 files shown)


call-home/src/math.rs
```diff
@@ -0,0 +1,70 @@
...
+pub fn percentile_exclusive(sorted_data: &[u64], percentile: f64) -> Option<f64> {
+    let n = sorted_data.len();
+    match n {
+        // Percentile for an empty data set is not defined.
+        0 => None,
+        // Only one data point. This is the only possible percentile value.
+        1 => Some(sorted_data[0] as f64),
+        _ => {
+            // Position of the required percentile value.
+            let pos = (n as f64 + 1.0) * percentile / 100.0;
```
tiagolobocastro:
```suggestion
    match sorted_data.len() {
        // Percentile for an empty data set is not defined.
        0 => None,
        // Only one data point. This is the only possible percentile value.
        1 => Some(sorted_data[0] as f64),
        len => {
            // Position of the required percentile value.
            let pos = (len as f64 + 1.0) * percentile / 100.0;
```

call-home/src/math.rs
```diff
@@ -0,0 +1,70 @@
...
+
+            // Positions before the first data point doesn't actually exist.
+            if pos <= 1.0 {
+                return Some(sorted_data[0] as f64);
+            // Positions beyond the length of the data set doesn't actually exist.
```
tiagolobocastro:
```suggestion
            // Positions beyond the length of the data set don't actually exist.
```

call-home/src/math.rs
```diff
@@ -0,0 +1,70 @@
...
+    match n {
+        // Percentile for an empty data set is not defined.
+        0 => None,
+        // Only one data point. This is the only possible percentile value.
+        1 => Some(sorted_data[0] as f64),
+        _ => {
+            // Position of the required percentile value.
+            let pos = (n as f64 + 1.0) * percentile / 100.0;
+
+            // Positions before the first data point doesn't actually exist.
```
tiagolobocastro:
```suggestion
            // Positions before the first data point don't actually exist.
```

tiagolobocastro (approved):
It's more readable than the previous code but I'm not really a math expert to do any meaningful review here :)

datacore-vvarakantham (approved):
LGTM

niladrih:
bors merge

PR #516, opened 2024-06-13
[ CI ] - Add kind cluster setup for helm testing

tiagolobocastro:
## Description

Adds k8s cluster deployment capabilities using kind clusters.
Adds simple initial test, running on GitHub actions directly using kind and kind on a kvm.

## Motivation and Context

Currently k8s and helm testing only ever happens after changes are merged, by system test (e2e).
This is non optimal because it allows bugs to creep in first, so they can be picked by e2e and then fixed.
We can skip this cycle altogether, for most cases, by simply doing more sanity tests in this repo, using kind! 

## Types of changes

## Checklist:

- [x] I have added unit tests to cover my changes.


niladrih:
🙌

scripts/k8s/deployer.sh
```diff
@@ -0,0 +1,192 @@
...
+- role: control-plane
+EOF
+
+start_core=1
+nodes=()
+for node_index in $(seq 1 $WORKERS); do
+  if [ "$node_index" == 1 ]; then
+    node="kind-worker"
+  else
+    node="kind-worker$node_index"
```
dsharma-dc, line 143: nitpick - `node="kind-worker-$node_index"`

scripts/helm/install.sh
```diff
@@ -0,0 +1,85 @@
...
+fi
+
+if [ -n "$WAIT" ]; then
+  WAIT_ARG=" --wait --timeout $TIMEOUT"
+fi
+
+echo "Installing Mayastor Chart"
+
+set -x
+helm install mayastor "$CHART_DIR" -n mayastor --create-namespace \
```
dsharma-dc, line 79: Should we better be installing using openebs chart and make replicated ON?

tests/bdd/minikube/README.md
dsharma-dc, line 1: Can we retain minikube scripts as well just as an option?

dsharma-dc (approved):

scripts/k8s/deployer.sh
tiagolobocastro: This is just generating the same names kind uses, and sadly we have no control over that :(
https://github.com/kubernetes-sigs/kind/issues/1401

scripts/helm/install.sh
tiagolobocastro: hmm good question
somehow feels wrong here to install openers chart as we have no control over it, ex: how do we tell it to use this chart, which specific version do we use, or what if it fails for bugs from the umbrella, it shouldn't affect this repo's ability to function.
However, we have to indeed do this kind of testing for the umbrella, somehow, somewhere...

tests/bdd/minikube/README.md
tiagolobocastro: I was thinking about it initially, but since we have kind and it works for multi-node don't know if there's any value in keeping two.
If required to reuse for something it's in the git history so it's not lost.

dsharma-dc: I would prefer to retain as 

1. They are only few files, no overhead.
2. If required at all, people will need to figure out restoration etc. Rather we can mention in some Readme that we have both options.

tiagolobocastro: Could you elaborate on why you'd want to keep it? Kind does everything, it's much simpler and it just works.
The existing minikube setup does not work for me, and it tried to do things like modify my OS which thankfully did not cause lasting damage. 
Also I'm not sure if it's also in a full working condition, ie can it create pools, volumes, run fio etc?
But besides that, unless there's something Minikube is giving us that kind can not, I'm not sure what the advantage of keeping two, other than confuse potential devs.

dsharma-dc: I haven't used `kind` yet so can't compare atm. No direct bearing to this repo though, but the idea was that until we change all localPV CIs also to use kind we may keep it just in case someone need to run cluster locally(Vanilla minikube setup works ok also though).
Your call...

scripts/helm/install.sh
Abhinandan-Purkait: I don't think we should use umbrella chart here, as umbrella chart is usually updated at last only after the releases from dependent, this might cause gaps?

Abhinandan-Purkait (approved):

scripts/helm/install.sh
tiagolobocastro: Yes agreed, but we still need to test the umbrella somehow...
Once we add more proper tests here, perhaps there's another run on the umbrella "core" which will call the very same tests to run, but against the umbrella chart.

tests/bdd/minikube/README.md
tiagolobocastro: In that case I suggest that we either switch to kind on those or we move these scripts there as it doesn't make sense to keep them here if they're to be used by local engine testing.
Also makes me wonder, we also have terraform k8s in the mayastor control-plane... perhaps we need an infra repo where these sort of things are kept?

tiagolobocastro:
bors merge

PR #517, opened 2024-06-14
Cherry-pick changes from mayastor PR

tiagolobocastro:
Cherry-pick user changes from stale mayastor PR.
Modified the initContainers change as we now have many init containers.

### DIFF
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -20,7 +20,7 @@ Usage:
 */}}
 {{- define "base_init_core_containers" -}}
     {{- if .Values.base.initCoreContainers.enabled }}
-    {{- include "render" (dict "value" .Values.base.initCoreContainers.containers "context" $) | nindent 8 }}
+    {{- include "render_init_containers" (dict "value" .Values.base.initCoreContainers.containers "context" $) | nindent 8 }}
     {{- end }}
 {{- end -}}
 
@@ -31,7 +31,7 @@ Usage:
 */}}
 {{- define "base_init_ha_node_containers" -}}
     {{- if .Values.base.initHaNodeContainers.enabled }}
-    {{- include "render" (dict "value" .Values.base.initHaNodeContainers.containers "context" $) | nindent 8 }}
+    {{- include "render_init_containers" (dict "value" .Values.base.initHaNodeContainers.containers "context" $) | nindent 8 }}
     {{- end }}
 {{- end -}}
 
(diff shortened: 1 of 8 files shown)


niladrih (commented):
The if condition here should also depend on `etcd.enabled`
https://github.com/openebs/mayastor-extensions/blob/1609f7a779426abc63b5015ad5fb9086003788e6/chart/templates/etcd/storage/localpv-storageclass.yaml

chart/templates/_helpers.tpl
```diff
@@ -287,4 +287,39 @@ Get the Jaeger URL
...
+{{- define "render_init_containers" -}}
+    {{- $containers := list }}
+    {{- $image := .Values.base.initContainers.image }}
+    {{- range .value -}}
+        {{ $container := . }}
+        {{- if not (hasKey . "image") }}
+            {{- with $image }}
+                {{- $_ := set $container "image" . }}
+            {{- else }}
+                {{- $_ := set $container "image" "docker.io/bitnami/bitnami-shell:10" }}
```
niladrih: Should the default image be configurable via values.yaml? This could pose a problem for air-gapped folks.

tiagolobocastro: hmm perhaps you misunderstood the code? the default is in the values.yaml
If not present, ex: upgrades with --reuse then we hardcode it here.

niladrih: Oh, that's fair.

chart/values.yaml
```diff
@@ -578,8 +576,8 @@ loki-stack:
       runAsUser: 1001
     # initContainers to chown the static hostpath PV by 1001 user
     initContainers:
-      - command: ["/bin/bash", "-ec", "chown -R 1001:1001 /data"]
-        image: docker.io/bitnami/bitnami-shell:10
+      - command: ["/bin/sh", "-ec", "chown -R 1001:1001 /data"]
+        image: docker.io/busybox:1.36.1-musl
```
niladrih: That's very appropriate :O

niladrih (approved):

Abhinandan-Purkait (approved):

chart/values.yaml
tiagolobocastro: though I wonder if we should put our own on openebs's dockerhub?
Sadly for etcd, their init container has the init hardcoded to use bash IIRC, so we'd still need one with bash.

niladrih: Sure, I can see the utility in that. For someone behind an air-gap, they might be taking stock of all the containers they'd be needing.
This one already lives on the openebs org: https://github.com/openebs/linux-utils/blob/main/Dockerfile
We could add bash to the apk list. Also it's alpine, so it'd have musl. But even if it didn't, everything we build is statically built and our shell scripts don't really depend on the libc implemention differences of musl/glibc, etc.

niladrih (approved):

niladrih (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
> The if condition here should also depend on `etcd.enabled` https://github.com/openebs/mayastor-extensions/blob/1609f7a779426abc63b5015ad5fb9086003788e6/chart/templates/etcd/storage/localpv-storageclass.yaml

Done

niladrih (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #525, opened 2024-07-08
feat(scripts/helm/install): make installation idempotent

niladrih:
The helm/install.sh script should succeed if the release already exists.
Adds a flag to choose failure if that's a failure scenario for the user.

This may be used when adding tests after the nixOsTest for helm install

### DIFF
--- a/scripts/helm/install.sh
+++ b/scripts/helm/install.sh
@@ -65,8 +65,9 @@ while [ "$#" -gt 0 ]; do
   esac
 done
 
+DEP_UPDATE_ARG=
 if [ -n "$DEP_UPDATE" ]; then
-  helm dependency update "$CHART_DIR"
+  DEP_UPDATE_ARG="--dependency-update"
 fi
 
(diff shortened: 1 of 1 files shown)


tiagolobocastro (approved):
I'm fine with this change but this may be a rabbit hole.
Consider some destructive tests which restart pods, scale down deployment etc - this change may not be sufficient
Also consider that the same may happen to the nodes themselves.

Abhinandan-Purkait (approved):

niladrih:
> I'm fine with this change but this may be a rabbit hole. Consider some destructive tests which restart pods, scale down deployment etc - this change may not be sufficient Also consider that the same may happen to the nodes themselves.

That is true. I decided on this change when considering next-steps after the nixOsTest `helm install`. Maybe the immediate next test could use the same helm release and carry on. This is not a good idea for quite a number of cases, I agree.

niladrih:
bors merge

PR #529, opened 2024-08-02
feat(eventing): add error logs and retries for rest api requests

niladrih:


call-home/src/bin/callhome/main.rs
```diff
@@ -256,5 +294,84 @@ async fn generate_report(
     report.storage_media = StorageMedia::new(valid_disks);
 
     report.nexus = Nexus::new(event_data);
+
+    let logs = logs.lock().unwrap();
+    // Taking a slice of the last hundred logs messages, to not make the payload too big.
```
tiagolobocastro: Shouldn't this be handled in the log side? Otherwise this will keep growing forever?

niladrih: I've added the length check while preparing the logs buffer. I'm doing it using a new function `rev_length_check`. I've removed the slicing when pushing to the Report.

call-home/src/bin/callhome/main.rs
```diff
@@ -256,5 +296,91 @@ async fn generate_report(
...
+    timestamp: String,
+    level: String,
+    file: String,
+    line: u32,
+    message: String,
+}
+
+/// Ensures the length of a vector is no more than the input limit, and removes elements from
+/// the front if it goes over the limit.
+fn rev_length_check<T>(logs: &mut Vec<T>, limit: usize) {
```
tiagolobocastro: Why not just something like this?

```rust
fn log_with_limit<T>(logs: &mut VecDeque<T>, message: T, limit: usize) {
    if logs.len() > limit-1 {
        logs.pop_front();
    }
    logs.push_back(message);
}
```

tiagolobocastro (commented):
Aren't we still missing a clear of the logs before every run? Otherwise we accumulate errors from previous days?

call-home/src/bin/callhome/main.rs
niladrih: Sure, using VecDeque across the board. Using this to limit size instead:
```rust
fn log_with_limit<T>(logs: &mut VecDeque<T>, message: T, limit: usize) {
    while logs.len().ge(&limit) {
        logs.pop_front();
    }
    logs.push_back(message);
}
```

niladrih:
>Aren't we still missing a clear of the logs before every run? Otherwise we accumulate errors from previous days?

@tiagolobocastro -- Yeah, it's an Arc. I forgot about that. I've added an explicit clear after the VecDeque contents are dumped on to the report.

tiagolobocastro (approved):

call-home/src/bin/callhome/main.rs
```diff
@@ -256,5 +296,92 @@ async fn generate_report(
...
+
+        // Only capture logs of level ERROR
+        if level == &Level::ERROR {
+            let mut logs = self.logs.lock().unwrap();
+            log_with_limit(&mut logs, log_entry, ERR_LOG_BUF_CAPACITY);
+        }
+    }
+}
+
+//
```
Abhinandan-Purkait: Remove this? Or maybe add the comment

Abhinandan-Purkait (approved):

call-home/src/bin/callhome/main.rs
niladrih: Oh! oops!

niladrih: Removed the comment.

niladrih:
bors merge

PR #532, opened 2024-08-23
feat(call-home): make product name configurable

niladrih:


call-home/src/common/constants.rs
```diff
@@ -6,7 +6,10 @@ use utils::version_info;
 
 /// PRODUCT is the name of the project for which this call-home component is deployed.
 pub fn product() -> String {
-    ::constants::product_train()
+    env::var("PRODUCT_NAME")
+        .ok()
+        .filter(|v| !v.is_empty())
+        .unwrap_or(::constants::product_train())
```
dsharma-dc: Is this working well for both - when env var is not set and also when it is set to blank "" ?

dsharma-dc (approved):

call-home/src/common/constants.rs
niladrih: Yeah, it is. `env::var()` returns a `Result::Err()` for unset env, but an `Ok()` for empty env. That's why there's a `filter(|v| !v.is_empty())` to cut that case out as a `None()`.

I'll add a unit test, as I had the same concern.

niladrih: Added a unit test.

call-home/src/common/constants.rs
```diff
@@ -6,7 +6,10 @@ use utils::version_info;
 
 /// PRODUCT is the name of the project for which this call-home component is deployed.
 pub fn product() -> String {
-    ::constants::product_train()
+    env::var("PRODUCT_NAME")
```
tiagolobocastro: What is the reason for this change?

call-home/src/common/constants.rs
```diff
@@ -4,9 +4,15 @@ use std::{
...
 
+/// CALLHOME_PRODUCT_NAME_ENV is the name of the ENV which configures the product-name for callhome.
+pub const CALLHOME_PRODUCT_NAME_ENV: &str = "CALLHOME_PRODUCT_NAME";
+
 /// PRODUCT is the name of the project for which this call-home component is deployed.
 pub fn product() -> String {
-    ::constants::product_train()
+    env::var(CALLHOME_PRODUCT_NAME_ENV)
+        .ok()
+        .filter(|v| !v.is_empty())
```
tiagolobocastro, line 15: Need to map to train as well

niladrih: Okay..

tiagolobocastro (approved):

niladrih:
bors merge

PR #543, opened 2024-09-16
Bump the version of the allowed umbrella chart for upgrade when a release/x.y branch is created

niladrih:
It is easy to miss this when a release branch is cut. We do not have automated tests for umbrella chart upgrade at the moment. This change should bring this change to notice (if the version is not bumped correctly).

### DIFF
--- a/.github/workflows/develop-to-release.yml
+++ b/.github/workflows/develop-to-release.yml
@@ -18,6 +18,7 @@ jobs:
         run: |
           branch="${{ github.ref_name }}"
           nix-shell --pure --run "./scripts/helm/publish-chart-yaml.sh --check-chart "$branch" --develop-to-release" ./scripts/helm/shell.nix
+          nix-shell --pure --run "./scripts/upgrade/develop-to-release.sh" ./shell.nix
           nix-shell --pure --run "SKIP_GIT=1 ./scripts/helm/generate-readme.sh" ./scripts/helm/shell.nix
       - name: Check if the submodules are correct
         run: |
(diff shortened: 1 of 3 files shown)


tiagolobocastro:
Seems like this should be driven down, rather than up? Though either way it's quite difficult to figure out versions.
Could you please remind me, why do we need to check the umbrella chart version again?

niladrih:
> Could you please remind me, why do we need to check the umbrella chart version again?

The upgrade-job is tied to a single version of Mayastor. It might be that a user uses this upgrade-job with a different version of Mayastor from a different openebs/openebs chart. The version 'lock' prevents this.

tiagolobocastro:
> > Could you please remind me, why do we need to check the umbrella chart version again?
> 
> The upgrade-job is tied to a single version of Mayastor. It might be that a user uses this upgrade-job with a different version of Mayastor from a different openebs/openebs chart. The version 'lock' prevents this.

What happens if we release a new umbrella chart, with no mayastor changes?

niladrih:
> > > Could you please remind me, why do we need to check the umbrella chart version again?
> > 
> > The upgrade-job is tied to a single version of Mayastor. It might be that a user uses this upgrade-job with a different version of Mayastor from a different openebs/openebs chart. The version 'lock' prevents this.
> 
> What happens if we release a new umbrella chart, with no mayastor changes?

The upgrade-job is not prepared to handle this. The new umbrella chart cannot be accommodated, with the way it is today. This is a bug.

I'll try to fix this. I'll try to remove this dependency on the version number.

niladrih:
This change is not required. It'd be better to verify the mayastor version attached to an installed umbrella chart release. Helm doesn't support viewing this information from the helm secret/configmap yet. Maybe [this](https://github.com/helm/helm/issues/12686) could resolve this. Alternatively, reading directly from the secret/configmap could be a good solution.

PR #549, opened 2024-09-30
feat(helm): add list of container images

tiagolobocastro:
Adds a list of container images which are used by the chart.
This list is contained in file: charts/images.txt

Sometimes the images can't be figured out using templates, and in such
cases we need to manually add them.
In some cases we might be able to do this using helm variables, which
is what we're doing with the localpv provisioner.

Github actions have been modified to verify the contents of the images
as well as testing them with a live cluster.
This is a pretty dumb test that just checks if the live pod images are
contained in the images files.


scripts/helm/template-images.sh
```diff
@@ -0,0 +1,72 @@
...
+  esac
+done
+
+cd "$CHART_DIR"
+
+if [ "$DEP_UPDATE" = "true" ]; then
+    $HELM dependency update
+fi
+
+$HELM template . --set "$ENABLE_ANALYTICS" | grep "image:" | awk '{ print $2 }' | tr -d \" > "$IMAGES"
```
Abhinandan-Purkait: Enable analytics is not needed, the analytics code is embedded in the csi-controller image. But anyway since this is helm template should be fine.

niladrih: nit: `--dependency-update` does the same thing
```suggestion
dependency_update_opt=
if [ "$DEP_UPDATE" = "true" ]; then
    dependency_update_opt="--dependency-update"
fi

$HELM template . --set "$ENABLE_ANALYTICS" "$dependency_update_opt" | grep "image:" | awk '{ print $2 }' | tr -d \" > "$IMAGES"
```

scripts/helm/verify-kind-images.sh
```diff
@@ -0,0 +1,52 @@
...
+
+echo_stderr() {
+  echo -e "${1}" >&2
+}
+
+die() {
+  local _return="${2:-1}"
+  echo_stderr "$1"
+  exit "${_return}"
+}
```
niladrih: There's /scripts/utils/log.sh you could source and use log_fatal or log_error.

scripts/helm/verify-kind-images.sh
```diff
@@ -0,0 +1,52 @@
...
+
+grep_image() {
+  local image="$1"
+  local file="$2"
+
+  grep -q "^$image$" "$file" || grep -q "^${image##docker.io/}$" "$file"
```
niladrih: The Hostpath LocalPV images don't start with "docker.io".

scripts/helm/verify-kind-images.sh
```diff
@@ -0,0 +1,52 @@
...
+
+cd "$TMP_KIND"
+for worker in kind-worker*; do
+  if ! [ -f "$worker/$KIND_IMAGES" ]; then
+    die "Missing $KIND_IMAGES file"
+  fi
+
+  curr_images=$($DOCKER exec $worker crictl image | tail -n+2 | awk '{ print $1 ":" $2 }')
+  for image in ${curr_images[@]}; do
+    if grep_image "$image" "$worker/$KIND_IMAGES"; then
```
niladrih: Does this also list the pause image?

scripts/helm/template-images.sh
tiagolobocastro: It's safer to enable all many options as we can, maybe tomorrow that can change, who knows.
Also we can enable jaeger here, just for completeness, though it's again tricky because it's operator based :(

tiagolobocastro: It requires more changes :)
Other than that, any advantage?

scripts/helm/verify-kind-images.sh
tiagolobocastro: Ok, are you suggesting any changes here?

scripts/helm/verify-kind-images.sh
tiagolobocastro: Probably yes

scripts/helm/verify-kind-images.sh
Abhinandan-Purkait: Right, sometimes the registry might not be specified, do we handle those cases?

tiagolobocastro: I'm confused.. it's handled in line 29, what's the problem?

scripts/helm/template-images.sh
niladrih: Less no of helm commands :). No advantage as such.

scripts/helm/verify-kind-images.sh
niladrih: In effect, there may be no changes required. `docker.io` doesn't add or remove anything meaningful to/from the image name.

scripts/helm/verify-kind-images.sh
niladrih: Okay.

scripts/helm/verify-kind-images.sh
tiagolobocastro: I'm confused on what's the ask here?

niladrih: I might have misinterpreted the line. I'm good here. I'm going to resolve this thread.

tiagolobocastro:
bors try

chart/images.txt
```diff
@@ -0,0 +1,29 @@
+bats/bats:1.8.2
```
Abhinandan-Purkait, line 1: Is this file needed?

file.yml
```diff
@@ -0,0 +1,4 @@
+a:
```
Abhinandan-Purkait: What is this file?

Abhinandan-Purkait (approved):

chart/images.txt
tiagolobocastro: Yeah, this keeps an exact list of images

file.yml
tiagolobocastro: doh, testing :D will delete

dsharma-dc (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
Addresses https://github.com/openebs/openebs/issues/3796 for the mayastor chart

PR #555, opened 2024-10-10
ci: add script to run pytests

niladrih:


Abhinandan-Purkait (approved):
Looks good. 🚀

Abhinandan-Purkait:
Thanks for adding this. We will now have some way of testing. 👍🏻

scripts/k8s/load-images-to-kind.sh
```diff
@@ -0,0 +1,59 @@
...
+  esac
+  shift
+done
+
+if [ -z "$TAG" ]; then
+  log_fatal "requires an image tag"
+fi
+
+IMAGE_TAG="v${TAG#v}"
+images=("upgrade-job" "obs-callhome" "obs-callhome-stats" "metrics-exporter-io-engine")
```
tiagolobocastro, line 61: Can this be loaded from other places? ex: sourcing release.sh or something

scripts/python/generate-test-tag.sh
tiagolobocastro, line 1: what is a test tag?

scripts/python/generate-test-tag.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]:-"$0"}")")"
+ROOT_DIR="$SCRIPT_DIR/../.."
+
+set -e
+
+CHART_VERSION=$(yq -eM '.version' "$ROOT_DIR/chart/Chart.yaml" | tr -d '[:space:]')
```
tiagolobocastro: Does this not work?
`yq '.version' "$CHART_FILE"`
We already fetch the chart version like this, would be good to keep consistent, either here or there

scripts/python/shell.nix
tiagolobocastro, line 1: Should we just add this to the root shell.nix?

scripts/python/test.sh
tiagolobocastro, line 1: Can we reuse the one from control-plane rather than maintaining and using different ones.
At the least they should work exactly the same

scripts/python/generate-test-tag.sh
niladrih: This is to reach a tag which looks like 'chart.next'. Like 'helm-core'. On develop, it's the next minor version, on the release branches it's the next patch version. When I say develop and release, I mean those two kinds of Chart.yaml file `.version` values, as PR branches are seldom named 'develop' or 'release/x.y'.

niladrih: ... for upgrade. But works for the rest as well, usually.

niladrih:
The RUSTFLAGS changes don't seem to making much of a difference
```rust
      RUSTFLAGS =
        if static then
          [ "-C" "target-feature=+crt-static" ]
        else
          [ "-C" "opt-level=1" "-C" "debuginfo=1" "-C" "codegen-units=32" ];
```
Without these, it's 12min something, and with this it's 13min something. I'm removing them. The dynamically linked binary is cutting most of the time. Seems fine, WDYT @tiagolobocastro ?

tiagolobocastro:
> The RUSTFLAGS changes don't seem to making much of a difference
> 
> ```rust
>       RUSTFLAGS =
>         if static then
>           [ "-C" "target-feature=+crt-static" ]
>         else
>           [ "-C" "opt-level=1" "-C" "debuginfo=1" "-C" "codegen-units=32" ];
> ```
> 
> Without these, it's 12min something, and with this it's 13min something. I'm removing them. The dynamically linked binary is cutting most of the time. Seems fine, WDYT @tiagolobocastro ?

Are you doing a debug build? What's the difference between release vs debug ?

scripts/python/generate-test-tag.sh
tiagolobocastro: This functionality already exists in the publish script IIRC?

niladrih: No, that script has strict boundaries for chart generation paths. The chart change requirements for this test violates these boundaries.

How about leaving this one in, and reworking the publish-chart-yaml.sh user API to support these one-off use-cases, at a later date?

In short, the publish chart yaml doesn't like to let develop branch charts pretend to be release branch charts, and upgrades expect some form of mock release and requires the resultant release chart to work.

scripts/python/generate-test-tag.sh
niladrih: Sure, it works. This command is no different, but anyhow I'll use the one you've specified. It'd be all the same. 👍

scripts/python/shell.nix
niladrih: Sure, no problem.

scripts/python/test.sh
niladrih: Sure, sounds good.

scripts/k8s/load-images-to-kind.sh
niladrih: The exact image names are built using string concat on this file: [nix/pkgs/images/default.nix](https://github.com/openebs/mayastor-extensions/blob/develop/nix/pkgs/images/default.nix)

The release.sh contains the nix-build targets, which are not the container image namespace-name thing.
I could add in a TODO to generate this list from the nix file somehow for a later date, WDYT?

scripts/python/test.sh
niladrih: Hmm, on second look... the control-plane script uses the control-plane's tests/bdd/setup.sh script, which uses the control-plane's requirements.txt when building the virtualenv. We'd have to make the script a bit more extensible so that it's able to accept the path to this repo's requirements.txt or even the path to this repo's tests/bdd/setup.sh.

Maybe this should be done at a later date?

niladrih:
> > The RUSTFLAGS changes don't seem to making much of a difference
> > ```rust
> >       RUSTFLAGS =
> >         if static then
> >           [ "-C" "target-feature=+crt-static" ]
> >         else
> >           [ "-C" "opt-level=1" "-C" "debuginfo=1" "-C" "codegen-units=32" ];
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Without these, it's 12min something, and with this it's 13min something. I'm removing them. The dynamically linked binary is cutting most of the time. Seems fine, WDYT @tiagolobocastro ?
> 
> Are you doing a debug build? What's the difference between release vs debug ?

@tiagolobocastro -- Using debug builds cuts the time down to 11m36s. Seems worthwhile to me. Using debug builds. PTAL.

scripts/k8s/load-images-to-kind.sh
tiagolobocastro: Ah right, maybe this can be addressed with the oep I raised
yep add a todo for now, thanks

scripts/python/test.sh
tiagolobocastro: Sure, using the same script can be done later, but keep the api the same, so remove the extra args on this one so at least it can be used the same way

scripts/python/generate-test-tag.sh
tiagolobocastro: Alright, I suggest your at least change this script to perhaps reuse the other scripts functions, maybe the `latest_release_branch`.
And use branches rater than tags, so it's compatible with unreleased upgrades, ie from unreleased 2.8.0 for example to develop 2.9.0

scripts/python/test.sh
niladrih: Yeah, that's a good idea.

scripts/python/generate-test-tag.sh
niladrih: Okay. Sounds good.

niladrih:
BTW, just hit ~8min run time 🍾 .

niladrih:
@tiagolobocastro -- I'm using debug builds, while also stripping away debug symbols. Also disabling the generation of debug symbols in the first place
```
      RUSTFLAGS =
        if static then
          [ "-C" "target-feature=+crt-static" ]
        else
          [ "-C" "strip=debuginfo" "-C" "debuginfo=0" ];
```

nix/pkgs/extensions/cargo-project.nix
```diff
@@ -133,7 +133,11 @@ let
         export OPENSSL_LIB_DIR=${static_ssl.out}/lib
         export OPENSSL_INCLUDE_DIR=${static_ssl.dev}/include
       '';
-      ${if static then "RUSTFLAGS" else null} = [ "-C" "target-feature=+crt-static" ];
+      RUSTFLAGS =
+        if static then
+          [ "-C" "target-feature=+crt-static" ]
+        else
+          [ "-C" "strip=debuginfo" "-C" "debuginfo=0" ];
```
tiagolobocastro: removal of debug should be opt-in

scripts/k8s/load-images-to-kind.sh
```diff
@@ -0,0 +1,69 @@
...
+  --tag)
+    test $# -lt 2 && log_fatal "missing value for the argument '$arg'"
+    TAG=$2
+    shift
+    ;;
+  --tag=*)
+    TAG=${arg#*=}
+    ;;
+  --trim-debug-suffix)
+    TRIM_DEBUG_SUFFIX=1
```
tiagolobocastro, line 40: maybe adding -debug to begin with was a mistake, maybe need to consider removing it in the future

scripts/python/tag-chart.sh
```diff
@@ -0,0 +1,34 @@
...
+# No tag specified.
+if [ -z "$1" ]; then
+  log_fatal "no tag specified"
+fi
+
+CHART_VERSION=${1#v}
+IMAGE_TAG="v$CHART_VERSION"
+CHART_DIR="$ROOT_DIR/chart"
+# chart/Chart.yaml
+yq_ibl "
```
tiagolobocastro, line 21: I wonder if we should clone  the chart, rather than modify it, not sure either way

.github/workflows/k8s-ci.yml
```diff
@@ -8,6 +8,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
```
tiagolobocastro: Not needed?

```suggestion
```

.github/workflows/k8s-ci.yml
```diff
@@ -16,18 +20,29 @@ jobs:
         run: |
           export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
           echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell ./shell.nix --run 'echo'
```
tiagolobocastro:
```suggestion
```

.github/workflows/k8s-ci.yml
```diff
@@ -16,18 +20,29 @@ jobs:
...
+          ./scripts/release.sh --tag $TAG --build-bins --build-binary-out $BIN --no-static-linking --skip-publish --debug
+          echo "tag=$TAG" >> $GITHUB_OUTPUT
+          echo "bin=$BIN" >> $GITHUB_OUTPUT
       - name: BootStrap k8s cluster
         run: |
           nix-shell ./scripts/k8s/shell.nix --run "./scripts/k8s/deployer.sh start --label"
-      - name: Install Helm Chart
-        run: |
-          nix-shell ./scripts/k8s/shell.nix --run "./scripts/helm/install.sh --dep-update --wait"
       - name: The job has failed
```
tiagolobocastro, line 38: Move this to the end?

tiagolobocastro (approved):

scripts/python/tag-chart.sh
niladrih: That's a good way to go IMO. I'll add a todo for now. But I agree that modifying the chart is a bad idea.

scripts/k8s/load-images-to-kind.sh
niladrih: Hmm, have to consider what was the reason for the debug. Maybe something depends on it?

nix/pkgs/extensions/cargo-project.nix
niladrih: Alright

niladrih: I'm inputting the RUSTFLAGS using an env.
```
RUSTFLAGS="-C debuginfo=0 -C strip=debuginfo"
```

scripts/python/tag-chart.sh
niladrih: I've added a TODO.

niladrih:
bors merge

PR #557, opened 2024-10-15
expose mayastor plugin as lib

abhilashshetty04:
With this PR, kubectl-plugin is exposed as lib. 

This has few more changes w.r.t kubectl-openebs plugin also.

### DIFF
--- a/k8s/plugin/src/main.rs
+++ b/k8s/plugin/src/main.rs
@@ -1,23 +1,22 @@
 use clap::Parser;
-use openapi::tower::client::Url;
-use plugin::{rest_wrapper::RestClient, ExecuteOperation};
-use resources::Operations;
+use plugin::ExecuteOperation;
+use resources::{init_rest, Error, Operations};
 
 use std::{env, ops::Deref};
 
-mod resources;
+pub mod resources;
 
 #[derive(Parser, Debug)]
 #[clap(name = utils::package_description!(), version = utils::version_info_str!())]
 #[group(skip)]
 struct CliArgs {
-    /// The rest endpoint to connect to.
-    #[clap(global = true, long, short)]
-    rest: Option<Url>,
-
     /// The operation to be performed.
     #[clap(subcommand)]
     operations: Operations,
+
+    /// Kubernetes namespace of mayastor service
+    #[clap(global = true, long, short = 'n', default_value = "mayastor")]
+    namespace: String,
 
     #[clap(flatten)]
     args: resources::CliArgs,
@@ -25,9 +24,12 @@ struct CliArgs {
 
 impl CliArgs {
     fn args() -> Self {
-        CliArgs::parse()
+        let mut args = CliArgs::parse();
+        args.args.namespace = args.namespace.clone();
+        args
     }
 }
+
 impl Deref for CliArgs {
     type Target = plugin::CliArgs;
 
@@ -61,7 +63,7 @@ async fn main() {
 impl CliArgs {
     async fn execute(self) -> Result<(), Error> {
         // Initialise the REST client.
-        init_rest(&self).await?;
+        init_rest(&self.args).await?;
 
         tokio::select! {
             shutdown = shutdown::Shutdown::wait_sig() => {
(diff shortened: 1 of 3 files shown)


k8s/plugin/Cargo.toml
```diff
@@ -8,6 +8,10 @@ edition = "2021"
 name = "kubectl-mayastor"
 path = "src/main.rs"
 
+[lib]
+name = "kubectl_plugin"
+path = "./src/lib.rs"
+
```
tiagolobocastro: Don't think it's needed to be specified?
 
```suggestion
```

k8s/plugin/src/main.rs
```diff
@@ -10,11 +9,7 @@ mod resources;
 #[derive(Parser, Debug)]
 #[clap(name = utils::package_description!(), version = utils::version_info_str!())]
 #[group(skip)]
-struct CliArgs {
-    /// The rest endpoint to connect to.
-    #[clap(global = true, long, short)]
-    rest: Option<Url>,
-
+pub struct CliArgs {
```
tiagolobocastro: Why is this made pub?

k8s/plugin/src/resources/mod.rs
```diff
@@ -17,16 +18,20 @@ use upgrade::{
...
+
     /// Path to kubeconfig file.
     #[clap(global = true, long, short = 'k')]
-    pub(super) kube_config_path: Option<PathBuf>,
+    pub kube_config_path: Option<PathBuf>,
 
     /// Kubernetes namespace of mayastor service
     #[clap(global = true, long, short = 'n', default_value = "mayastor")]
-    pub(super) namespace: String,
+    pub namespace: String,
```
tiagolobocastro, line 31: Please move namespace to the binary since it doesn't make sense to default to mayastor on the openebs umbrella plugin.
Also I wonder if rest and kube_config_path should also be part of the binary. Not sure on this one, but at least init_rest should still be called outside of the execute, on the main

k8s/plugin/src/resources/mod.rs
```diff
@@ -124,3 +131,46 @@ impl ExecuteOperation for Operations {
         Ok(())
     }
 }
+
+pub enum Error {
```
tiagolobocastro, line 135: missing doc comments

k8s/plugin/src/resources/mod.rs
abhilashshetty04: We are not calling main function of this crate from unified plugin. We directly run `Operations.execute` . If `init_rest` is in main function then we might have to do `init_rest` on `oss_plugin` main also. That will cause issues incase mayastor is not installed on cluster?

Here is the draft PR of unified plugin
https://github.com/openebs/openebs/pull/3794

k8s/plugin/src/main.rs
abhilashshetty04: I am using CliArgs from this crate. I have created it in `kubectl-openebs` also but that CliArgs points to this ClArgs

k8s/plugin/src/resources/mod.rs
tiagolobocastro: You can call the init from the unified plugin on the mayastor execute on the unified.

k8s/plugin/src/main.rs
tiagolobocastro: It can't be pointing to this CliArgs, this is part of the binary, not the library

k8s/plugin/Cargo.toml
abhilashshetty04: It was not required. I wonder why we have defined it in some toml?

k8s/plugin/src/main.rs
abhilashshetty04: I am using CliArgs from lib. Ill remove this pub

Abhinandan-Purkait (approved):

niladrih (approved):

tiagolobocastro (commented):
commits should be refactor and not chore
@abhilashshetty04  is this the PR you had mentioned will remain open? Any reason to leave it open and if so, then should it stay as a draft?
Alternatively I suggest creating a feature branch instead, and raise and merge PRs to feature branch as required

abhilashshetty04:
> commits should be refactor and not chore @abhilashshetty04 is this the PR you had mentioned will remain open? Any reason to leave it open and if so, then should it stay as a draft? Alternatively I suggest creating a feature branch instead, and raise and merge PRs to feature branch as required

Hi @tiagolobocastro , I had kept it open to see if any changes required on CLI output specific code. Im dont with all changes here. Ill merge it after amending commit.

abhilashshetty04:
bors merge

PR #545, opened 2024-09-20
Add workflow to update the HelmChart post-release

tiagolobocastro:
build: update control-plane submodule
    
    Signed-off-by: Tiago Castro <tiagolobocastro@gmail.com>

---

    chore: manually bump chart to v2.7.2
    
    Signed-off-by: Tiago Castro <tiagolobocastro@gmail.com>

---

    fix(helm-chart): package README.md in the chart
    
    Signed-off-by: Tiago Castro <tiagolobocastro@gmail.com>

---

    ci: bump release chart version
    
    When a tag is pushed on a release branch we update the future
    chart release version by bumping it's patch.
    
    Signed-off-by: Tiago Castro <tiagolobocastro@gmail.com>

PR #561, opened 2024-10-28
test: fix chart.next version on release branches by removing semver bump

niladrih:
The tag generation logic reads the chart/Chart.yaml version and bumps it to the next patch version. This is not required since 2.7.0, as the helm chart is updated automatically after a tag is cut. Removing the semver bump patch.

Also, adds a library function to figure out the latest tag on a branch.

## Types of changes

## Checklist:

### DIFF
--- a/scripts/python/generate-test-tag.sh
+++ b/scripts/python/generate-test-tag.sh
@@ -52,7 +52,7 @@ done
 
 case "$CHART_VERSION" in
 0.0.0)
-  latest_branch=$(git fetch -q && latest_release_branch $GIT_REMOTE $ROOT_DIR)
+  latest_branch=$(git fetch -q && latest_release_branch "$GIT_REMOTE" "$ROOT_DIR")
   latest=${latest_branch#release/}
   if [[ "$latest" =~ ^([0-9]+\.[0-9]+)$ ]]; then
     latest="$latest.0"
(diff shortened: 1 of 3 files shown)


tiagolobocastro:
> The tag generation logic doesn't account for the latest tag on a branch

Why does it need to take that into account to begin with?

niladrih:
> > The tag generation logic doesn't account for the latest tag on a branch
> 
> Why does it need to take that into account to begin with?

Upgrade tests on a branch needs to concern itself with the next patch release, instead of the next minor/major release.
The Chart.yaml isn't updated whenever a tag is cut. That would mean, the latest tag could be 2.7.3 and the Chart.yaml would say 2.7.0. Upgrade should try to go to 2.7.4, instead of 2.7.1.

The initial chart would pull in 2.7.3 or whatever is the latest one (this bit is not in this PR, nor will it use any of the scripts on the PR.)

tiagolobocastro:
> Upgrade tests on a branch needs to concern itself with the next patch release, instead of the next minor/major release. The Chart.yaml isn't updated whenever a tag is cut. That would mean, the latest tag could be 2.7.3 and the Chart.yaml would say 2.7.0. Upgrade should try to go to 2.7.4, instead of 2.7.1.

This is no longer the case as of: https://github.com/openebs/mayastor-extensions/pull/545

niladrih:
@tiagolobocastro fair enough. I've removed the `semver bump patch` that the script was doing for release branches.

Leaving the function in as a lib asset. Not used anywhere.

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

PR #564, opened 2024-10-30
Upgrade basic sanity test

niladrih:
Changes:
- Adds a pytest for testing basic sanity in upgrade
- Add enhancement to the scripts/helm/install.sh script to support installing from a hosted registry
- Add a script to format python files

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,7 @@
+[pytest]
+log_cli = true
+log_cli_level = INFO
+log_cli_format = %(asctime)s [%(levelname)s] %(message)s
+log_cli_date_format = %Y-%m-%d %H:%M:%S
+log_file = pytest.log
+log_file_level = DEBUG
--- a/shell.nix
+++ b/shell.nix
@@ -18,6 +18,8 @@ in
 mkShell {
   name = "extensions-shell";
   buildInputs = [
+    autoflake
+    black
     cacert
     cargo-expand
     cargo-udeps
@@ -27,6 +29,7 @@ mkShell {
     cowsay
     git
     helm-docs
+    isort
     kubectl
     kubernetes-helm-wrapped
     llvmPackages.libclang
--- a/tests/bdd/common/helm.py
+++ b/tests/bdd/common/helm.py
@@ -1,0 +1,272 @@
+import json
+import logging
+import os
+import subprocess
+from enum import Enum
+from shutil import which
+
+from common.environment import get_env
+from common.repo import root_dir, run_script
+
+logger = logging.getLogger(__name__)
+
+helm_bin = which("helm")
+
+
+def repo_ls():
+    try:
+        result = subprocess.run(
+            [helm_bin, "repo", "ls", "-o", "json"],
+            capture_output=True,
+            check=True,
+            text=True,
+        )
+        return json.loads(result.stdout.strip())
+
+    except subprocess.CalledProcessError as e:
+        logger.error(
+            f"Error: command 'helm repo ls -o json' failed with exit code {e.returncode}"
+        )
+        logger.error(f"Error Output: {e.stderr}")
+        return None
+
+    except Exception as e:
+        logger.error(f"An unexpected error occurred: {e}")
+        return None
+
+
+def repo_add_mayastor():
+    repos = repo_ls()
+    if repos is not None:
+        for r in repos:
+            if r["url"] == "https://openebs.github.io/mayastor-extensions":
+                return r["name"]
+
+    try:
+        repo_name = "mayastor"
+        subprocess.run(
+            [
+                helm_bin,
+                "repo",
+                "add",
+                repo_name,
+                "https://openebs.github.io/mayastor-extensions",
+            ],
+            capture_output=True,
+            check=True,
+            text=True,
+        )
+
+        subprocess.run(
+            [
+                helm_bin,
+                "repo",
+                "update",
+            ],
+            capture_output=True,
+            check=True,
+            text=True,
+        )
+        return repo_name
+
+    except subprocess.CalledProcessError as e:
+        logger.error(
+            f"Error: command 'helm repo add mayastor https://openebs.github.io/mayastor-extensions' failed with exit code {e.returncode}"
+        )
+        logger.error(f"Error Output: {e.stderr}")
+        return None
+
+    except Exception as e:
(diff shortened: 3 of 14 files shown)


shell.nix
```diff
@@ -18,6 +18,8 @@ in
 mkShell {
   name = "extensions-shell";
   buildInputs = [
+    autoflake
```
tiagolobocastro, line 21: what's autoflake and isort?

tests/bdd/common/assertion.py
```diff
@@ -0,0 +1,18 @@
+import time
+
+
+def retry_until_true(closure, retry_seconds=1, max_retries=5):
+    """
+    Runs the provided closure until it returns True, with a specified retry interval
```
tiagolobocastro: just use the retry package?

tests/bdd/common/kubectl_mayastor.py
```diff
@@ -0,0 +1,39 @@
...
+    if bins:
+        return os.path.join(bins, "kubectl-mayastor")
+    logging.warning(f"Environmental variable 'BIN' is not set")
+    return "kubectl-mayastor"
+
+
+def kubectl_mayastor(args: str):
+    command = [
+        "/bin/bash",
+        "-c",
```
tiagolobocastro: can you not run the bin path directly?

tests/bdd/common/assertion.py
niladrih: Oh, thanks!

shell.nix
niladrih: They format the import sequence on the .py files.

tests/bdd/common/kubectl_mayastor.py
niladrih: Yes, I think running it directly would work.

niladrih:
Moving @tiagolobocastro's review comment here so that I can track it.
>```python
>        install_command = []
>        if source == ChartSource.HOSTED:
>            repo_name = repo_add_mayastor()
>```
>ah ok, we want to be able to do both, got it.
Let's change the script to support this as well, rather than have separate commands here

shell.nix
niladrih: Correction. isort sorts the imports, and autoflake removes unused imports.

tests/bdd/common/helm.py
```diff
@@ -0,0 +1,306 @@
...
+
+            install_command += [
+                helm_bin,
+                "install",
+                "mayastor",
+                repo_name + "/" + source.value,
+                "--namespace=" + self.namespace,
+                "--create-namespace",
+                "--set",
+                "obs.callhome.sendReport=false,localpv-provisioner.analytics.enabled=false,etcd.livenessProbe.initialDelaySeconds=5,etcd.readinessProbe.initialDelaySeconds=5,etcd.replicaCount=1,eventing.enabled=false",
```
tiagolobocastro: let's use the script for this

tests/bdd/features/test_upgrade.py
```diff
@@ -0,0 +1,87 @@
...
+    """the installed chart should be upgraded to the kubectl mayastor plugin's version."""
+
+    upgrade_target_version = get_env("UPGRADE_TARGET_VERSION")
+    if upgrade_target_version is None:
+        upgrade_target_version = repo.run_script("scripts/python/generate-test-tag.sh")
+    upgrade_target_version = upgrade_target_version.lstrip("v")
+    logger.info(f"Value of upgrade_target_version={upgrade_target_version}")
+
+    @retry(
+        stop_max_attempt_number=15,
```
tiagolobocastro: suggest you check with more tries and less interval each time

tests/bdd/requirements.txt
```diff
@@ -1,2 +1,7 @@
-pytest-bdd==7.3.0
+gitpython==3.1.43
```
tiagolobocastro: what gitpython for?

pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
+log_cli = true
+log_cli_level = INFO
+log_cli_format = %(asctime)s [%(levelname)s] %(message)s
+log_cli_date_format = %Y-%m-%d %H:%M:%S
+log_file = pytest.log
```
tiagolobocastro, line 6: Should we upload the logs on failure?

tests/bdd/common/helm.py
niladrih: Okay.

tests/bdd/requirements.txt
niladrih: That's not needed any more.

tests/bdd/features/test_upgrade.py
niladrih: Instead of checking for a maximum of 15 times at an interval of 1 min, I've now resorted to checking for a maximum of 450 times at an interval of 2 seconds.

On the other function...
Instead of checking for a maximum of 20 times at an interval of 1 min, I've now resorted to checking for a maximum of 600 times at an interval of 2 seconds.

tests/bdd/common/helm.py
niladrih: Done.

pytest.ini
niladrih: Hmm, like control-plane? I'd defer that to a different PR.

niladrih:
> Moving @tiagolobocastro's review comment here so that I can track it.
> 
> > ```python
> >        install_command = []
> >        if source == ChartSource.HOSTED:
> >            repo_name = repo_add_mayastor()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ah ok, we want to be able to do both, got it.
> > Let's change the script to support this as well, rather than have separate commands here

This change is added to this PR.

dsharma-dc (approved):

abhilashshetty04 (approved):

niladrih:
bors merge

pytest.ini
niladrih: Resolving this thread for now. The intention here is to 'init' a test-run CI on this repo. And I think it succeeds in doing so. I agree with the suggestion above. I could see it happen in a different PR from this one. Thanks.

niladrih:
bors merge

PR #567, opened 2024-11-06
build: build static binaries with latest pkgs

tiagolobocastro:
On nixpkgs bug where windows platform is set incorrectly for static
packages: 281596
This can be worked around by using pkgStatic.pkgsCross rather than
the other way around, or simply by replacing the package.
As it happens, the tls build for windows doesn't need openssl, so
we can remove it entirely and skip this issue altogether.
On darwin, libiconv is failing to build, but also seems to not be
required anymore, and so also removing it!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2144,6 +2144,7 @@ dependencies = [
  "kube 0.94.2",
  "serde_json",
  "shutdown",
+ "thiserror",
  "tokio",
  "tokio-stream",
  "tower 0.5.1",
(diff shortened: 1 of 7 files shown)


Abhinandan-Purkait (approved):

niladrih (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
hmm weird build error.. let's try again
bors merge

PR #568, opened 2024-11-08
chore: add csi-node nvme-connect-fallback to helm values

dsharma-dc:
## Description
Add `nvme-connect-fallback` option to helm for allowing users to let initiator node connect over tcp in case the node is not rdma capable, even though volume target exposes rdma transport.

## How Has This Been Tested?
By templating the helm values.

## Types of changes

## Checklist:

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -349,6 +349,8 @@ csi:
       ctrl_loss_tmo: "1980"
       # Kato (keep alive timeout) in seconds
       keep_alive_tmo: ""
+      # -- Fallback to nvme-tcp if nvme-rdma is enabled for Mayastor but rdma is not available on a particular csi-node
+      tcpFallback: true
     # -- The kubeletDir directory for the csi-node plugin
     kubeletDir: /var/lib/kubelet
     pluginMountPath: /csi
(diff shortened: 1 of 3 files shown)


niladrih (approved):
Ideally all of our snake_case yaml keys should be camelCase, i.e. nvmeConnectFallback. However we don't do that often. Just a heads-up.

chart/values.yaml
```diff
@@ -349,6 +349,8 @@ csi:
       ctrl_loss_tmo: "1980"
       # Kato (keep alive timeout) in seconds
       keep_alive_tmo: ""
+      # Fallback to nvme connect over tcp if host is not rdma capable
+      nvme_connect_fallback: false
```
tiagolobocastro: This should be true by default?

dsharma-dc: > This should be true by default?

The default is kept `false` to let users choose explicitly if they wish to fallback connect, and not taking a decision on their behalf.

dsharma-dc: As discussed, made it `true` for now.

chart/values.yaml
```diff
@@ -349,6 +349,8 @@ csi:
       ctrl_loss_tmo: "1980"
       # Kato (keep alive timeout) in seconds
       keep_alive_tmo: ""
+      # Fallback to nvme connect over tcp if host is not rdma capable
```
tiagolobocastro: Maybe something like this?

```suggestion
      # -- Fallback to nvme-tcp if nvme-rdma is enabled but not available on a particular csi-node
```

chart/values.yaml
```diff
@@ -349,6 +349,8 @@ csi:
...
+      nvme_connect_fallback: true
```
tiagolobocastro: Maybe something like this?

```suggestion
      tcpFallback: true
```

tiagolobocastro (approved):

chart/values.yaml
dsharma-dc: > Maybe something like this?

I can change this but we'll have inconsistency then, with other existing fields. wdyt?

tiagolobocastro: hmm can't remember why the names are snake case, perhaps in order to match the kernel parameter names?
In this case this is not a known kernel parameter name?
(maybe it had been better to name these as camelCase and add list kernel parameter name as a comment, my bad)

sinhaashish (approved):

chart/values.yaml
dsharma-dc: done

chart/values.yaml
dsharma-dc: converted to camelCase for this new value.

dsharma-dc:
bors merge

PR #576, opened 2024-12-02
fix(supportability): skip logs for pods with missing details

sinhaashish:
The supoortability logs errors out if some of the pods are in pending state as the node is not populated for those pods. 

image

In this case the etcd is in pending state . the command `kubectl mayastor system dump ` doesnt create the logs folder and error out  stating 
`Error occurred while collecting logs`

In-oder to rectify , the supportability collects logs for pods if its not in pending state . 

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/k8s/supportability/src/collect/logs/mod.rs
+++ b/k8s/supportability/src/collect/logs/mod.rs
@@ -14,8 +14,12 @@ use crate::collect::{
     utils::log,
 };
 use async_trait::async_trait;
-use k8s_openapi::api::core::v1::Pod;
-use std::{collections::HashSet, iter::Iterator, path::PathBuf};
+use k8s_openapi::api::core::v1::{Node, Pod};
+use std::{
+    collections::{HashMap, HashSet},
+    iter::Iterator,
+    path::PathBuf,
+};
 
 /// Error that can occur while interacting with logs module
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


k8s/supportability/src/collect/logs/mod.rs
```diff
@@ -137,6 +137,28 @@ impl LogCollection {
 
             let mut hostname = None;
             if is_host_name_required(service_name.clone()) {
+                let status = pod.status.clone().ok_or_else(|| {
```
tiagolobocastro:
```suggestion
                let status = pod.status.as_ref().ok_or_else(|| {
```

k8s/supportability/src/collect/logs/mod.rs
```diff
@@ -137,6 +137,28 @@ impl LogCollection {
...
+                    K8sResourceError::invalid_k8s_resource_value(format!(
```
tiagolobocastro: This could still potentially throw an error if we the pod is being recreated, and we catch it without a state.
I suggest you do this bit where the second function is per-pod on a separate function, and then we can catch any error, example:

```
for pod in pods {
    match self.pod_logging_resources() {
        Ok(resources) => logging_resources.extend(resources),
        Err(error) => log(format!(
                        "Skipping the pod in pending state: {:?}",
                        pod.metadata.name
                    )),
    }
}
```

I suspect we may also have this type of issue elsewhere in the support bundle code, probably we need to analyse the rest of the code at some point.

sinhaashish: done

k8s/supportability/src/collect/logs/mod.rs
sinhaashish: removed the fetching status and `Pending` logic

k8s/supportability/src/collect/logs/mod.rs
```diff
@@ -109,95 +113,114 @@ impl LogCollection {
...
+                    })?
+                    .clone(),
+            );
+        }
+        // Since pod object fetched from Kube-apiserver there will be always
+        // spec associated to pod
+        let containers = pod
+            .spec
+            .ok_or_else(|| {
+                K8sResourceError::invalid_k8s_resource_value("Pod sepc not found".to_string())
```
Abhinandan-Purkait: Pod sepc --> pod spec

Abhinandan-Purkait (approved):

k8s/supportability/src/collect/logs/mod.rs
```diff
@@ -109,95 +113,114 @@ impl LogCollection {
...
+
+        for pod in pods {
+            match self
+                .pod_logging_resources(pod.clone(), nodes_map.clone())
+                .await
+            {
+                Ok(resources) => logging_resources.extend(resources),
+                Err(error) => log(format!(
+                    "Skipping the pod in pending state: {:?} error : {:?}",
+                    pod.metadata.name, error
```
tiagolobocastro: Not just pending, can be any error

```suggestion
                    "Skipping the pod {:?} due to error: {error:?}", pod.metadata.name
```

k8s/supportability/src/collect/logs/mod.rs
```diff
@@ -109,95 +113,114 @@ impl LogCollection {
         }))
     }
 
-    async fn get_logging_resources(
+    async fn pod_logging_resources(
         &self,
-        pods: Vec<Pod>,
+        pod: Pod,
+        nodes_map: HashMap<String, Node>,
```
tiagolobocastro: We can pass by reference?

```suggestion
        nodes_map: &HashMap<String, Node>,
```

tiagolobocastro (approved):

sinhaashish:
bors merge

PR #577, opened 2024-12-04
build: update control-plane dep

tiagolobocastro:
Also updates style script to match control-plane.


niladrih (commented):
Maybe doesn't require the submodule update any longer

PR #581, opened 2024-12-09
feat(chart): add health probes to api-rest

niladrih:


chart/values.yaml
```diff
@@ -254,6 +254,27 @@ apis:
   rest:
     # -- Log level for the rest service
     logLevel: info
+    healthProbes:
```
tiagolobocastro, line 257: Should we add an enable flag?

niladrih: We could. But should we? If the api-rest is not live, a restart should be the expected outcome?
Also, if the api-rest is not ready, it should be removed from the kubernetes service endpoint list?

Is there a reason to not want these? Should these be made mandatory, or should these be good-to-have features.

tiagolobocastro: If there's a bug, or something with these, we may want to disable them?

niladrih: Okay, I'll add toggles..

niladrih: I've added toggles

chart/templates/mayastor/apis/api-rest-deployment.yaml
```diff
@@ -53,6 +53,9 @@ spec:
             - "--core-grpc=https://{{ .Release.Name }}-agent-core:50051"
             - "--ansi-colors={{ .Values.base.logging.color }}"
             - "--fmt-style={{ include "logFormat" . }}"
+            {{- if .Values.apis.rest.healthProbes.readiness.enabled }}
+            - "--core-health-freq={{ .Values.apis.rest.healthProbes.readiness.agentCoreProbeFreq }}"
```
tiagolobocastro, line 57: If it's not specified does it disable the core health freq or does it just not change the freq?

chart/templates/mayastor/apis/api-rest-deployment.yaml
```diff
@@ -63,3 +66,23 @@ spec:
...
+          {{- if .Values.apis.rest.healthProbes.liveness.enabled }}
+          livenessProbe:
+            httpGet:
+              path: /live
+              port: 8081
+            failureThreshold: {{ .Values.apis.rest.healthProbes.liveness.failureThreshold }}
+            initialDelaySeconds: {{ .Values.apis.rest.healthProbes.liveness.initialDelaySeconds }}
+            periodSeconds: {{ .Values.apis.rest.healthProbes.liveness.periodSeconds }}
+            timeoutSeconds: {{ .Values.apis.rest.healthProbes.liveness.timeoutSeconds }}
+          {{ end }}
```
tiagolobocastro:
```suggestion
          {{- end }}
```

tiagolobocastro (approved):

chart/templates/mayastor/apis/api-rest-deployment.yaml
niladrih: Doesn't disable it, it just uses the default frequency of 2m on the binary.
However, it also disable readiness probe, so in effect, it will call the core Node endpoint once on new(), and then never again. Updates to the cache only happen on `get_or_update` calls.

Abhinandan-Purkait (approved):

niladrih:
bors merge

niladrih:
bors merge

PR #592, opened 2024-12-18
feat(helm chart): add value to modify ps retries

Abhinandan-Purkait:
- Adds a helm value to modify the io-engine `ps-retries`.

### DIFF
--- a/chart/templates/mayastor/io/io-engine-daemonset.yaml
+++ b/chart/templates/mayastor/io/io-engine-daemonset.yaml
@@ -111,6 +111,7 @@ spec:
         - "--reactor-freeze-detection"{{ end }}
         - "--tgt-crdt={{ .Values.io_engine.target.nvmf.hostCmdRetryDelay.crdt1 }}"{{ if .Values.eventing.enabled }}
         - "--events-url=nats://{{ .Release.Name }}-nats:4222"{{ end }}
+        - "--ps-retries={{ default 300 .Values.io_engine.pstorRetries }}"
         command:
         - io-engine
         securityContext:
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -466,6 +466,8 @@ io_engine:
   priorityClassName: ""
   # -- Runtime class to use. Defaults to cluster standard
   runtimeClassName: ""
+  # -- Number of retries for pstor persistence before the volume target self shutdowns
+  pstorRetries: 300
 
 etcd:
   # -- Disable when using an external etcd cluster.
--- a/chart/README.md
+++ b/chart/README.md
@@ -176,6 +176,7 @@ This removes all the Kubernetes components associated with the chart and deletes
 | io_engine.&ZeroWidthSpace;nvme.&ZeroWidthSpace;ioTimeout | Timeout for IOs The default here is exaggerated for local disks, but we've observed that in shared virtual environments having a higher timeout value is beneficial. Please adjust this according to your hardware and needs. | `"110s"` |
 | io_engine.&ZeroWidthSpace;nvme.&ZeroWidthSpace;tcp.&ZeroWidthSpace;maxQueueDepth | You may need to increase this for a higher outstanding IOs per volume | `"32"` |
 | io_engine.&ZeroWidthSpace;priorityClassName | Set PriorityClass, overrides global | `""` |
+| io_engine.&ZeroWidthSpace;pstorRetries | Number of retries for pstor persistence before the volume target self shutdowns | `300` |
 | io_engine.&ZeroWidthSpace;resources.&ZeroWidthSpace;limits.&ZeroWidthSpace;cpu | Cpu limits for the io-engine | `""` |
 | io_engine.&ZeroWidthSpace;resources.&ZeroWidthSpace;limits.&ZeroWidthSpace;hugepages2Mi | Hugepage size available on the nodes | `"2Gi"` |
 | io_engine.&ZeroWidthSpace;resources.&ZeroWidthSpace;limits.&ZeroWidthSpace;memory | Memory limits for the io-engine | `"1Gi"` |


chart/templates/mayastor/io/io-engine-daemonset.yaml
```diff
@@ -110,7 +110,8 @@ spec:
         - "--env-context=--{{ .Values.io_engine.envcontext }}"{{ end }}{{ if .Values.io_engine.reactorFreezeDetection.enabled }}
         - "--reactor-freeze-detection"{{ end }}
         - "--tgt-crdt={{ .Values.io_engine.target.nvmf.hostCmdRetryDelay.crdt1 }}"{{ if .Values.eventing.enabled }}
-        - "--events-url=nats://{{ .Release.Name }}-nats:4222"{{ end }}
+        - "--events-url=nats://{{ .Release.Name }}-nats:4222"{{ end }}{{ if .Values.io_engine.ps_retries}}
+        - "--ps-retries=300"{{ end }}
```
tiagolobocastro:
```suggestion
        - "--ps-retries={{ .Values.io_engine.pstorRetries }}"{{ end }}
```

chart/values.yaml
```diff
@@ -466,6 +466,8 @@ io_engine:
   priorityClassName: ""
   # -- Runtime class to use. Defaults to cluster standard
   runtimeClassName: ""
+  # -- Number of retries for pstor persistence before target self shutdown
+  ps_retries: "300"
```
```suggestion
  pstorRetries: "300"
```

chart/values.yaml
```diff
@@ -466,6 +466,8 @@ io_engine:
   priorityClassName: ""
   # -- Runtime class to use. Defaults to cluster standard
   runtimeClassName: ""
+  # -- Number of retries for pstor persistence before target self shutdown
```
```suggestion
  # -- Number of retries for pstor persistence before the volume target self shutdowns
```

chart/templates/mayastor/io/io-engine-daemonset.yaml
Abhinandan-Purkait: Oops. Thanks for pointing this out.

dsharma-dc (approved):

chart/templates/mayastor/io/io-engine-daemonset.yaml
```diff
@@ -110,7 +110,8 @@ spec:
         - "--env-context=--{{ .Values.io_engine.envcontext }}"{{ end }}{{ if .Values.io_engine.reactorFreezeDetection.enabled }}
         - "--reactor-freeze-detection"{{ end }}
         - "--tgt-crdt={{ .Values.io_engine.target.nvmf.hostCmdRetryDelay.crdt1 }}"{{ if .Values.eventing.enabled }}
-        - "--events-url=nats://{{ .Release.Name }}-nats:4222"{{ end }}
+        - "--events-url=nats://{{ .Release.Name }}-nats:4222"{{ end }}{{ if .Values.io_engine.pstorRetries}}
+        - "--ps-retries={{ .Values.io_engine.pstorRetries }}"{{ end }}
```
tiagolobocastro: Should we use default rather than if? For upgrades with old values?

tiagolobocastro (approved):

chart/templates/mayastor/io/io-engine-daemonset.yaml
```diff
@@ -110,7 +110,8 @@ spec:
         - "--env-context=--{{ .Values.io_engine.envcontext }}"{{ end }}{{ if .Values.io_engine.reactorFreezeDetection.enabled }}
         - "--reactor-freeze-detection"{{ end }}
         - "--tgt-crdt={{ .Values.io_engine.target.nvmf.hostCmdRetryDelay.crdt1 }}"{{ if .Values.eventing.enabled }}
-        - "--events-url=nats://{{ .Release.Name }}-nats:4222"{{ end }}
+        - "--events-url=nats://{{ .Release.Name }}-nats:4222"{{ end }}{{ if .Values.io_engine.pstorRetries}}
```
tiagolobocastro:
```suggestion
        - "--events-url=nats://{{ .Release.Name }}-nats:4222"{{ end }}{{ if .Values.io_engine.pstorRetries }}
```

chart/templates/mayastor/io/io-engine-daemonset.yaml
Abhinandan-Purkait: Yeah we can? Should I do that now?

niladrih: This is new value, this should get pulled in. Upgrades via upgrade job should be fine.

chart/values.yaml
```diff
@@ -466,6 +466,8 @@ io_engine:
   priorityClassName: ""
   # -- Runtime class to use. Defaults to cluster standard
   runtimeClassName: ""
+  # -- Number of retries for pstor persistence before the volume target self shutdowns
+  pstorRetries: "300"
```
niladrih: This looks like an int, strange to use it as a string. Unless it makes some logical sense, e.g. the rust data type is a String.

Abhinandan-Purkait: The parser parses it into a u16 anyway

chart/templates/mayastor/io/io-engine-daemonset.yaml
Abhinandan-Purkait: Ohh then we don't need it? I have added it anyway. Should I remove it?

niladrih: It's fine either way. Don't need to change IMO.

chart/values.yaml
niladrih: Yeah, I'd say we should remove the `""` here in that case.

Abhinandan-Purkait: Okay.

niladrih (approved):

chart/values.yaml

Abhinandan-Purkait:
bors merge

niladrih:
> bors merge

Needs conversation resolution. This will likely fail otherwise.

Abhinandan-Purkait:
> > bors merge
> 
> Needs conversation resolution. This will likely fail otherwise.

I think I did resolve before the comment. Did I not?

Abhinandan-Purkait:
bors merge

PR #594, opened 2024-12-30
ci: switch from jenkins to github actions

tiagolobocastro:
Jenkins has become troublesome to maintain and makes it harder for users to see and test output without adding new users.
We could probably spend some time improving that, but the current Jenkins is also setup on systems managed by the sponsoring company so it makes sense to move away from that at least.
In order to run the CI tests we need a relatively powerful CI system so we are making use of the ubuntu-latest-16-cores runners.

The tests seem to run slower than on Jenkins. Part of it may be explained because the dependencies are not pre-cache and also the pre-requisites which take time to install.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors try

niladrih (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #595, opened 2025-01-06
fix(upgrade): remove etcd.initialClusterState from helm values

### LINKED ISSUE #1735: Handle removal of etcd.initialClusterState during helm upgrade in upgrade-job
Ref: https://github.com/openebs/mayastor-extensions/pull/536

The upgrade-job, as of 2.7.0 reuses all of the existing helm values from the source chart. The removal of the parameter from the values.yaml would mean that during upgrade that value has to be forcibly removed when moving to 2.7.0+ versions. This needs to be handled somehow. 

If there is no new design changes to upgrade, a new entry to the [helm::values.rs](https://github.com/openebs/mayastor-extensions/blob/develop/k8s/upgrade/src/bin/upgrade-job/helm/values.rs) module is required which explicitly removes the value from the values set. This might also be required even if helm's `--reset-then-reuse-values` flag is used, as previous upgrade-jobs would have used the `-f values.yaml` option during upgrade and would have explicitly set ALL of the source chart's values (and they would be now visible on `helm get values` output).

niladrih:
Fixes https://github.com/openebs/mayastor/issues/1735

This removes the helm value for the key etcd.initialClusterState for all versions greater than or equal to 2.0.0-rc.0 and less than 2.7.3.

### DIFF
--- a/k8s/upgrade/src/common/constants.rs
+++ b/k8s/upgrade/src/common/constants.rs
@@ -57,3 +57,6 @@ pub(crate) const TWO_DOT_SIX: &str = "2.6.0";
 
 /// Version value for 2.7.2 release.
 pub(crate) const TWO_DOT_SEVEN_DOT_TWO: &str = "2.7.2";
+
+/// Version value for 2.7.3.
+pub(crate) const TWO_DOT_SEVEN_DOT_THREE: &str = "2.7.3";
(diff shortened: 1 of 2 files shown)


Abhinandan-Purkait (approved):

tiagolobocastro (approved):

niladrih:
bors merge

niladrih:
bors merge

niladrih:
bors merge

PR #599, opened 2025-01-12
fix(upgrade-job): explicitly set localpv image tags always

niladrih:
The upgrade-job isn't setting the localpv-provisioner dependency helm chart's openebs/provisioner-localpv image tag when upgrading. Due to the preference for existing configuration, the localpv-provisioner chart is being configured to use the source chart's openebs/provisioner-localpv container image with the target chart's helm templates.

This change adds value sets for openebs/provisioner-localpv image tag and the openebs/linux-utils image tag.

This also removes `localpv-provisioner.deviceClass`, `localpv-provisioner.localpv.waitForBDBindTimeoutRetryCount` and `localpv-provisioner.openebsNDM` helm values keys when upgrading from Mayastor <2.6.0's LocalPV 3.x to Mayastor 2.8's LocalPV v4. These keys were removed from the LocalPV v4 chart. This cleans up the values yaml a bit. Because, these values are no longer used, removing them doesn't affect any behavior.

### DIFF
--- a/k8s/upgrade/src/helm/chart.rs
+++ b/k8s/upgrade/src/helm/chart.rs
@@ -259,11 +259,6 @@ impl CoreValues {
         self.loki_stack.grafana_sidecar_image_tag()
     }
 
-    /// This is a getter for the localpv-provisioner sub-chart's release version.
-    pub(crate) fn localpv_release_version(&self) -> &str {
-        self.localpv_provisioner.release_version()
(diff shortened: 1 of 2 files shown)


abhilashshetty04 (approved):

dsharma-dc (approved):

niladrih:
bors merge

PR #602, opened 2025-01-15
fix: account for builtin nvme-tcp module in initContainer

### LINKED ISSUE #1704: InitContainers - Handle builtin kernel modules
**Describe the bug**
Some components, by default, use init containers to check for nvme kernel modules.
The problem is these modules might actually get compiled builtin, and these checks are failing in this case.

**To Reproduce**
Install on nodes with builtin nvme-tcp kernel modules.

**Expected behavior**
InitContainer should not fail.

** OS info (please complete the following information):**
 - Distro: Talos
Example: https://discuss.kubernetes.io/t/talos-linux-cluster-missing-module/27519

dsharma-dc:
- If sysfs check pass - driver is available as either loaded already or builtin
 - If sysfs check fails - driver isn't builtin but can be possibly loaded
 - For nvme_rdma, a separate change will be done as that is dependent on helm option nvmf.target.rdma.enable

Fixes https://github.com/openebs/mayastor/issues/1704

## Description
Some components, by default, use init containers to check for nvme kernel modules.
The problem is these modules might actually get compiled builtin, and these checks are failing in this case.
This change accounts for the nvme-tcp module being builtin.

## How Has This Been Tested?

- Removed module, checking initContainer logs.
- Insert module again, check initContainer logs
- **### Yet to test on Talos** 

## Types of changes

## Checklist:

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -364,7 +364,7 @@ csi:
       enabled: false
       containers:
         - name: nvme-tcp-probe
-          command: ['sh', '-c', 'trap "exit 1" TERM; until $(lsmod | grep nvme_tcp &>/dev/null); do [ -z "$WARNED" ] && echo "nvme_tcp module not loaded..."; WARNED=1; sleep 60; done;']
+          command: ['sh', '-c', 'trap "exit 1" TERM; until [ -d /sys/module/nvme_tcp ]; do [ -z "$WARNED" ] && echo "nvme_tcp module not loaded..."; WARNED=1; sleep 60; done;']
 
 io_engine:
   # -- Log level for the io-engine service
--- a/scripts/k8s/setup-io-prereq.sh
+++ b/scripts/k8s/setup-io-prereq.sh
@@ -114,7 +114,7 @@ if [ -n "$HUGE_PAGES" ]; then
 fi
 
 if [ -n "$NVME_TCP" ]; then
-  if ! lsmod | grep "nvme_tcp" >/dev/null; then
+  if ! [ -d "/sys/module/nvme_tcp" ] >/dev/null; then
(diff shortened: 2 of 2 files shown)


chart/values.yaml
```diff
@@ -364,7 +364,7 @@ csi:
       enabled: false
       containers:
         - name: nvme-tcp-probe
-          command: ['sh', '-c', 'trap "exit 1" TERM; until $(lsmod | grep nvme_tcp &>/dev/null); do [ -z "$WARNED" ] && echo "nvme_tcp module not loaded..."; WARNED=1; sleep 60; done;']
+          command: ['sh', '-c', 'trap "exit 1" TERM; if [ -d /sys/module/nvme_tcp ]; then echo "nvme_tcp available."; exit 0; fi; until $(lsmod | grep nvme_tcp &>/dev/null); do [ -z "$WARNED" ] && echo "nvme_tcp module not loaded..."; WARNED=1; sleep 60; done;']
```
niladrih:
```suggestion
          command: ['sh', '-c', 'trap "exit 1" TERM; until [ -d /sys/module/nvme_tcp ]; do [ -z "$WARNED" ] && echo "nvme_tcp module not loaded..."; WARNED=1; sleep 60; done;']
```

dsharma-dc: done

tiagolobocastro (approved):

tiagolobocastro (commented):
Could you also fix the same on setup-io-prereq.sh#L123 ?

niladrih (approved):

dsharma-dc:
> Could you also fix the same on setup-io-prereq.sh#L123 ?

Done that.

dsharma-dc:
bors merge

PR #608, opened 2025-01-23
feat(plugin): add support to fetch ns from kubeconfig context

abhilashshetty04:
If '---ns_from_context' flag is set then we fetch ns from the kubeconfig context, otherwise, --namespace arg is used. '---ns_from_context' takes precedence over -n if both are used.

### DIFF
--- a/k8s/plugin/src/main.rs
+++ b/k8s/plugin/src/main.rs
@@ -1,10 +1,9 @@
 use clap::Parser;
 use plugin::ExecuteOperation;
 use resources::{init_rest, Error, Operations};
+pub mod resources;
 
 use std::{env, ops::Deref};
-
-pub mod resources;
 
 #[derive(Parser, Debug)]
 #[clap(name = utils::package_description!(), version = utils::version_info_str!())]
@@ -14,19 +13,32 @@ struct CliArgs {
     #[clap(subcommand)]
     operations: Operations,
 
-    /// Kubernetes namespace of mayastor service
-    #[clap(global = true, long, short = 'n', default_value = "mayastor")]
-    namespace: String,
+    /// Kubernetes namespace of mayastor service [default: mayastor].
+    #[clap(global = true, long, short = 'n')]
+    namespace: Option<String>,
 
     #[clap(flatten)]
     args: resources::CliArgs,
+
+    /// Use namespace from the current `kubeconfig` context.
+    #[clap(global = true, long, hide = true, default_value = "false")]
+    namespace_from_context: bool,
 }
 
 impl CliArgs {
-    fn args() -> Self {
+    async fn args() -> Result<Self, anyhow::Error> {
         let mut args = CliArgs::parse();
-        args.args.namespace = args.namespace.clone();
-        args
+        args.args.namespace = if let Some(namespace) = &args.namespace {
+            namespace.to_string()
+        } else if args.namespace_from_context {
+            let client = kube_proxy::client_from_kubeconfig(args.args.kube_config_path.clone())
+                .await
+                .map_err(|err| anyhow::anyhow!("{err}"))?;
+            client.default_namespace().to_string()
+        } else {
+            constants::DEFAULT_PLUGIN_NAMESPACE.to_string()
+        };
+        Ok(args)
     }
 }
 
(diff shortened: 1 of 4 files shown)


k8s/plugin/src/main.rs
```diff
@@ -20,12 +20,22 @@ struct CliArgs {
...
+    ns_from_context: bool,
 }
 
 impl CliArgs {
-    fn args() -> Self {
+    async fn args() -> Self {
         let mut args = CliArgs::parse();
-        args.args.namespace = args.namespace.clone();
+        args.args.namespace = if args.ns_from_context {
+            let client = Client::try_default().await.expect("Client init failed");
```
tiagolobocastro: hmm this is not taking into account the kubeconfig path

k8s/plugin/src/main.rs
```diff
@@ -20,12 +20,22 @@ struct CliArgs {
 
     #[clap(flatten)]
     args: resources::CliArgs,
+
+    /// Use namespace from the current `kubeconfig` context,
+    /// Takes precedence over --namespace or -n flag.
+    #[clap(global = true, long, short = 'c', default_value = "false")]
```
tiagolobocastro: We should hide this, since it's not meant to be used for now IIRC

```suggestion
    #[clap(global = true, hide = true, long = "namespace-from-context", default_value = "false")]
```

tiagolobocastro: Also should make it conflict with --namespace? Both supplied at the same time doesn't make sense

abhilashshetty04: I suppose user should not supply both at a same time. However, if they do then we give precedence to config file..We had to make it conflict as both will set ars.args.namespace otherwise

This flag anyway we can remove on the next major release. We wanted to remove namespace getting set to mayastor by default right?

k8s/plugin/src/main.rs
```diff
@@ -15,17 +15,36 @@ struct CliArgs {
...
+            namespace.to_string()
+        } else if args.namespace_from_context {
+            let client = if let Some(path) = &args.args.kube_config_path {
+                kube_proxy::client_from_kubeconfig(Some(path.clone()))
+                    .await
+                    .expect("Cant create client from supplied kubeconfig")
+            } else {
+                Client::try_default()
+                    .await
+                    .expect("Cant create client from default kubeconfig")
```
niladrih:
```suggestion
                    .expect("Couldn't create client from supplied kubeconfig")
            } else {
                Client::try_default()
                    .await
                    .expect("Couldn't create client from default kubeconfig")
```

niladrih (approved):

k8s/plugin/src/main.rs
```diff
@@ -15,17 +15,36 @@ struct CliArgs {
...
+    async fn args() -> Self {
         let mut args = CliArgs::parse();
-        args.args.namespace = args.namespace.clone();
+        args.args.namespace = if let Some(namespace) = &args.namespace {
+            namespace.to_string()
+        } else if args.namespace_from_context {
+            let client = if let Some(path) = &args.args.kube_config_path {
+                kube_proxy::client_from_kubeconfig(Some(path.clone()))
+                    .await
+                    .expect("Couldnt create client from supplied kubeconfig")
```
tiagolobocastro: Let's return an error here rather than panic

k8s/plugin/src/main.rs
```diff
@@ -15,17 +15,36 @@ struct CliArgs {
     operations: Operations,
 
     /// Kubernetes namespace of mayastor service
-    #[clap(global = true, long, short = 'n', default_value = "mayastor")]
-    namespace: String,
+    #[clap(global = true, long, short = 'n')]
+    namespace: Option<String>,
```
tiagolobocastro, line 18: Not sure about this, because this now modifies the help message and becomes unclear that mayastor is still the default namespace?

abhilashshetty04: Modified the docs now of -n flag.

k8s/plugin/src/main.rs
abhilashshetty04: Done

k8s/plugin/src/main.rs
```diff
@@ -14,19 +15,40 @@ struct CliArgs {
...
+                kube_proxy::client_from_kubeconfig(Some(path.clone()))
+                    .await
+                    .map_err(|_| {
+                        anyhow::anyhow!("Couldnt create client from supplied kubeconfig")
+                    })?
+            } else {
+                Client::try_default()
+                    .await
+                    .map_err(|_| anyhow::anyhow!("Couldnt create client from default kubeconfig"))?
+            };
```
tiagolobocastro: Can't we just do this?

```suggestion
            let client = kube_proxy::client_from_kubeconfig(args.args.kube_config_path.clone())
                    .await
                    .map_err(|_| {
                        anyhow::anyhow!("Couldn't create client from supplied kubeconfig")
                    })?;
```

k8s/plugin/src/main.rs
```diff
@@ -40,23 +62,27 @@ impl Deref for CliArgs {
...
+                        exit_code = error.into();
+                    }
+                    Error::Generic(error) => eprintln!("{error}"),
+                }
+                std::process::exit(exit_code);
             }
-            Error::Generic(error) => eprintln!("{error}"),
         }
-        std::process::exit(exit_code);
+        Err(e) => eprintln!("{e}"),
```
tiagolobocastro: This needs to be flattened, it's not erroring out with -1

k8s/plugin/src/main.rs
```diff
@@ -1,11 +1,12 @@
+//use anyhow::anyhow;
```
Abhinandan-Purkait: Remove this?

k8s/plugin/src/main.rs
```diff
@@ -14,19 +15,40 @@ struct CliArgs {
...
+                        anyhow::anyhow!("Couldnt create client from supplied kubeconfig")
+                    })?
+            } else {
+                Client::try_default()
+                    .await
+                    .map_err(|_| anyhow::anyhow!("Couldnt create client from default kubeconfig"))?
+            };
+            client.default_namespace().to_string()
+        } else {
+            "mayastor".to_string()
```
Abhinandan-Purkait: Should come from a constant.

k8s/plugin/src/main.rs
abhilashshetty04: Yes should work.

k8s/plugin/src/main.rs
abhilashshetty04: I have added std::process::exit to error arm

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

abhilashshetty04:
bors merge

PR #614, opened 2025-01-31
Upgrade change for builtin-nvme_tcp-module checking initContainer

niladrih:


k8s/upgrade/src/helm/chart.rs
```diff
@@ -447,6 +465,21 @@ impl GenericContainers {
     }
 }
 
+/// This is used to deserialize various 'initContainer' yaml objects.
+#[derive(Default, Deserialize)]
+#[serde(rename_all(deserialize = "camelCase"))]
+struct GenericInitContainer {
+    /// InitContainer config.
+    init_container: Vec<Container>,
```
dsharma-dc, line 473:
```suggestion
    init_containers: Vec<Container>,
```

k8s/upgrade/src/helm/values.rs
```diff
@@ -512,16 +498,99 @@ where
...
+        let command_to_replace = String::from(
+            r#"trap "exit 1" TERM; until $(lsmod | grep nvme_tcp &>/dev/null); do [ -z "$WARNED" ] && echo "nvme_tcp module not loaded..."; WARNED=1; sleep 60; done;"#,
+        );
+        if source_values
+            .csi_node_init_containers()
+            .iter()
+            .filter_map(|container| container.command.clone())
+            .flat_map(|commands| commands.into_iter())
+            .collect::<Vec<_>>()
+            .contains(&command_to_replace)
```
dsharma-dc: Shall we simply filter by the init container's name instead of the command, if that can work? Will be cleaner and less prone to errors if command changes in older versions.

Abhinandan-Purkait (approved):

k8s/upgrade/src/helm/values.rs
niladrih: Sure, that would work in this case. 👍

k8s/upgrade/src/helm/chart.rs
niladrih: It's InitContainer, e.g.: https://github.com/openebs/mayastor-extensions/blob/develop/chart/values.yaml#L107

k8s/upgrade/src/helm/values.rs
niladrih: Done.

dsharma-dc (approved):

niladrih:
bors merge

PR #616, opened 2025-02-01
feat(chart): add eviction tolerations to localpv,dsp operator,csi controller

niladrih:
Changes:
- Add eviction tolerations to the dsp operator deployment, csi-controller
- Update localpv provisioner chart to 4.2
- rename `tolerations_with_early_eviction` to `_tolerations_with_early_eviction_two` to avoid conflict with the localpv-provisioner _helpers.tpl function with the same name

### DIFF
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -182,9 +182,9 @@ Usage:
 {{/*
 Creates the tolerations based on the global and component wise tolerations, with early eviction
 Usage:
-{{ include "tolerations_with_early_eviction" (dict "template" . "localTolerations" .Values.path.to.local.tolerations) }}
+{{ include "_tolerations_with_early_eviction" (dict "template" . "localTolerations" .Values.path.to.local.tolerations) }}
 */}}
-{{- define "tolerations_with_early_eviction" -}}
+{{- define "_tolerations_with_early_eviction" -}}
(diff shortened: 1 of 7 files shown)


dsharma-dc (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

chart/templates/_helpers.tpl
```diff
@@ -182,9 +182,9 @@ Usage:
 {{/*
 Creates the tolerations based on the global and component wise tolerations, with early eviction
 Usage:
-{{ include "tolerations_with_early_eviction" (dict "template" . "localTolerations" .Values.path.to.local.tolerations) }}
+{{ include "_tolerations_with_early_eviction" (dict "template" . "localTolerations" .Values.path.to.local.tolerations) }}
```
tiagolobocastro, line 185: Why has this been renamed to have an _ ?

niladrih: The helm template command fails because the localpv template tries to call the mayastor helpers function. This is because the two helper functions in the two charts are named the same. 

The templating must be pooling the helpers functions together

tiagolobocastro: Ah I see, yes it works like that..
A comment explaining this would have been nice

PR #623, opened 2025-02-06
Cherry-pick PR 549

tiagolobocastro:
Adds a list of container images which are used by the chart.
This list is contained in file: charts/images.txt

Sometimes the images can't be figured out using templates, and in such
cases we need to manually add them.
In some cases we might be able to do this using helm variables, which
is what we're doing with the localpv provisioner.

Github actions have been modified to verify the contents of the images
as well as testing them with a live cluster.
This is a pretty dumb test that just checks if the live pod images are
contained in the images files.


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #633, opened 2025-02-19
fix(gha): handle multiple workflows

tiagolobocastro:
If for some reason a binary workflow is re-run we end up with multiple results.
In this case, any should be "correct" because we're matching the sha. Let's pick one, ex: the last updated one.


Abhinandan-Purkait (approved):

dsharma-dc (approved):

tiagolobocastro:
bors merge

PR #648, opened 2025-03-17
refactor(upgrade): make paginated_list error generic

niladrih:
The APIs `upgrade::common::kube::client::paginated_list` and `upgrade::common::kube::client::paginated_list_metadata` accept inputs for the Snafu error context they would return in the case of an error. This API becomes difficult to use when the Snafu error context isn't defined in this crate itself.

This change removes the input for the tailored error context, and depends on the error type from the 'kube' crate to describe the error context in more detail.

### DIFF
--- a/k8s/upgrade/src/common/error.rs
+++ b/k8s/upgrade/src/common/error.rs
@@ -215,62 +215,6 @@ pub enum Error {
         pod_namespace: String,
     },
 
-    /// Error for when a Kubernetes API request for GET-ing a list of Pods filtered by label(s)
-    /// and field(s) fails.
-    #[snafu(display(
-        "Failed to list Pods with label '{label}', and field '{field}' in namespace '{namespace}': {source}",
-    ))]
-    ListPodsWithLabelAndField {
-        source: kube::Error,
-        label: String,
-        field: String,
-        namespace: String,
-    },
-
-    /// Error for when listing Kubernetes Secrets from a the kubeapi fails.
-    #[snafu(display(
-        "Failed to list Secrets with label '{label}', and field '{field}' in namespace '{namespace}': {source}",
-    ))]
-    ListSecretsWithLabelAndField {
-        source: kube::Error,
-        label: String,
-        field: String,
-        namespace: String,
-    },
-
-    #[snafu(display(
-        "Failed to list ConfigMaps with label '{label}', and field '{field}' in namespace '{namespace}': {source}",
-    ))]
-    ListConfigMapsWithLabelAndField {
-        source: kube::Error,
-        label: String,
-        field: String,
-        namespace: String,
-    },
-
-    /// Error for when a Kubernetes API request for GET-ing a list of ControllerRevisions
-    /// filtered by label(s) and field(s) fails.
-    #[snafu(display(
-        "Failed to list ControllerRevisions with label '{label}', and field '{field}' in Namespace '{namespace}': {source}",
-    ))]
-    ListCtrlRevsWithLabelAndField {
-        source: kube::Error,
-        label: String,
-        field: String,
-        namespace: String,
-    },
-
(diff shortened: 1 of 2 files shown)


niladrih:
bors try

k8s/upgrade/src/common/error.rs
```diff
@@ -727,6 +727,9 @@ pub enum Error {
     /// entry for the CORE_CHART version.
     #[snafu(display("Helm release data doesn't have chart version or contains an invalid version for dependency chart '{CORE_CHART_NAME}'"))]
     InvalidDependencyVersionInHelmReleaseData,
+
+    #[snafu(display("Error: {}", source))]
```
tiagolobocastro:
```suggestion
    #[snafu(display("{source}"))]
```

k8s/upgrade/src/common/kube/client.rs
```diff
@@ -328,11 +336,11 @@ pub async fn paginated_list<K, C, E2>(
     resources: &mut Vec<K>,
     list_params: Option<ListParams>,
     list_err_ctx: C,
-) -> Result<()>
+) -> std::result::Result<(), Box<dyn std::error::Error>>
```
tiagolobocastro: I'm confused as to why this is necessary, isn't `Result::Err` here already a `std::error::Error`?

niladrih: `list_err_ctx` is the Snafu context to return for the failure for this particular list.

tiagolobocastro: This seems like a complicated api for a very simple operation.
I suggest remove `list_err_ctx` completely from here, and let the caller do the mapping

Something like this, change:
```rust
let list_pods_error_ctx = ListPodsWithLabelAndField {
        label: label_selector.unwrap_or_default(),
        field: field_selector.unwrap_or_default(),
        namespace: namespace.clone(),
    };

    paginated_list(pods_api, &mut pods, Some(list_params), list_pods_error_ctx)
        .await
        .context(Other)?;
```
to:
```
let list_pods_error_ctx = ListPodsWithLabelAndField {
        label: label_selector.unwrap_or_default(),
        field: field_selector.unwrap_or_default(),
        namespace: namespace.clone(),
    };

    paginated_list(pods_api, &mut pods, Some(list_params))
        .await
        .context(list_pods_error_ctx)?;
```

niladrih: Yes, that makes sense.

niladrih: Done.

k8s/upgrade/src/common/error.rs
```diff
@@ -677,7 +677,7 @@ pub enum Error {
     },
 
     #[snafu(display("failed to list CustomResourceDefinitions: {source}"))]
-    ListCrds { source: kube::Error },
+    ListCrds { source: Box<Error> },
```
tiagolobocastro: Do these still need to be box?

niladrih: The actual error type will depend on the Enum variant.

tiagolobocastro: Oh this Error is Self here? So you want to stack up these errors?
What's the use case for this?
If we really need this, I would say do this for the ones which really require this and not throughout. We had this kind of thing in the data-plane and this becomes very complicated to match against.

niladrih: I've removed the `Box<Error>`s. The underlying `kube::Error` should be able to describe the issue, may be unnecessary to wrap it with explicit Snafu variants.

tiagolobocastro (approved):

niladrih:
bors merge

Abhinandan-Purkait (approved):

niladrih:
Bors merge

PR #656, opened 2025-04-11
fix: add knob for volume health

tiagolobocastro:
Volume health information was missing from the public api, leading into showing incorrect status for volumes.
In case this affects any existing scripts/tests, we add a knob to disable this information.
This is disabled on pre-2.9 by default to keep existing behavior.


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #658, opened 2025-04-14
build(nix-shell): remove linux specific utils from aarch-darwin

tiagolobocastro:
Allows us to run the nix-shell on darwin, simply in a way to build the plugin and any non-linux dependent binaries.
The nvme-cli and kmod was used for testing on kind clusters, which is assuming that the host is linux.


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #662, opened 2025-04-25
Update cargo deps and submodule deps

niladrih:
Changes:
- Update cargo deps to versions to those in the 'develop' branch. This is so that the code in openebs/openebs may compile correctly. Currently two different versions of kube-rs/kube and/or Arnavion/k8s-openapi doesn't compile correctly.
- Update submodule dependency

### DIFF
--- a/call-home/Cargo.toml
+++ b/call-home/Cargo.toml
@@ -20,8 +20,8 @@ path = "src/bin/stats/main.rs"
 [dependencies]
 constants = { path = "../constants" }
 openapi = { path = "../dependencies/control-plane/openapi" }
-kube = { version = "0.87.0", features = ["runtime", "derive"] }
-k8s-openapi = { version = "0.20.0", features = ["v1_22"] }
+kube = { version = "0.94.2", features = ["runtime", "derive"] }
+k8s-openapi = { version = "0.22.0", features = ["v1_24"] }
 futures = "0.3.31"
 tokio = { version = "1.41.0", features = ["full"] }
 clap = { version = "4.5.20", features = ["cargo", "derive", "string"] }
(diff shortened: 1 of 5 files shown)


Abhinandan-Purkait (approved):

call-home/Cargo.toml
```diff
@@ -20,8 +20,8 @@ path = "src/bin/stats/main.rs"
 [dependencies]
 constants = { path = "../constants" }
 openapi = { path = "../dependencies/control-plane/openapi" }
-kube = { version = "0.87.0", features = ["runtime", "derive"] }
-k8s-openapi = { version = "0.20.0", features = ["v1_22"] }
+kube = { version = "0.94.2", features = ["runtime", "derive"] }
```
abhilashshetty04, line 23: This will need changes on openebs/openebs repo also right? We use this dependency there also.

abhilashshetty04 (approved):

call-home/Cargo.toml
abhilashshetty04: We might not use call-home there. So should be okay i guess.

k8s/upgrade/Cargo.toml
```diff
@@ -17,16 +17,11 @@ utils = { path = "../../dependencies/control-plane/utils/utils-lib" }
...
-anyhow = "1.0.92"
+kube = { version = "0.94.2", default-features = true, features = ["derive", "runtime"] }
 clap = { version = "4.5.20", features = ["derive", "env", "string", "color"] }
 humantime = "2.1.0"
 maplit = "1.0.2"
-k8s-openapi = { version = "0.20.0", features = ["v1_22"] }
-tower = { version = "0.5.1", features = ["timeout", "util"] }
-hyper = { version = "1.5.0", features = ["client", "http1", "http2"] }
-http = "1.1.0"
+k8s-openapi = { version = "0.22.0", features = ["v1_24"] }
```
abhilashshetty04, line 24: Plugin in openebs/openebs uses upgrade as dep.

niladrih: Yes, that's why I'm doing this update here.

call-home/Cargo.toml
niladrih: Yes, we use this as a dependency. The outdated dependencies cause compilation issues.

niladrih:
bors merge

PR #663, opened 2025-04-25
chore: update loki chart, deprecate promtail, use alloy

Abhinandan-Purkait:
### Changes

1. Updates loki to singleBinary monolithic chart.
2. Adds alloy as new dependency.
3. Removes promtail dependency.
4. Adds configuration for alloy.
   1. We can now defined multiple label selectors for scraping.
   2. We can now define tenant id, to segregate logs, based on tenant.
5. Updates the storage configs for the newer components.


k8s/upgrade/src/common/error.rs
niladrih, line 1: Let's not change these just yet

k8s/upgrade/src/helm/chart.rs
niladrih, line 1: Let's not change these just yet

k8s/upgrade/src/helm/values.rs
niladrih, line 1: Let's not change these just yet

k8s/upgrade/src/common/error.rs
Abhinandan-Purkait: Done. Removed them.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

chart/values.yaml
```diff
@@ -5,7 +5,6 @@ crds:
     volumeSnapshots:
       # -- Install Volume Snapshot CRDs
       enabled: true
-
```
niladrih: Let's keep these blank lines..

niladrih (approved):

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

abhilashshetty04 (approved):

Abhinandan-Purkait:
bors merge

chart/values.yaml
tiagolobocastro, line 1: why did you remove the new lines between different components? It's harder to read now

scripts/helm/install.sh
```diff
@@ -165,7 +165,10 @@ else
        --set="etcd.livenessProbe.initialDelaySeconds=5,etcd.readinessProbe.initialDelaySeconds=5,etcd.replicaCount=1" \
        --set="obs.callhome.enabled=true,obs.callhome.sendReport=false,localpv-provisioner.analytics.enabled=false" \
        --set="eventing.enabled=true,nats.cluster.enabled=false,nats.cluster.replicas=1" \
+       --set="loki.loki.commonConfig.replication_factor=2,loki.singleBinary.replicas=2,loki.minio.replicas=2" \
+       --set="loki-stack.enabled=false" \
```
tiagolobocastro, line 169: loki-stack doesn't exist anymore?

scripts/helm/install.sh
```diff
@@ -165,7 +165,10 @@ else
        --set="etcd.livenessProbe.initialDelaySeconds=5,etcd.readinessProbe.initialDelaySeconds=5,etcd.replicaCount=1" \
        --set="obs.callhome.enabled=true,obs.callhome.sendReport=false,localpv-provisioner.analytics.enabled=false" \
        --set="eventing.enabled=true,nats.cluster.enabled=false,nats.cluster.replicas=1" \
+       --set="loki.loki.commonConfig.replication_factor=2,loki.singleBinary.replicas=2,loki.minio.replicas=2" \
```
tiagolobocastro, line 168: Can we set the simplest loki config here?

Abhinandan-Purkait: Yes we can, we can use single replica without minio. That would actually make install a bit faster. Should I do it?

chart/values.yaml
Abhinandan-Purkait: I don't remember removing it intentionally. Will revert this.

scripts/helm/install.sh
Abhinandan-Purkait: This is needed till the upgrade test is fixed.

Abhinandan-Purkait: This will disable loki-stack on the old version, thus on upgrade there will be no conflict.

Abhinandan-Purkait: That's the reason I have added the comment below.

scripts/helm/install.sh
tiagolobocastro: Yes we should do that here, since this script is meant for faster testing and not e2e (etcd and nats are already set to 1)

PR #664, opened 2025-04-30
refactor: make the EventRecorder extensible for openebs/openebs upgrade

niladrih:
When testing OpenEBS upgrade (for the openebs/openebs helm chart), it appears that the upgrade-job cites its Events' reason as 'MayastorUpgrade'. Seems to be related to this name being hard-coded in the implementation.

### DIFF
--- a/k8s/upgrade/src/events/event_recorder.rs
+++ b/k8s/upgrade/src/events/event_recorder.rs
@@ -1,5 +1,4 @@
 use crate::common::{
-    constants::product_train,
     error::{
         EventChannelSend, EventPublish, EventRecorderOptionsAbsent, GetPod, JobPodHasTooManyOwners,
         JobPodOwnerIsNotJob, JobPodOwnerNotFound, Result, SerializeEventNote,
@@ -49,6 +48,7 @@ pub struct EventRecorderBuilder {
     namespace: Option<String>,
     source_version: Option<String>,
     target_version: Option<String>,
+    product_name: String,
 }
 
 impl EventRecorderBuilder {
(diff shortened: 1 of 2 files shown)


niladrih:
bors try

k8s/upgrade/src/events/event_recorder.rs
```diff
@@ -44,14 +43,15 @@ impl EventNote {
 
 /// A builder for the Kubernetes event publisher.
 #[derive(Default)]
-pub struct EventRecorderBuilder {
+pub struct EventRecorderBuilder<'a> {
     pod_name: Option<String>,
     namespace: Option<String>,
     source_version: Option<String>,
     target_version: Option<String>,
+    product_name: &'a str,
```
tiagolobocastro: Just make it a string like everything, no point complicating the api just for this when everything else is already String

k8s/upgrade/src/events/event_recorder.rs
```diff
@@ -170,21 +177,29 @@ impl EventRecorderBuilder {
...
             source_version,
             target_version,
+            product_name: self.product_name,
         })
     }
 }
 
 /// This is a wrapper around a kube::runtime::events::Recorder.
-pub struct EventRecorder {
+pub struct EventRecorder<'a> {
```
tiagolobocastro: Are all the other fields already "settable" ?

tiagolobocastro (approved):

k8s/upgrade/src/events/event_recorder.rs
niladrih: No, only the source-version and the target-version are settable. This is essentially due to the API not being designed well. The rest of the members' values are available at startup, so they are populated via the builder. The source-version and the target-version are picked up at run-time, so they are populated later.

k8s/upgrade/src/events/event_recorder.rs
niladrih: Alright.

niladrih:
bors merge

dsharma-dc (approved):

niladrih:
bors merge

PR #672, opened 2025-05-15
Etcd dependency chart upgrade to v12.0.11

niladrih:


niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors cancel

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
~bors merge~

niladrih:
bors try

niladrih:
bors try

niladrih:
bors cancel

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
Etcd upgrade invokes an etcd PreUpgrade Kubernetes Job which carries the same Affinity rules as the Etcd StatefulSet, so it tries to schedule on a node where no Etcd StatefulSet Pod exists. The affinity rules aren't configurable via the helm chart values. Also, The Affinity section of the PodSpec is immutable via the kubernetes api server. The issue here lies in upgrade timing out when there are exactly the same number of worker nodes as the Job Pod is unable to get itself scheduled.
What could we do about this?
- Can't patch the affinity rules out, they're immutable
- Should we disable the PreUpgrade Job and add a copy of the PreUpgrade Job ourselves (to the mayastor helm chart), without Affinity rules?
- Should we scale down the Etcd StatefulSet, if it has more than 1 replicas --> perform upgrade --> scale Etcd back up; scale Etcd back up in case of upgrade failure, before exit?

> Should we disable the PreUpgrade Job and add a copy of the PreUpgrade Job ourselves (to the mayastor helm chart), without Affinity rules?

I prefer this approach

niladrih:
bors try

niladrih:
bors try

niladrih:
bors cancel

niladrih:
bors try

niladrih:
Adding etcd preUpgradeJob affinity configuration knobs to the bitnami/etcd chart: https://github.com/bitnami/charts/pull/34367

niladrih:
bors try

niladrih:
bors cancel

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

chart/values.yaml
```diff
@@ -549,27 +555,26 @@ etcd:
       repository: openebs/alpine-bash
       tag: 4.2.0
       pullSecrets: []
-  # extra debug information on logs
-  debug: false
-  # -- Pod anti-affinity preset
-  # Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity
-  podAntiAffinityPreset: "hard"
```
tiagolobocastro, line 568: Is this not needed now?

chart/values.yaml
```diff
@@ -514,8 +522,11 @@ etcd:
     storageClass: "mayastor-etcd-localpv"
     # -- Volume size
     size: 2Gi
+  persistentVolumeClaimRetentionPolicy:
     # -- PVC's reclaimPolicy
-    reclaimPolicy: "Delete"
+    enabled: true
+    whenDeleted: Delete
```
tiagolobocastro: hmm won't this mean we'll delete the pvc during the upgrade step?

k8s/upgrade/src/common/kube/client.rs
```diff
@@ -66,27 +69,74 @@ pub async fn configmaps_api(namespace: &str) -> Result<Api<ConfigMap>> {
...
-            release_name,
-            namespace,
+        .context(FailedToDeleteStatefulSet {
+            namespace: namespace.to_string(),
         })
 }
 
+/// This deletes the etcd statefulset as per the instructions
+/// here: https://github.com/bitnami/charts/tree/etcd/9.0.0/bitnami/etcd#to-900
+pub async fn delete_etcd_sts(release_name: String, namespace: String) -> Result<()> {
```
tiagolobocastro: so this means upgrade without the plugin won't work? or at least without performing this step manually?

tiagolobocastro: and if so can we do this as part of a pre-upgrade job?

k8s/upgrade/src/helm/values.rs
```diff
@@ -89,7 +90,7 @@ where
         if source_values.io_engine_log_level().eq(log_level_to_replace)
             && target_values.io_engine_log_level().ne(log_level_to_replace)
         {
-            yq.set_literal_value(
+            yq.set_quoted_string_value(
```
tiagolobocastro, line 93: what are these changes doing here?

niladrih: The module upgrade::yaml::yq has functions to set values to the temporary file we use as the values.yaml file for helm upgrade. This is renaming these functions so that the caller can now choose if they want `""` around their values or not. Previously, special keywords like `true`, `false`, `null` and integer values were set without quotes. In this case there was a value which was changed from an integer to a string (`etcd.autoCompactionRetention`). Automatically detecting if the value required `""` around it wasn't helping in this case. So split the API into two different functions, one for with-quotes, one without.

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

chart/values.yaml
niladrih: Making a change upstream to accommodate the 'hard' podAntiAffinityPreset: https://github.com/bitnami/charts/pull/35003

niladrih:
bors try

k8s/upgrade/src/common/kube/client.rs
niladrih: I've moved the etcd delete and label steps to a pre-upgrade-hook Job.

chart/values.yaml
niladrih: Set these to Retain.

chart/values.yaml
niladrih: Removing this is no longer needed.

niladrih:
bors try

niladrih:
bors try

chart/label-etcd-for-helm-release.sh
```diff
@@ -0,0 +1,312 @@
...
+
+# This step does these few things..
+#  - Add 'app.kubernetes.io/component: etcd' to the StatefulSet's .metadata.labels
+#  - Add 'app.kubernetes.io/component: etcd' to the StatefulSet's .spec.selector
+#  - Add 'app.kubernetes.io/component: etcd' to the StatefulSet's podAntiAffinity's matchLabels
+#  - Change the value of ETCD_INITIAL_CLUSTER_STATE to 'existing'
+modified_etcd_yaml="$(echo "$EXISTING_ETCD_STS" | \
+awk -v label="app.kubernetes.io/component: etcd" \
+    -v init_state='value: "existing"' '
+  # As soon as we hit "status:", stop processing.
```
tiagolobocastro, line 272: wow, this is a very complicated operation

chart/label-etcd-for-helm-release.sh
```diff
@@ -0,0 +1,312 @@
+#!/usr/bin/env sh
```
tiagolobocastro, line 1: Do you really want sh rather than bash?
Please run the script via shellcheck, it's a got a few warnings

chart/values.yaml
```diff
@@ -6,6 +6,10 @@ crds:
       # -- Install Volume Snapshot CRDs
       enabled: true
 
+global:
+  security:
+    allowInsecureImages: true
```
tiagolobocastro, line 11: what does this mean?

chart/values.yaml
```diff
@@ -73,7 +77,7 @@ base:
       - name: agent-core-grpc-probe
         command: ['sh', '-c', 'trap "exit 1" TERM; until nc -vzw 5 {{ .Release.Name }}-agent-core 50051; do date; echo "Waiting for agent-core-grpc services..."; sleep 1; done;']
       - name: etcd-probe
-        command: ['sh', '-c', 'trap "exit 1" TERM; until nc -vzw 5 {{ .Release.Name }}-etcd {{.Values.etcd.service.port}}; do date; echo "Waiting for etcd..."; sleep 1; done;']
+        command: ['sh', '-c', 'trap "exit 1" TERM; until nc -vzw 5 {{ .Release.Name }}-etcd {{ .Values.etcd.service.ports.client }}; do date; echo "Waiting for etcd..."; sleep 1; done;']
```
tiagolobocastro, line 80: do we need to default to .Values.etcd.service.port in case the new one isn't set?

tiagolobocastro (commented):
Have you tested "manual" upgrades without using the upgrade-job ?

chart/label-etcd-for-helm-release.sh
niladrih: The bitnami/kubectl container image doesn't have bash on it.
I've fixed the shellcheck issues.

niladrih:
bors try

niladrih:
> Have you tested "manual" upgrades without using the upgrade-job

Yes I have tested manual helm upgrade.

chart/label-etcd-for-helm-release.sh
niladrih: Yes, the reason I've preferred to go with `awk`, is because we're on sh and because there's multiple operations and I wanted to avoid calling the `sed` binary several times. So, no real reason to prefer `awk` :).

chart/values.yaml
niladrih: This is required so that we may use openebs/alpine-bash for the volume permissions run that Etcd does. PTAL here: https://github.com/bitnami/charts/issues/30850

chart/values.yaml
niladrih: Yeah, I had that there earlier, until I noticed that the actual etcd chart doesn't work with `.Values.etcd.service.port` anymore, even if our etcd-probes work correctly. So, to keep things consistent, I've removed it. WDYT?

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
That script looks a bit difficult to review, thanks for the nice documentation 👏🏻  , looks good! I hope it's been tested considering the facts:
1. Local Hostpath Storage!
2. All etcd pods shouldn't be down at once? Also do we need to worry about which is the master and kill that at last if it makes sense?

chart/label-etcd-for-helm-release.sh
```diff
@@ -0,0 +1,313 @@
...
+#
+# But, that's easily fixable with the v9.0.0 upgrade steps.. we label the Pods, we delete the StatefulSet and orphan the
+#   Pods, and after helm upgrade the new StatefulSet comes up and fixes everything, right? No! The new StatefulSet gets
+#   to come up only after the Etcd pre-upgrade Job is through. And for users for whom, no. of nodes == Etcd replicas,
+#   and they're on v8.6.0, they can't get the pre-upgrade Job to schedule. So, they can only fix podAntiAffinity w.r.t.
+#   the pre-upgrade Job after the pre-upgrade Job has been scheduled and has run to completion.
+#
+# The solution for this problem, on this script, is to grab the existing 8.6.0 Etcd StatefulSet object, plug the labels
+#   in, set the cluster env to join an existing cluster. Then we follow the usual v9.x-like flow, we label the Pods and
+#   delete the StatefulSet, while orphaning the Pods. Then we delete the StatefulSet and create the labelled one. The
```
Abhinandan-Purkait: Sorry did I catch this correct, 2 times deletion of statefulset?

Abhinandan-Purkait: When did we create the second statefulset?

chart/label-etcd-for-helm-release.sh
```diff
@@ -0,0 +1,313 @@
...
+#   new 8.6.0 StatefulSet's rollout controller will fix the affinity on the Pods one-by-one and bring them up. All of
```
Abhinandan-Purkait: Do you mean new <new version> rather than `8.6.0`?

chart/label-etcd-for-helm-release.sh
niladrih: One delete, not two, fixed the writeup. 

Get Etcd statefulset --> label it locally in-memory --> Label Etcd Pods --> Delete StatefulSet --> Create labelled StatefulSet

chart/label-etcd-for-helm-release.sh
niladrih: No, we delete 8.6.0, and re-create 8.6.0 with the new labels.
After the script is through, eventually helm upgrade will change 8.6.0 to the new one.

niladrih:
@Abhinandan-Purkait 

> Local Hostpath Storage!

This should be fine, the StatefulSet controllers can find older volumes and schedule appropriately, as long as the name is correct. We don't change the name, so it should always work.

> All etcd pods shouldn't be down at once? Also do we need to worry about which is the master and kill that at last if it makes sense?

This is unfortunately not possible with the current approach. The bitnami folks seem to suggest leaving the rebalancing to the StatefulSet controller, PTAL here: https://github.com/bitnami/charts/pull/35003#issuecomment-3069584958

chart/values.yaml
tiagolobocastro: Wow I see, it's started...

niladrih:
bors try

chart/templates/_helpers.tpl
```diff
@@ -354,3 +354,31 @@ Get the Events Jetstream Replica Count
...
+
+{{/*
+Returns matched if the Etcd StatefulSet is of v8.6.0
+Usage:
+  {{- if include "is_etcd_8.6.0" . }}
+    Do something
+  {{- end }}
+*/}}
+{{- define "etcd_is_8.6.0" -}}
+  {{- $sts  := lookup "apps/v1" "StatefulSet" .Release.Namespace (printf "%s-etcd" .Release.Name) -}}
```
tiagolobocastro, line 366: You can also short circuit this for upgrades as well? If it's an install we don't need to run this?

chart/values.yaml
```diff
@@ -526,8 +534,11 @@ etcd:
     storageClass: "mayastor-etcd-localpv"
     # -- Volume size
     size: 2Gi
+  persistentVolumeClaimRetentionPolicy:
     # -- PVC's reclaimPolicy
-    reclaimPolicy: "Delete"
+    enabled: true
```
tiagolobocastro: Let's leave as false since Retain is default anyway?

```suggestion
    enabled: false
```

tiagolobocastro (approved):

chart/templates/_helpers.tpl
niladrih: Going with an outer `{{- if .Release.IsUpgrade }}`, coz the go-template and is a function and it gets called with the results of the two operands. So we'd not be able to short-circuit it as we'd do with an && operator. But it's an uglier looking short cicuit nonetheless.

niladrih:
bors merge

PR #673, opened 2025-05-20
fix(upgrade): add support for loki upgrade

niladrih:


abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

k8s/upgrade/src/helm/upgrade.rs
```diff
@@ -329,6 +332,15 @@ impl HelmUpgrader for CoreHelmUpgrader {
             // This handle is dropped when this closure returns, after helm upgrade.
             let _values_file = self.upgrade_values_file;
 
+            let two_dot_o_rc_one = Version::parse(TWO_DOT_O_RC_ONE).context(SemverParse {
```
avishnu, line 335: What's the significance of this variable?

niladrih: This is parsing '2.0.0-rc.1' to a Version type. The Version type is coming from the semver crate. It has the functionality to compare and see >, <, == relationships between two Version type instances. So, the parsed 2.0.0-rc.1 thing is being compared to 2.9.0 thing. The 2.9.0 is simple and is intialised elsewhere using the Version::new() helper function. Unfortunately the new() function does not take the prerelease stuff as input. So that is being initialised using a parse. No helper function there. So that is why this variable is there, and there is no similar line for 2.9.0.

avishnu: Ok. Couple more questions:
1. Does this handle the case where Loki was disabled on source?
2. Instead of doing version check 2.0.0.rc1 < src ver < 2.9.0, will it be better instead to just delete old loki stack if present?

niladrih: We're ignoring a 404 NotFound error, if we get it. That handles the not present thing.


This codebase will extend to 2.10 and onwards. Without the check, this will delete loki sts for 2.9.0 to 2.10+ upgrades.

avishnu: I meant deletion of old loki, not the new one. Unless of course if we can't differentiate..

niladrih: Version is our only point of differentiation at the moment. The old and the new statefulset are quite similar otherwise. I don't exactly remember the cause for the lowerbound of the version check. We do the upper and lower check for value setting, so doing it here as well.

avishnu: Ok

PR #674, opened 2025-05-21
revert: disable loki-stack during helm install ci

niladrih:
A workaround was introduced in PR 663 for testing core upgrade on the CI. This workaround is to disable loki-stack so that the stateful set from the loki chart doesn't cause conflicts with the statefulset from the loki-stack chart. This is not needed any more, a fix has been merged in PR 673


niladrih:
bors try

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

niladrih:
bors merge

PR #681, opened 2025-05-29
fix: set tmp directory to /tmp in env

r1jt:


.github/workflows/helm-testing-chart.yml
```diff
@@ -11,6 +11,8 @@ on:
 jobs:
   helm-testing-chart:
     runs-on: ubuntu-latest
+    env:
+      TMPDIR: /tmp
```
pchandra19: Are you sure this will fix the problem ? Irrespective, this change is not required here. It should go to [stefanprodan/helm-gh-pages@v1.7.0]) step and please test this before merging

pchandra19 (changes requested):

.github/workflows/helm-testing-chart.yml
pchandra19: Also at the some point we should look at moving away from [stefanprodan/helm-gh-pages@v1.7.0] as its not actively maintained any more.

r1jt: Yes. This will fix the problem. I have tested the change.

r1jt: > Also at the some point we should look at moving away from [stefanprodan/helm-gh-pages@v1.7.0] as its not actively maintained any more.

We can revisit at replacing it at a later stage. The github action has been working well for us for quite a while

avishnu (approved):

.github/workflows/helm-testing-chart.yml
tiagolobocastro: how did you test this? we also use same action on another workflow, so is that one also broken?

r1jt: I tested it by creating a fork of the repository and initiating the action with the relevant change. 
If the other workflow has also been replaced with uses: cachix/install-nix-action@v31.3.0 recently, it could be broken.

tiagolobocastro: So you're saying that if you revert the cachix change then it works?

tiagolobocastro: I mean, without TMPDIR modification

tiagolobocastro: Also, please test moving the env to the step itself as Prateek requested above

r1jt: > So you're saying that if you revert the cachix change then it works?

Yes. It will work as before.

r1jt: > Also, please test moving the env to the step itself as Prateek requested above

What value add does moving the env make?

tiagolobocastro: it's not required elsewhere, only for this specific step

pchandra19 (approved):

.github/workflows/helm-testing-chart.yml
r1jt: Tested and moved the tmpdir section

r1jt:
bors merge

pchandra19 (approved):

niladrih (approved):

PR #682, opened 2025-05-29
ci: create auto pr with user token

tiagolobocastro:
Create PRs with the user ci token rather than the github token. This explains why the PR's weren't running the actions until someone would close&reopen the PR.

### DIFF
--- a/.github/workflows/develop-to-release.yml
+++ b/.github/workflows/develop-to-release.yml
@@ -30,7 +30,7 @@ jobs:
           nix-shell ./scripts/helm/shell.nix --pure --run "./scripts/helm/images.sh patch"
       - name: Create Pull Request
         id: cpr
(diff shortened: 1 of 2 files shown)


Abhinandan-Purkait (approved):

pchandra19 (approved):
The github token has the permissions to create PRs so it should work. I don't think we need this change as such but since its a cosmetic change should be fine.

tiagolobocastro:
> The github token has the permissions to create PRs so it should work. I don't think we need this change as such but since its a cosmetic change should be fine.

Nope, not just a cosmetic change:
https://github.com/peter-evans/create-pull-request/issues/48
Here's the proof of the top PRs raised by the bot:
https://github.com/openebs/mayastor-extensions/pulls?q=is%3Apr+prepare-release-branch+is%3Aclosed

bors merge

PR #692, opened 2025-06-06
Fix Helm Chart Images on Prereleases

tiagolobocastro:
ci: update chart images on prereleases

---

    build: cleanup helm deps as folders
    
    The helm release.sh downloads the deps as folders which is then used by the
    build sandbox.
    This however creates a buggy behaviour in the image list script, if it is
    run ahead of the release script.
    Here we simply clean up the folders.
    todo: we don't even need them to begin with, we can use the tar files!


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #698, opened 2025-06-10
ci(helm-pins): add helm pins chart publishing

tiagolobocastro:
Publishes a "helm-pins" chart by pinning all mayastor container images to a commit hash as well as setting the helm chart version to a monotonically increasing number.
For release branches: 2.9.0-pin-devrel.1
For develop branches: 2.10.0-pin-dev.1 where the version is the next one, as determined via the latest release branch.

### DIFF
--- a/nix/pkgs/images/default.nix
+++ b/nix/pkgs/images/default.nix
@@ -58,6 +58,7 @@ let
     # Script doesn't need to be used with main branch `--alias-tag <main-branch-style-tag>`.
     # The repo chart is already prepared.
     if [[ "$(semver validate ${tag})" == "valid" ]] &&
+      [[ ! ${tag} =~ ^(v?[0-9]+\.[0-9]+\.[0-9]+-pin-(dev|devrel).([0-9]+))$ ]] &&
       [[ ! ${tag} =~ ^(v?[0-9]+\.[0-9]+\.[0-9]+-0-(main|release)-unstable(-[0-9]+){6}-0)$ ]]; then
       CHART_FILE=build/chart/Chart.yaml build/scripts/helm/publish-chart-yaml.sh --app-tag ${tag} --override-index ""
     fi
--- a/scripts/helm/install.sh
+++ b/scripts/helm/install.sh
@@ -35,6 +35,7 @@ FAIL_IF_INSTALLED=
 HOSTED=
 VERSION=
 PULL_POLICY=
+INS_LOKI="true"
 DEFAULT_REGISTRY="https://openebs.github.io/mayastor-extensions"
 HELM="helm"
 KUBECTL="kubectl"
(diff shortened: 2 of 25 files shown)


tiagolobocastro:
bors try

nix/pkgs/images/default.nix
```diff
@@ -58,6 +58,7 @@ let
     # Script doesn't need to be used with main branch `--alias-tag <main-branch-style-tag>`.
     # The repo chart is already prepared.
     if [[ "$(semver validate ${tag})" == "valid" ]] &&
+      [[ ! ${tag} =~ ^(v?[0-9]+\.[0-9]+\.[0-9]+-pin-(dev|devrel).([0-9]+))$ ]] &&
```
Abhinandan-Purkait, line 61: nit: Can we keep it similar to what we follow for locals? i.e develop and prerelease, rather than dev and devrel? Just to keep things similar.

scripts/helm/install.sh
```diff
@@ -72,6 +74,51 @@ die() {
...
+  echo -n "$(set_args "loki.loki.commonConfig.replication_factor=$replicas")" \
+          "$(set_args "loki.singleBinary.replicas=$replicas")" \
+          "$(set_args "loki.minio.mode=$mode") "
+
+  if [ "$minio" = "true" ]; then
+    echo -n "$(set_args "loki.minio.enabled=true,loki.minio.replicas=$replicas")"
+  else
+    echo -n "$(set_args "loki.loki.storage.type=filesystem")" \
+            "$(set_args "loki.singleBinary.persistence.enabled=true")" \
+            "$(set_args "loki.minio.enabled=false")"
```
Abhinandan-Purkait, line 118: This might not be enough tbh, we need to set object_store as filesystem as well. You can check the loki-storage doc

Abhinandan-Purkait (approved):

niladrih (approved):

nix/pkgs/images/default.nix
tiagolobocastro: but this is not the same though, that's just vXXX-(develop|prerelease), doesn't have the pinning;
though I agree we should change mayastor 0.0.0 to v$next-develop and the release to -prerelease to keep it consistent.
and once we change, it won't clash with the pinning, should we ever place them on the same repo

scripts/helm/install.sh
tiagolobocastro: Yeah I know... it's broken but atm fixing is more effort then what's worth it for me as I can just disable loki

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #702, opened 2025-06-20
ci(pin): don't check if "latest" already exists

tiagolobocastro:
Since it will exist and it's not an error, only if the pinned one exists is an error


Abhinandan-Purkait (approved):

niladrih (approved):

tiagolobocastro:
bors merge

PR #723, opened 2025-08-07
fix: move bitnami images to openebs

tiagolobocastro:
The bitnami docker repo is going to be migrated on the 25Nov, so 
let's just move these to our dockerhub since that one is not going anywhere.
We should consider doing this for other dependencies.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -549,8 +549,11 @@ etcd:
       repository: openebs/alpine-bash
       tag: 4.2.0
       pullSecrets: []
-  # extra debug information on logs
-  debug: false
+  image:
+    registry: docker.io
+    repository: openebs/etcd
+    # extra debug information on logs
+    debug: false
   # -- Pod anti-affinity preset
   # Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity
   podAntiAffinityPreset: "hard"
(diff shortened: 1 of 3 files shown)


tiagolobocastro:
bors try

avishnu (approved):

tiagolobocastro:
@niladrih do we need upgrade changes for this?

Abhinandan-Purkait (approved):

chart/values.yaml
```diff
@@ -549,8 +549,11 @@ etcd:
       repository: openebs/alpine-bash
       tag: 4.2.0
       pullSecrets: []
-  # extra debug information on logs
-  debug: false
+  image:
+    registry: docker.io
+    repository: openebs/etcd
+    # extra debug information on logs
+    debug: false
```
niladrih, line 556: I see the `debug: false` being moved inside `image:`. I am not sure why this is here. This change was already made, and develop already has debug inside `image:`. But your base branch seems to not have this change, how?

niladrih (approved):

chart/values.yaml
niladrih: Nvm. Noticed this is PR to a release branch, and not develop.

niladrih:
> @niladrih do we need upgrade changes for this?

Yes, but this change is already being done in 2.10, we may ignore this?

niladrih:
Add upgrade changes to PR 724

tiagolobocastro:
bors merge

PR #730, opened 2025-08-19
Kubectl Plugin: Snapshot Deletion

tiagolobocastro:
chore: use stdenv with no CC in the nix-shell
    
    This reduces the derivation size when not using internal rust and also
    can prevent ring rebuilds due to CC env var mismatch.

---

    fix: move bitnami images to openebs
    
    The bitnami docker repo is going to be migrated on the 25Nov, so
    let's just move these to our dockerhub since that one is not going anywhere.
    We should consider doing this for other dependencies.

---

    chore: add pre-commit trampoline
    
    Allows us to erase the the references which pre-commit now seems
    to leave around and which interfere with nix-shell and venv.
    
----

    feat: add snapshot delete subcommand
    
    Similar to the volume, it first ensures that the volumesnapshotcontent is
    not present! If it is, then it bails out with an error.

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -54,10 +54,6 @@ dependencies:
     condition: localpv-provisioner.enabled
 annotations:
   helm.sh/images: |
-    - name: etcd
-      image: docker.io/bitnamilegacy/etcd:3.6.4-debian-12-r0
-    - name: kubectl
-      image: docker.io/bitnamilegacy/kubectl:1.25.15
     - name: alloy
       image: docker.io/grafana/alloy:v1.8.1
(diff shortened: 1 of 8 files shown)


tiagolobocastro:
kubectl-mayastor plugin changes for https://github.com/openebs/mayastor/issues/1892

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

niladrih (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #744, opened 2025-08-30
refactor(plugin/support): add wrapper for kubeconfig path and opts

tiagolobocastro:
We're now parsing the context name, but this was not passed to the support bundle.
Rather than passing more arguments, create a new type which includes the path and the options.

### DIFF
--- a/k8s/supportability/src/collect/common.rs
+++ b/k8s/supportability/src/collect/common.rs
@@ -15,8 +15,8 @@ pub struct DumpConfig {
     etcd_uri: Option<String>,
     /// Period states to collect logs from specified duration
     since: humantime::Duration,
-    /// Path to kubeconfig file, which requires to interact with Kube-Apiserver
-    kube_config_path: Option<std::path::PathBuf>,
+    /// The kubeconfig options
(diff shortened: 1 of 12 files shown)


tiagolobocastro:
@voyvodov this bit was missing

Abhinandan-Purkait (approved):

k8s/supportability/src/collect/common.rs
```diff
@@ -49,7 +49,7 @@ impl DumpConfig {
         loki_uri: Option<String>,
         etcd_uri: Option<String>,
         since: humantime::Duration,
-        kube_config_path: Option<std::path::PathBuf>,
+        kubeconfig: crate::KubeConfigArgs,
```
niladrih, line 52: 👍

niladrih (approved):

k8s/supportability/src/collect/common.rs
tiagolobocastro: hmm, that's not there?

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #749, opened 2025-09-10
feat(scripts): add script to replace one etcd laggard member

niladrih:
This adds a script to replace one laggard etcd member.

## How to test this:

Get hold of a kubernetes cluster with the mayastor chart installed (2.9.x or older). Put in these two lines at line 533

```bash
# In this one we're trying to get rid of mayastor-etcd-1.
# We're doing a simple force removal to test the behaviour, otherwise the script won't remove anything on a healthy cluster.
#

say "Merged best-known revisions (per pod):"
for pod in "${!BEST_REV[@]}"; do printf '  - %s  %s\n' "$pod" "${BEST_REV[$pod]}"; done

LAGGARDS=("mayastor-etcd-1")
BEST_REV["mayastor-etcd-1"]=$(( max_rev>0 ? max_rev-1 : 0 ))

# Guardrails: either everybody matches, or exactly one laggard; otherwise bail.
if ((${#LAGGARDS[@]} == 0)); then
  say "All ${REPLICAS} etcd instances are in sync. Revision=${max_rev}"
  exit 0
fi
if ((${#LAGGARDS[@]} > 1)); then
  say "More than one member appears behind: ${LAGGARDS[*]}. Refusing to auto-replace."
  exit 4
fi
```

Or.. you may use tc netfilters to block 2380 traffic on the pod network to simulate a network partition. `iptables` won't work with some CNIs which use eBPF to get around the iptables rules.

You may generate writes to the cluster using the shell on one of the etcd containers. This shell will close itself when we update the ETCD_INITIAL_CLUSTER_STATE env across the cluster, so watch out...
```bash
for i in $(seq 1 200000); do
  etcdctl put /lagtest/$i "$(date +%s%N)"
done
```

### DIFF
--- a/scripts/k8s/replace-lagging-etcd-member/README.md
+++ b/scripts/k8s/replace-lagging-etcd-member/README.md
@@ -1,0 +1,175 @@
+# replace-one-laggard.sh
+
+## Overview
+
+This script provides automated, safe single-node remediation for a Bitnami-based etcd cluster running as a Kubernetes StatefulSet. It detects and replaces exactly one lagging member that has fallen behind others in terms of raft applied index (revision), ensuring the cluster maintains quorum and health throughout the operation.
+
+## What It Does
+
+The script performs the following operations:
+
+1. **Detects** exactly one member that is lagging behind others by comparing raft applied indices
+2. **Validates** the entire cluster is on the same StatefulSet ControllerRevision (ensuring consistent templates/flags)
+3. **Performs double health gates**:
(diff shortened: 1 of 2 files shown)


tiagolobocastro (approved):
A bit insane reviewing a 1k lines single-file bash script tbh, from a birds eye looks ok, but don't take my work for it
I hope you've tested all the error cases here and I'd still suggest adding a big "use at your own risk" banner

niladrih:
bors merge

PR #776, opened 2025-11-17
feat(upgrade): normalize k8s name

niladrih:
This change trims kubernetes names to at-max 63 characters long.


tiagolobocastro:
This is not needed anymore right?

PR #787, opened 2025-11-24
ci: publish mayastor helm chart as oci

tiagolobocastro:
Publish mayastor helm chart as OCI

### DIFF
--- a/.github/workflows/release-chart.yml
+++ b/.github/workflows/release-chart.yml
@@ -31,12 +31,25 @@ jobs:
             nix-shell --pure --run "./scripts/helm/publish-chart-yaml.sh --check-chart "$branch"" ./scripts/helm/shell.nix
           fi
           nix-shell --pure --run "SKIP_GIT=1 ./scripts/helm/generate-readme.sh" ./scripts/helm/shell.nix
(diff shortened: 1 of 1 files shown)


Abhinandan-Purkait (approved):

.github/workflows/release-chart.yml
```diff
@@ -31,12 +31,25 @@ jobs:
...
+        if: ${{ github.ref_type == 'tag' }}
+        with:
+          name: mayastor
+          repository: ${{ github.repository_owner }}/charts
+          tag: ${{ env.CHART_VERSION }}
+          path: ./chart
+          registry: ghcr.io
+          registry_username: ${{ github.actor }}
+          registry_password: ${{ secrets.GITHUB_TOKEN }}
+          update_dependencies: 'false'
```
Abhinandan-Purkait, line 52: We probably need the deps update here, as this is not staging, so it probably doesn't have the deps?

tiagolobocastro: Also already updated above on the publish script

tiagolobocastro: Or rather, the images I think: https://github.com/openebs/mayastor-extensions/actions/runs/19643647415/job/56253395323

tiagolobocastro:
@mergify backport develop

tiagolobocastro:
@mergify backport develop

PR #791, opened 2025-12-05
chore: add staging workflow

pchandra19:
## Description

Added staging workflow so that we can stage changes and test them prior to release.

## Motivation and Context

The staging workflows will act as a safe, controlled pre-production validation layer which ensures:
- Everything builds
- Everything passes checks
- Charts are valid
- Images exist
- Workflows run correctly
- No surprises during release

## How Has This Been Tested?
This has been tested in a fork with a simulated environment.

## Types of changes
- [x] Breaking change (This change modifies our CI workflows and scripts)


.github/workflows/develop-release-chart.yml
```diff
@@ -1,8 +1,9 @@
-name: Develop Helm Chart
+name: Release and Develop Helm Chart
 on:
   push:
     branches:
       - develop
+      - 'release/**'
```
tiagolobocastro: Why do we need to run this on the release chart? It's not publishing it

.github/workflows/release-binaries.yml
```diff
@@ -30,6 +34,7 @@ jobs:
           - os: macos-13
             target: apple-darwin
             arch: x86_64
+            platform: x86_64-darwin
```
tiagolobocastro: Why is this being added here, doesn't seem related to staging?

.github/workflows/release-binaries.yml
```diff
@@ -56,15 +62,26 @@ jobs:
...
           enable_kvm: false
       - run: |
-          nix-build -A utils.release.${{ matrix.arch }}.${{ matrix.target }}.kubectl-plugin --arg incremental false ${{ matrix.system }}
+          TAG_ARG=""
+          SYSTEM_ARG=""
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            TAG_ARG="--argstr tag ${{ inputs.tag }}"
+          fi
+          if [[ -n "${{ matrix.platform }}" ]]; then
+            SYSTEM_ARG="--argstr system ${{ matrix.platform }}"
```
tiagolobocastro: s/platform/system

.github/workflows/release.yml
```diff
@@ -0,0 +1,168 @@
...
+        uses: appany/helm-oci-chart-releaser@v0.4.2
+        with:
+          name: mayastor
+          repository: ${{ github.repository_owner }}/charts
+          tag: ${{ env.CHART_VERSION }}
+          path: ./chart
+          registry: ghcr.io
+          registry_username: ${{ github.actor }}
+          registry_password: ${{ secrets.GITHUB_TOKEN }}
+          update_dependencies: 'false'
```
tiagolobocastro:
```suggestion
          update_dependencies: 'false'

```

tiagolobocastro (approved):

.github/workflows/release-binaries.yml
Abhinandan-Purkait: Let's not do that, this is specific if you want to run on macos-14 which is m1 chip.

Abhinandan-Purkait: I mean either way, if we are doing it then let's run on macos-14

.github/workflows/release.yml
Abhinandan-Purkait: Line.

Abhinandan-Purkait (approved):

.github/workflows/release.yml
pchandra19: fixed

.github/workflows/release-binaries.yml
pchandra19: Reverted the change

.github/workflows/release-binaries.yml
pchandra19: NA anymore as reverted the change.

.github/workflows/develop-release-chart.yml
```diff
@@ -1,4 +1,4 @@
-name: Develop Helm Chart
+name: Release and Develop Helm Chart
```
tiagolobocastro:
```suggestion
name: Develop Helm Chart
```

pchandra19: Fixed. Have also updated the workflow file name accordingly.

tiagolobocastro (commented):
Do we need to merge mayastor and control-plane first?

pchandra19:
> Do we need to merge mayastor and control-plane first?

Need to first merge dependencies and then update submodules on each of DP , CP and extensions and then perform the merge.  CP, DP, and extensions are independent, but merging them in order is preferable (because of submodule dependencies).

pchandra19:
bors merge

pchandra19:
bors merge

PR #803, opened 2025-12-30
Fix helm pins

tiagolobocastro:
ci: fix pin.sh check for missing repo
    
---

    fix(scripts): use latest semantic versioned release branch
    
    The git refname sorting is lexical only, and thus 2.9 is > 2.10!
    Fix this by using `sort -V` instead.

### DIFF
--- a/scripts/helm/pin.sh
+++ b/scripts/helm/pin.sh
@@ -46,24 +46,6 @@ Examples:
 EOF
 }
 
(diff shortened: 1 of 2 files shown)


niladrih (approved):

abhilashshetty04 (approved):

niladrih:
bors merge

PR #804, opened 2026-01-02
fix: Prerequisites hyperlink in helm chart's readme file

krishnaGajabi:
This PR fixes the prerequisites hyperlink in helm chart's readme file.

Earlier it was pointing to the older gitbook website.

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -8,7 +8,7 @@ Mayastor Helm chart for Kubernetes
 
 ### Prerequisites
 
- - Make sure the [system requirement pre-requisites](https://mayastor.gitbook.io/introduction/quickstart/prerequisites) are met.
+ - Make sure the [system requirement pre-requisites](https://openebs.io/docs/quickstart-guide/prerequisites#replicated-pv-mayastor-prerequisites) are met.
  - Label the storage nodes same as the mayastor.nodeSelector in values.yaml
  - Create the namespace you want the chart to be installed, or pass the `--create-namespace` flag in the `helm install` command.
    ```sh
--- a/chart/README.md.tmpl
+++ b/chart/README.md.tmpl
@@ -7,7 +7,7 @@
 
 ### Prerequisites
 
- - Make sure the [system requirement pre-requisites](https://mayastor.gitbook.io/introduction/quickstart/prerequisites) are met.
+ - Make sure the [system requirement pre-requisites](https://openebs.io/docs/quickstart-guide/prerequisites#replicated-pv-mayastor-prerequisites) are met.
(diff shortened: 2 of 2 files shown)


tiagolobocastro (approved):
@balaharish7 can we have a versioned url here?

balaharish7:
> @balaharish7 can we have a versioned url here?

It’s generally not advised to version the URL for the latest documentation site because it creates long-term usability and operational problems. A Stable Entry Point Is Critical for Users. Search engines, bookmarks, and support links always land on current guidance.

If the latest docs were versioned (for example /v4.4/), every release would:
- Instantly make existing links outdated
- Require constant redirects
- Increase user confusion (“Am I reading old docs?”)
cc: @Alex130469

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
bors d+

tiagolobocastro:
> > @balaharish7 can we have a versioned url here?
> 
> It’s generally not advised to version the URL for the latest documentation site because it creates long-term usability and operational problems. A Stable Entry Point Is Critical for Users. Search engines, bookmarks, and support links always land on current guidance.

But this is for the helm docs of a specific version; so let's say you install v2.10.0, if we point to the latest then it'd point to v2.11.0 for example, and the user may not even notice it

balaharish7:
> > > @balaharish7 can we have a versioned url here?
> > 
> > 
> > It’s generally not advised to version the URL for the latest documentation site because it creates long-term usability and operational problems. A Stable Entry Point Is Critical for Users. Search engines, bookmarks, and support links always land on current guidance.
> 
> But this is for the helm docs of a specific version; so let's say you install v2.10.0, if we point to the latest then it'd point to v2.11.0 for example, and the user may not even notice it

Our docs intentionally default to the latest version at a stable URL, but we provide a clearly visible version selector that allows users to switch to the exact version they are running. This ensures new users always see the most up-to-date guidance, while users installing or operating a specific release can explicitly select and follow version-accurate documentation.
image

tiagolobocastro:
> Our docs intentionally default to the latest version at a stable URL, but we provide a clearly visible version selector that allows users to switch to the exact version they are running. This ensures new users always see the most up-to-date guidance, while users installing or operating a specific release can explicitly select and follow version-accurate 

Only if they're looking at it, and cross-check it to the version which they've installed.
We've had at least 1 such incident on slack before.

Another problem is that the exact url may change form version to version, so an unversioned url may actually become out of date, and we've also hit this before.

krishnaGajabi:
bors merge

krishnaGajabi:
bors cancel

krishnaGajabi:
bors merge

PR #814, opened 2026-02-02
fix(chart): wait for etcd upgrade rollout before exiting

niladrih:
It's hard to reason about https://github.com/openebs/openebs/issues/4150. One possible way to end up in a state like this may be that there are 2 rollouts of the etcd happening -- First rollout happens when the 8.6.0 etcd is deleted with orphaned pods and recreated with new labels. The bitnami preUpgradeJob runs against this new rollout Pods. Now, while this rollout is still in-progress the bitnami job may exit quickly and helm may apply the 12.x etcd chart, triggering another rollout on a cluster which is potentially unhealthy. Maybe the bitnami preUpgradeJob's changes couldn't propagate to the entire cluster as the rollout was partially done. And then the second rollout happens from helm's apply. The reasoning is that between the successful Bitnami preUpgradeJob completion and the 8.6.0 Etcd cluster rebalancing itself and reaching a healthy state, the helm 12.x rollout tears it apart.

If the preUpgradeJob's write operations are able to succeed, this means, the cluster has quorum. Else the job will fail and be retried, until it succeeds (tested scenario). If the cluster has quorum, this would mean it has data availability. But compromised survivability, and the new helm rollout may start killing Pods and destabilize the cluster.

There's also another possibility, less likely. The initial rollout was so slow that the etcd preUpgradeJob was able to spawn and finish against the 8.6.0 cluster. And while the rollout was happening, the next rollout was also issued, and the cluster was on the brink of quorum, and the eventually lost consistency.

Anyway, these are assumptions, and it might be a good idea for us to wait for the first rollout to complete completely.


tiagolobocastro (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

PR #816, opened 2026-02-06
Feat: OEP 4111 node status metrics

### LINKED ISSUE #4111: [OEP 4111]: Mayastor Node-Status Metrics via Metrics-Exporter REST Integration
### Enhancement Description

- Introduce a standardized method for exposing Mayastor node-status metrics by extending the metrics-exporter with a lightweight REST client that sources authoritative state from the control-plane.
- OpenEBS Enhancement Proposal: openebs/openebs/pull/4107
- Discussion Link: 
    - Initial implementation PR: openebs/mayastor-control-plane/pull/1035
    - Link to Slack: (https://kubernetes.slack.com/archives/CUAKPFU78/p1762287322900259)
- Primary contact (assignee): @IyanekiB / @pjgranieri
- Responsible OWNERs: @tiagolobocastro @Abhinandan-Purkait @niladrih 
- Targeted milestone:
  - OpenEBS vx.y (TBD)
- [ ] Targets
  - [x] OEP update PR(s):
    - [x] [OEP-4111: Mayastor Node Status Metrics Implementation Updates#4158](https://github.com/openebs/openebs/pull/4158)
  - [x] Code update PR(s):
    - [x] [Feat: OEP 4111 node status metrics#816](https://github.com/openebs/mayastor-extensions/pull/816)
  - [ ] E2E Test PR(s):
  - [ ] Docs update PR(s):

IyanekiB:
Implements [OEP-4111](https://github.com/openebs/openebs/pull/4107) to expose Mayastor node status metrics through the metrics-exporter. Aims to resolve [#4111](https://github.com/openebs/openebs/issues/4111).

## Description

This PR implements [OEP-4111](https://github.com/IyanekiB/oep-mayastor-node-status-metrics/blob/feat/oep-4111-node-status-metrics/designs/replicated-pv/mayastor/mayastor-node-status-metrics-via-metrics-exporter.md), extending the metrics-exporter with REST client capabilities to expose Mayastor node status metrics. The implementation adds a lightweight REST client that periodically polls the control-plane `/v0/nodes` endpoint and exposes node state as Prometheus-compatible gauges.

Key changes:

- REST client for `/v0/nodes` endpoint with connection pooling and timeout handling
- Five new Prometheus metrics:

  - `mayastor_node_online` (0/1) - node online status
  - `mayastor_node_cordoned` (0/1) - node cordoned status
  - `mayastor_node_draining` (0/1) - node draining status
  - `mayastor_node_status_last_fetch_seconds` - staleness detection (not used now)
  - `mayastor_node_status_fetch_errors_total` - failure tracking (not used now)

- Periodic polling with jitter (15s interval + 0-5s random jitter)
- In-memory caching with thread-safe RwLock
- Immediate fetch on startup (no initial delay)
- Configuration via CLI flags and environment variables

## Motivation and Context

This consolidates node-status metrics into the metrics-exporter, eliminating duplicated business logic and fragmented observability surfaces. Based on what we learned from [PR #1035](https://github.com/openebs/mayastor-control-plane/pull/1035), this approach aims to create a maintainable pattern for REST-backed metrics while preserving compatibility with existing Prometheus/Grafana pipelines.
The control-plane remains the authoritative source of node state; the exporter simply polls, caches, and exposes this data in a Prometheus-compatible format.

## Regression

No

## How Has This Been Tested?

_Client tests (wiremock-based):_

- Successful multi-node fetch
- Cordoned/draining/offline node detection
- HTTP error handling (500 responses)
- Invalid JSON handling
- Empty response handling
- Rapid state transitions

_Collector tests:_

- Metric descriptor validation (5 metrics)
- Empty cache behavior
- Online/offline/cordoned/draining state verification
- State transition handling
- OEP-4111 metric naming compliance

_Manual validation:_

- Local Prometheus scrape validation confirmed metric format and labels
- Verified immediate startup fetch (no 15s delay)
- Tested staleness detection via last_fetch_seconds gauge
- Validated error counter increments on REST failures

## Types of changes

## Checklist:

- [x] I have added unit tests to cover my changes.

**CC:** @pjgranieri

### DIFF
--- a/metrics-exporter/Cargo.toml
+++ b/metrics-exporter/Cargo.toml
@@ -23,3 +23,9 @@ tonic = { workspace = true }
 serde = { workspace = true }
 mime = { workspace = true }
 tracing = { workspace = true }
+
+# OEP-4111: Node status metrics dependencies
+openapi = { path = "../dependencies/control-plane/openapi" }
+anyhow = { workspace = true }
+humantime = { workspace = true }
+url = { workspace = true }
--- a/metrics-exporter/src/bin/io_engine/collector/node_status.rs
+++ b/metrics-exporter/src/bin/io_engine/collector/node_status.rs
@@ -1,0 +1,367 @@
+use openapi::models::{CordonDrainState, Node, NodeStatus};
+use prometheus::{
+    core::{Collector, Desc},
+    proto, GaugeVec, Opts,
+};
+
+/// Returns true if the node is online.
+fn is_online(node: &Node) -> bool {
+    node.state
+        .as_ref()
+        .map(|s| matches!(s.status, NodeStatus::Online))
+        .unwrap_or(false)
+}
+
+/// Returns true if the node is draining or drained.
+fn is_draining(node: &Node) -> bool {
+    node.spec
+        .as_ref()
+        .and_then(|s| s.cordondrainstate.as_ref())
+        .map(|cds| {
+            matches!(
+                cds,
+                CordonDrainState::drainingstate(_) | CordonDrainState::drainedstate(_)
+            )
+        })
+        .unwrap_or(false)
+}
+
+/// Returns true if the node is cordoned (including implicit cordon from drain).
+fn is_cordoned(node: &Node) -> bool {
+    is_draining(node)
+        || node
+            .spec
+            .as_ref()
+            .and_then(|s| s.cordondrainstate.as_ref())
+            .map(|cds| matches!(cds, CordonDrainState::cordonedstate(_)))
+            .unwrap_or(false)
+}
+
+/// Collector for node status metrics.
+pub(crate) struct NodeStatusCollector {
+    /// Node data fetched at scrape time (None if unavailable).
+    node: Option<Node>,
+    /// Gauge for node online status (0/1).
+    node_online: GaugeVec,
+    /// Gauge for node cordoned status (0/1).
+    node_cordoned: GaugeVec,
+    /// Gauge for node draining status (0/1).
+    node_draining: GaugeVec,
+    /// Descriptors for Prometheus.
+    descs: Vec<Desc>,
+}
+
+impl NodeStatusCollector {
+    /// Create a new NodeStatusCollector from node data fetched at scrape time.
+    pub(crate) fn new(node: Option<Node>) -> Self {
+        let mut descs = Vec::new();
+
+        let node_online = GaugeVec::new(
+            Opts::new("node_online", "Indicates if Mayastor node is online")
+                .subsystem("mayastor")
+                .variable_labels(vec!["node_id".to_string()]),
+            &["node_id"],
+        )
+        .expect("Unable to create gauge metric for node_online");
+        descs.extend(node_online.desc().into_iter().cloned());
+
+        let node_cordoned = GaugeVec::new(
+            Opts::new("node_cordoned", "Indicates if Mayastor node is cordoned")
+                .subsystem("mayastor")
+                .variable_labels(vec!["node_id".to_string()]),
+            &["node_id"],
+        )
+        .expect("Unable to create gauge metric for node_cordoned");
+        descs.extend(node_cordoned.desc().into_iter().cloned());
+
+        let node_draining = GaugeVec::new(
+            Opts::new("node_draining", "Indicates if Mayastor node is draining")
+                .subsystem("mayastor")
+                .variable_labels(vec!["node_id".to_string()]),
+            &["node_id"],
+        )
+        .expect("Unable to create gauge metric for node_draining");
+        descs.extend(node_draining.desc().into_iter().cloned());
+
+        Self {
+            node,
+            node_online,
+            node_cordoned,
+            node_draining,
+            descs,
+        }
+    }
+
+    /// Update metrics from node data fetched at scrape time.
+    fn update_metrics(&self) {
+        let node = match &self.node {
+            Some(n) => n,
+            None => return,
+        };
+
+        let node_id = node.id.as_str();
+
+        self.node_online
+            .with_label_values(&[node_id])
+            .set(if is_online(node) { 1.0 } else { 0.0 });
+
+        self.node_cordoned
+            .with_label_values(&[node_id])
+            .set(if is_cordoned(node) { 1.0 } else { 0.0 });
+
+        self.node_draining
+            .with_label_values(&[node_id])
+            .set(if is_draining(node) { 1.0 } else { 0.0 });
+    }
+}
+
+impl Collector for NodeStatusCollector {
+    fn desc(&self) -> Vec<&Desc> {
+        self.descs.iter().collect()
+    }
+
+    fn collect(&self) -> Vec<proto::MetricFamily> {
+        self.update_metrics();
+
+        let mut metrics = Vec::new();
+        metrics.extend(self.node_online.collect());
+        metrics.extend(self.node_cordoned.collect());
+        metrics.extend(self.node_draining.collect());
+        metrics
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use openapi::models::{CordonedState, DrainState, Node, NodeSpec, NodeState};
+    use prometheus::core::Collector;
+
+    /// Create a test node with the specified status and cordon/drain state.
+    fn create_test_node(id: &str, online: bool, cordon_drain: Option<CordonDrainState>) -> Node {
+        let status = if online {
+            NodeStatus::Online
+        } else {
+            NodeStatus::Offline
+        };
+        Node {
+            id: id.to_string(),
+            status: None,
+            spec: Some(NodeSpec::new_all(
+                format!("http://{id}:10124"),
+                id,
+                None::<std::collections::HashMap<String, String>>,
+                cordon_drain,
+                None::<String>,
+                None::<String>,
+                None::<bool>,
+            )),
+            state: Some(NodeState::new_all(
+                format!("http://{id}:10124"),
+                id,
+                status,
+                None::<String>,
+                None::<String>,
+            )),
+        }
+    }
+
+    /// Helper to extract metric value from collected metrics.
+    fn get_metric_value(metrics: &[proto::MetricFamily], name: &str, node_id: &str) -> Option<f64> {
+        for mf in metrics {
+            if mf.get_name() == name {
+                for metric in mf.get_metric() {
+                    for label in metric.get_label() {
+                        if label.get_name() == "node_id" && label.get_value() == node_id {
+                            return Some(metric.get_gauge().get_value());
+                        }
+                    }
+                }
+            }
+        }
+        None
+    }
+
+    #[test]
+    fn test_collector_creation() {
+        let collector = NodeStatusCollector::new(None);
+        // 3 metrics: online, cordoned, draining.
+        assert_eq!(collector.desc().len(), 3);
+    }
+
+    #[test]
+    fn test_collector_no_node() {
+        let collector = NodeStatusCollector::new(None);
+        let metrics = collector.collect();
+        // 3 metric families returned, but with no data points.
+        assert_eq!(metrics.len(), 3);
+        for mf in &metrics {
+            assert_eq!(mf.get_metric().len(), 0);
+        }
+    }
+
+    #[test]
+    fn test_collector_online_node() {
+        let node = create_test_node("io-engine-1", true, None);
+        let collector = NodeStatusCollector::new(Some(node));
+
+        let metrics = collector.collect();
+
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_online", "io-engine-1"),
+            Some(1.0)
+        );
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_cordoned", "io-engine-1"),
+            Some(0.0)
+        );
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_draining", "io-engine-1"),
+            Some(0.0)
+        );
+    }
+
+    #[test]
+    fn test_collector_offline_node() {
+        let node = create_test_node("io-engine-1", false, None);
+        let collector = NodeStatusCollector::new(Some(node));
+
+        let metrics = collector.collect();
+
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_online", "io-engine-1"),
+            Some(0.0)
+        );
+    }
+
+    #[test]
+    fn test_collector_cordoned_node() {
+        let cordon =
+            CordonDrainState::cordonedstate(CordonedState::new(vec!["maintenance".to_string()]));
+        let node = create_test_node("io-engine-1", true, Some(cordon));
+        let collector = NodeStatusCollector::new(Some(node));
+
+        let metrics = collector.collect();
+
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_cordoned", "io-engine-1"),
+            Some(1.0)
+        );
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_online", "io-engine-1"),
+            Some(1.0)
+        );
+    }
+
+    #[test]
+    fn test_collector_draining_node() {
+        let draining = CordonDrainState::drainingstate(DrainState::new(
+            vec!["maintenance".to_string()],
+            vec!["drain-volumes".to_string()],
+        ));
+        let node = create_test_node("io-engine-1", true, Some(draining));
+        let collector = NodeStatusCollector::new(Some(node));
+
+        let metrics = collector.collect();
+
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_draining", "io-engine-1"),
+            Some(1.0)
+        );
+        // Draining implies cordoned.
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_cordoned", "io-engine-1"),
+            Some(1.0)
+        );
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_online", "io-engine-1"),
+            Some(1.0)
+        );
+    }
+
+    #[test]
+    fn test_collector_state_transitions() {
+        // Online -> Cordoned -> Draining -> Drained -> Online.
+        let node = create_test_node("io-engine-1", true, None);
+        let metrics = NodeStatusCollector::new(Some(node)).collect();
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_online", "io-engine-1"),
+            Some(1.0)
+        );
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_cordoned", "io-engine-1"),
+            Some(0.0)
+        );
+
+        // Cordoned.
+        let cordon =
+            CordonDrainState::cordonedstate(CordonedState::new(vec!["maintenance".to_string()]));
+        let node = create_test_node("io-engine-1", true, Some(cordon));
+        let metrics = NodeStatusCollector::new(Some(node)).collect();
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_cordoned", "io-engine-1"),
+            Some(1.0)
+        );
+
+        // Draining.
+        let draining = CordonDrainState::drainingstate(DrainState::new(
+            vec!["maintenance".to_string()],
+            vec!["drain-volumes".to_string()],
+        ));
+        let node = create_test_node("io-engine-1", true, Some(draining));
+        let metrics = NodeStatusCollector::new(Some(node)).collect();
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_draining", "io-engine-1"),
+            Some(1.0)
+        );
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_cordoned", "io-engine-1"),
+            Some(1.0)
+        );
+
+        // Drained.
+        let drained = CordonDrainState::drainedstate(DrainState::new(
+            vec!["maintenance".to_string()],
+            vec!["drain-volumes".to_string()],
+        ));
+        let node = create_test_node("io-engine-1", true, Some(drained));
+        let metrics = NodeStatusCollector::new(Some(node)).collect();
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_draining", "io-engine-1"),
+            Some(1.0)
+        );
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_cordoned", "io-engine-1"),
+            Some(1.0)
+        );
+
+        // Back to online.
+        let node = create_test_node("io-engine-1", true, None);
+        let metrics = NodeStatusCollector::new(Some(node)).collect();
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_online", "io-engine-1"),
+            Some(1.0)
+        );
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_cordoned", "io-engine-1"),
+            Some(0.0)
+        );
+        assert_eq!(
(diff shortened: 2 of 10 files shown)


metrics-exporter/src/bin/io_engine/node_status/client.rs
```diff
@@ -0,0 +1,368 @@
...
+            .pool_idle_timeout(Some(Duration::from_secs(90)))
+            .pool_max_idle_per_host(2)
+            .build()
+            .context("Failed to build HTTP client")?;
+
+        Ok(Self { client, base_url })
+    }
+
+    /// Fetch all nodes from the control-plane REST API
+    pub(crate) async fn fetch_nodes(&self) -> Result<Vec<Node>> {
```
tiagolobocastro: Shouldn't we fetch only the node status for the current node, rather than all nodes?

metrics-exporter/src/bin/io_engine/node_status/client.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        debug!("Successfully fetched {} nodes", nodes.len());
+
+        Ok(nodes)
+    }
+
+}
+
+#[cfg(test)]
+mod tests {
```
tiagolobocastro: I'm not sure about the usefulness of these tests, it's testing the rest api which is already tested

metrics-exporter/src/bin/io_engine/node_status/types.rs
```diff
@@ -0,0 +1,101 @@
+use serde::{Deserialize, Serialize};
+
+/// Node ID type
+pub(crate) type NodeId = String;
+
+/// Node status enum
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub(crate) enum NodeStatus {
```
tiagolobocastro: Could we use the existing types rather than duplicating here?

metrics-exporter/src/bin/io_engine/main.rs
```diff
@@ -60,15 +70,45 @@ pub(crate) struct Cli {
     /// Use ANSI colors for the logs.
     #[clap(long, default_value_t = true, action = clap::ArgAction::Set)]
     ansi_colors: bool,
+
+    /// REST endpoint for control-plane API (for node status metrics)
+    #[clap(long)]
+    rest_endpoint: Option<String>,
```
tiagolobocastro:
```suggestion
    rest_endpoint: Option<tonic::transport::Uri>,
```

metrics-exporter/src/bin/io_engine/main.rs
```diff
@@ -60,15 +70,45 @@ pub(crate) struct Cli {
...
+
+    /// Polling interval for node status metrics (e.g., "15s", "1m")
+    #[clap(long, default_value = "15s")]
+    polling_interval: String,
```
tiagolobocastro: From the naming it's not clear this is referring to nodes only

metrics-exporter/src/bin/io_engine/main.rs
```diff
@@ -60,15 +70,45 @@ pub(crate) struct Cli {
...
 }
 
 impl Cli {
     fn args() -> Self {
-        Cli::parse()
+        let mut cli = Cli::parse();
+
+        // Override with environment variables if present
+        if cli.rest_endpoint.is_none() {
+            cli.rest_endpoint = env::var("MAYASTOR_REST_ENDPOINT").ok();
```
tiagolobocastro: this can be added directly via CLAP

metrics-exporter/src/bin/io_engine/main.rs
```diff
@@ -60,15 +70,45 @@ pub(crate) struct Cli {
...
+        cli
     }
 }
 
 static GRPC_CLIENT: OnceCell<GrpcClient> = OnceCell::new();
+static NODE_STATUS_CACHE: OnceCell<Arc<RwLock<Vec<Node>>>> = OnceCell::new();
+/// Timestamp of last successful node status fetch (for staleness detection)
+static NODE_STATUS_LAST_FETCH: OnceCell<Arc<RwLock<Option<Instant>>>> = OnceCell::new();
+/// Counter for node status fetch errors
+static NODE_STATUS_FETCH_ERRORS: OnceCell<Arc<AtomicU64>> = OnceCell::new();
```
tiagolobocastro: Can we use a single `OnceCell`, and add all the fields there?

metrics-exporter/src/bin/io_engine/main.rs
```diff
@@ -60,15 +70,45 @@ pub(crate) struct Cli {
...
+    #[clap(long)]
+    rest_endpoint: Option<String>,
+
+    /// Polling interval for node status metrics (e.g., "15s", "1m")
+    #[clap(long, default_value = "15s")]
+    polling_interval: String,
+
+    /// Timeout for REST requests (e.g., "10s")
+    #[clap(long, default_value = "10s")]
+    scrape_timeout: String,
```
tiagolobocastro: You can define the type directly here as well

```suggestion
    scrape_timeout: humantime::Duration,
```

metrics-exporter/Cargo.toml
```diff
@@ -23,3 +23,16 @@ tonic = { workspace = true }
 serde = { workspace = true }
 mime = { workspace = true }
 tracing = { workspace = true }
+
+# OEP-4111: Node status metrics dependencies
+reqwest = { version = "0.12", features = ["json", "rustls-tls"], default-features = false }
+anyhow = "1.0"
+rand = "0.8"
+serde_json = "1.0"
+humantime = "2.1.0"
```
tiagolobocastro: Most of these are already in the workspace, you can pull from there
See imports above

metrics-exporter/src/bin/io_engine/main.rs
```diff
@@ -77,6 +117,74 @@ pub(crate) fn grpc_client<'a>() -> &'a GrpcClient {
...
+    last_fetch: Arc<RwLock<Option<Instant>>>,
+    fetch_errors: Arc<AtomicU64>,
+    interval: Duration,
+) {
+    tokio::spawn(async move {
+        loop {
+            // Fetch FIRST (immediate fetch on startup and after each interval)
+            info!("Polling node status from control-plane REST API");
+
+            match client.fetch_nodes().await {
```
tiagolobocastro: I had also asked on the OEP
Is the cache only the when the rest-api is offline, or always used?

metrics-exporter/src/bin/io_engine/node_status/client.rs
IyanekiB: That makes sense. I originally fetched all nodes to keep the client simple and leave room for future cluster-wide use cases. Though given that the exporter runs as a DaemonSet and already has `MY_NODE_NAME`, each pod really only needs its own node status. I’ll switch to `GET /v0/nodes/{node_id}`, rename `fetch_nodes()` to `fetch_node(&str)`, and return a single `Node `instead of `Vec<Node>`. Does that approach work?

metrics-exporter/src/bin/io_engine/node_status/client.rs
IyanekiB: I added the broader tests to validate deserialization and error handling without relying on a live cluster, since we’re mapping JSON into our own types. I thought they still added value by exercising the client’s behavior rather than the API itself, but I could be wrong and I’m happy to adjust. You’re right about `test_rapid_state_transitions` that belongs in the collector layer, not the HTTP client, so I’ll remove it. Let me know if you’d like me to trim this down further/remove it.

metrics-exporter/src/bin/io_engine/main.rs
IyanekiB: I initially used `String `to keep CLI parsing simple, but I’m not fully familiar with what switching to `tonic::transport::Uri` changes here. My understanding is that it gives us earlier validation and a fail-fast error before the client runs, is that correct?

metrics-exporter/src/bin/io_engine/main.rs
IyanekiB: I split them originally so each field could be initialized independently, but since they’re always set and accessed together, a single `OnceCell` makes more sense. I’ll bundle them into a `NodeStatusState` and have the accessors delegate through it.

metrics-exporter/src/bin/io_engine/main.rs
IyanekiB: The cache is always used, not only when the REST API is offline. Per the [OEP design](https://github.com/IyanekiB/oep-mayastor-node-status-metrics/blob/feat/oep-4111-node-status-metrics/designs/replicated-pv/mayastor/mayastor-node-status-metrics-via-metrics-exporter.md), scrapes read from the in-memory cache to decouple latency from REST calls, and it also serves last-known state when the API is unavailable. Right now, staleness and failures are exposed via metrics rather than returning empty results, I wasn’t fully sure what the best practice is here, so let me know if you’d prefer a different behavior (e.g., empty results instead). I’ll also add a short comment in the polling loop to clarify the intent.

abhilashshetty04:
The current metrics-exporter implementation fetches stats/states from io-engine when it gets queries by the Prometheus on `/metrics`? Cant we use the the same procedure. Prometheus stores the data in time series manner. We will introduce incorrect states w.r.t time if we don't fetch when polled by Prometheus.

We use to have a separate polling period but not anymore. Its in-line with prometheus scraping.  Please refer

https://github.com/openebs/mayastor-extensions/blob/develop/metrics-exporter/src/bin/io_engine/serve/handler.rs#L18

metrics-exporter/src/bin/io_engine/node_status/client.rs
tiagolobocastro: That sounds good

metrics-exporter/src/bin/io_engine/main.rs
tiagolobocastro: Yes correct, it'll do validation by clap itself.
If you add it as I suggested, it should also be simple and "just work".

metrics-exporter/src/bin/io_engine/main.rs
tiagolobocastro: There's a comment from Abhilash [below](https://github.com/openebs/mayastor-extensions/pull/816#issuecomment-3918945776).

pjgranieri:
> The current metrics-exporter implementation fetches stats/states from io-engine when it gets queries by the Prometheus on `/metrics`? Cant we use the the same procedure. Prometheus stores the data in time series manner. We will introduce incorrect states w.r.t time if we don't fetch when polled by Prometheus.
> 
> We use to have a separate polling period but not anymore. Its in-line with prometheus scraping. Please refer
> 
> https://github.com/openebs/mayastor-extensions/blob/develop/metrics-exporter/src/bin/io_engine/serve/handler.rs#L18

We absolutely could, we are aligned with that approach. The inline fetch at handler.rs:18 (store_resource_data(grpc_client()).await) is the pattern we will be following. Node status is going to be fetched the way via an inline REST call in metrics_handler() right alongside the existing gRPC fetch, so values will always reflect the actual state at scrape time:

 store_resource_data(grpc_client()).await;                   // existing gRPC fetch
let node = fetch_node(&node_id, &client).await.ok();        // new inline REST fetch
let node_status_collector = NodeStatusCollector::new(node);  // takes Option<Node>

The background poller, cache, and the polling-specific metrics are all going to be removed. Thank you for the pointer, let us know if you agree that we are working in the right direction with these ideas.

IyanekiB:
Hi everyone, just wanted to follow up on the review feedback. The background poller, cache, and polling-specific metrics have been removed; node status is now fetched inline at scrape time in `metrics_handler()`, which should be consistent with the existing pattern at [`handler.rs:18`](https://github.com/openebs/mayastor-extensions/blob/develop/metrics-exporter/src/bin/io_engine/serve/handler.rs#L18). The custom `Node`/`NodeSpec`/`NodeState` types in `types.rs` (now deleted) have been replaced with `openapi::models` equivalents, and the CLI has been cleaned up (`polling_interval` removed, `rest_endpoint` typed as `Uri`, `scrape_timeout` as `humantime::Duration`). Would appreciate a re-review when you get a chance.

metrics-exporter/src/bin/io_engine/collector/node_status.rs
```diff
@@ -0,0 +1,328 @@
...
+fn is_online(node: &Node) -> bool {
+    node.state
+        .as_ref()
+        .map(|s| matches!(s.status, NodeStatus::Online))
+        .unwrap_or(false)
+}
+
+/// Returns true if the node is cordoned.
+fn is_cordoned(node: &Node) -> bool {
+    node.spec
```
tiagolobocastro: might want to also check is_draining, since drain has implicit cordon, ex:

```suggestion
    is_draining(node) || node.spec
```

metrics-exporter/src/bin/io_engine/collector/node_status.rs
```diff
@@ -0,0 +1,328 @@
...
+    core::{Collector, Desc},
+    proto, GaugeVec, Opts,
+};
+
+/// Returns true if the node is online.
+fn is_online(node: &Node) -> bool {
+    node.state
+        .as_ref()
+        .map(|s| matches!(s.status, NodeStatus::Online))
+        .unwrap_or(false)
```
tiagolobocastro, line 12: I think you can do this now (might require a rebase)

```suggestion
    node.status
        .as_ref()
        .map(|s| matches!(s, NodeStatus::Online))
        .unwrap_or(false)
```

metrics-exporter/src/bin/io_engine/collector/node_status.rs
```diff
@@ -0,0 +1,328 @@
...
+        .map(|cds| {
+            matches!(
+                cds,
+                CordonDrainState::drainingstate(_) | CordonDrainState::drainedstate(_)
+            )
+        })
+        .unwrap_or(false)
+}
+
+/// Collector for node status metrics
```
tiagolobocastro: nit: end all comments with `.` (just to keep up existing convention here)

```suggestion
/// Collector for node status metrics.
```

metrics-exporter/src/bin/io_engine/collector/node_status.rs
```diff
@@ -0,0 +1,328 @@
...
+mod tests {
+    use super::*;
+    use openapi::models::Node;
+    use prometheus::core::Collector;
+
+    /// Create a test node using JSON deserialization to avoid brittle struct construction
+    fn create_test_node(id: &str, online: bool, cordoned: bool, draining: bool) -> Node {
+        let status_str = if online { "Online" } else { "Offline" };
+        let mut spec_json = serde_json::json!({
+            "grpcEndpoint": format!("http://{}:10124", id),
```
tiagolobocastro:
```suggestion
            "grpcEndpoint": format!("http://{id}:10124"),
```

metrics-exporter/src/bin/io_engine/collector/node_status.rs
```diff
@@ -0,0 +1,328 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use openapi::models::Node;
+    use prometheus::core::Collector;
+
+    /// Create a test node using JSON deserialization to avoid brittle struct construction
+    fn create_test_node(id: &str, online: bool, cordoned: bool, draining: bool) -> Node {
+        let status_str = if online { "Online" } else { "Offline" };
+        let mut spec_json = serde_json::json!({
```
tiagolobocastro: Can you not just build it with the Node type itself?

metrics-exporter/src/bin/io_engine/collector/node_status.rs
```diff
@@ -0,0 +1,328 @@
...
+        );
+        assert_eq!(
+            get_metric_value(&metrics, "mayastor_node_online", "io-engine-1"),
+            Some(1.0)
+        );
+    }
+
+    #[test]
+    fn test_collector_state_transitions() {
+        // Online → Cordoned → Draining → Online
```
tiagolobocastro: `Drained` as well?

metrics-exporter/src/bin/io_engine/node_status/client.rs
```diff
@@ -0,0 +1,165 @@
...
+#[derive(Clone, Debug)]
+pub(crate) struct NodeStatusClient {
+    client: Client,
+    base_url: String,
+}
+
+impl NodeStatusClient {
+    /// Create a new NodeStatusClient
+    pub(crate) fn new(base_url: String, timeout: Duration) -> Result<Self> {
+        let client = Client::builder()
```
tiagolobocastro: we have a client bundled in the `openapi` crate itself, we don't need reqwest here?

metrics-exporter/src/bin/io_engine/node_status/client.rs
tiagolobocastro: > we’re mapping JSON into our own types

I think this where we should not be doing that, the openapi crate is fully inclusive so there should be no need to be hand crafting json into the types
I think this is a problem we can remove simply by using the type system

metrics-exporter/src/bin/io_engine/main.rs
```diff
@@ -60,15 +64,19 @@ pub(crate) struct Cli {
...
     ansi_colors: bool,
-}
 
-impl Cli {
-    fn args() -> Self {
-        Cli::parse()
-    }
+    /// REST endpoint for control-plane API (for node status metrics).
+    #[clap(long, env = "MAYASTOR_REST_ENDPOINT")]
+    rest_endpoint: Option<tonic::transport::Uri>,
```
tiagolobocastro, line 70: these need to be set on the helm template as well?

metrics-exporter/src/bin/io_engine/collector/node_status.rs
IyanekiB: `is_cordoned` should now delegate to `is_draining` first to capture the implicit cordon from both draining and drained states

metrics-exporter/src/bin/io_engine/collector/node_status.rs
IyanekiB: The generated `Node` module in our current pinned control-plane dependency only exposes `id`, `spec`, and `state `and no top-level `status` field. So `node.status` might not compile as is. Would this require a rebase against a specific control-plane change, and if so could you point us to it?

metrics-exporter/src/bin/io_engine/main.rs
IyanekiB: `--rest-endpoint` is already in the daemonset template pointing at `(--rest-endpoint=http://{{  .Release.Name  }}-api-rest:8081`. Is that the right endpoint, or did you have something else in mind?

metrics-exporter/src/bin/io_engine/collector/node_status.rs
IyanekiB: `create_test_node` now uses `NodeSpec::new_all`, `NodeState::new_all`, and the `CordonDrainState` constructors directly. Is this approach reasonable?

metrics-exporter/src/bin/io_engine/node_status/client.rs
IyanekiB: That makes sense, we've now switched to the `ApiCLient `and `Configuration::builder()` from the openapi crate directly

metrics-exporter/src/bin/io_engine/collector/node_status.rs
tiagolobocastro: Yeah probably a rebase, but it's fine as it's now as well

metrics-exporter/src/bin/io_engine/node_status/client.rs
```diff
@@ -0,0 +1,44 @@
...
+        let client = ApiClient::new(config);
+        Ok(Self { client })
+    }
+
+    /// Fetch a single node from the control-plane REST API.
+    pub(crate) async fn fetch_node(
+        &self,
+        node_id: &str,
+    ) -> Result<Node, Error<RestJsonError>> {
+        debug!("Fetching node {node_id} from control-plane");
```
tiagolobocastro: nit: maybe these should be trace, since this will happen very often?
```suggestion
        trace!("Fetching node {node_id} from control-plane");
```

metrics-exporter/src/bin/io_engine/main.rs
tiagolobocastro: no sorry I had missed that

metrics-exporter/src/bin/io_engine/main.rs
```diff
@@ -94,6 +114,29 @@ async fn main() -> Result<(), ExporterError> {
...
+    // Store node name for use by the metrics handler on every scrape.
+    NODE_NAME
+        .set(get_node_name()?)
+        .expect("Node name should be initialised only once");
+
+    // Initialize node status REST client if endpoint is configured.
+    if let Some(ref endpoint) = args.rest_endpoint {
+        info!(
+            "Initializing node status REST client with endpoint: {}",
+            endpoint
```
tiagolobocastro:
```suggestion
            "Initializing node status REST client with endpoint: {endpoint}"
           
```

tiagolobocastro (commented):
LGTM

@IyanekiB could you please squash your commits?

IyanekiB:
> LGTM
> 
> @IyanekiB could you please squash your commits?

@tiagolobocastro just finished squashing the commits and applied the changes you suggested. Would like a re-review if you don't mind. Thanks!

tiagolobocastro (approved):
thanks for contributing this @IyanekiB !

IyanekiB:
> thanks for contributing this @IyanekiB !

Thanks @tiagolobocastro! Just wanted to ask, is there anything else needed for the OEP that we should finalize?

tiagolobocastro:
> > thanks for contributing this @IyanekiB !
> 
> Thanks @tiagolobocastro! Just wanted to ask, is there anything else needed for the OEP that we should finalize?

Nothing I can think of, just need a few more reviews here, CC @abhilashshetty04 @niladrih @Abhinandan-Purkait

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
@pchandra19 any clues why this is stuck?

tiagolobocastro:
bors merge

pchandra19:
bors merge

pchandra19:
bors merge

pchandra19:
bors try

tiagolobocastro:
bors cancel

tiagolobocastro:
bors ping

tiagolobocastro:
bors merge

tiagolobocastro:
@IyanekiB, sorry about the delay, we had some CI issues with bors.
There's some test errors, I think because a new status was added to node, would you be able to fix that? Thanks

IyanekiB:
> @IyanekiB, sorry about the delay, we had some CI issues with bors. There's some test errors, I think because a new status was added to node, would you be able to fix that? Thanks

No worries @tiagolobocastro! Yeah we can take a look at that.

IyanekiB:
@tiagolobocastro as you pointed out, the upstream introduced a `status` field to `Node` and `shutdown` field to `NodeSpec` (see [node_status.rs#L149](https://github.com/IyanekiB/mayastor-extensions/commit/040ae97b0272f88b6c3ce624191580286ba4e3cd#diff-913d8ca9953bc95d11da3a357458877b8d4adbd9f67afad01ffc1b683f672a0bR149) and [node_status.rs#L157](https://github.com/IyanekiB/mayastor-extensions/commit/040ae97b0272f88b6c3ce624191580286ba4e3cd#diff-913d8ca9953bc95d11da3a357458877b8d4adbd9f67afad01ffc1b683f672a0bR157) for fixes), which broke our test helper. The issue should be resolved now and I have squashed the related commits.

PR #831, opened 2026-04-01
feat: add global helm global values

### LINKED ISSUE #4171: preUpgradeHook and .mayastor.preUpgradeHook imagePullSecrets key is not uniform with rest of chart
The helm value for setting the ImagePullSecrets for the preUpgradeHook and the mayastor preUpgradeHook uses a different syntax than what the .mayastor.image.pullSecrets key uses. They should be homogenised.

krishnaGajabi:
Resolves https://github.com/openebs/openebs/issues/4171

## Why is this PR required? What issue does it fix?

Currently, the Helm charts do not provide global variables for configuring common parameters such as the image registry, image pull secrets, image pull policy, and analytics settings.

When deploying the OpenEBS Helm charts, users often need to override these values across multiple subcharts and components. This makes it difficult to locate and configure all the required fields consistently. Providing global variables simplifies configuration and improves the user experience.

## What does this PR do?

This PR introduces the following global configuration variables:
```yaml
global:
  # -- Global override for image registry
  imageRegistry: ""
  # -- Global overide for image pull policy
  imagePullPolicy: ""
  # -- Global override for image pull secrets
  # - secret
  imagePullSecrets: []
  analytics:
    # -- Global overide for call home
    enabled:
```

These variables allow users to configure common settings in a single place, which will then be applied across the chart components.

## Does this PR require any upgrade changes?

No.
This change is backward compatible and does not require any upgrade steps.

## Verification

The changes were verified using the following scenarios:

Rendered the templates using helm template to confirm correct value substitution.

Deployed the chart on a Kubernetes cluster with 3 worker nodes (kubeadm-based setup).

Verified that the global values are rendered and applied correctly across the components.

### DIFF
--- a/chart/templates/pre-upgrade-hook.yaml
+++ b/chart/templates/pre-upgrade-hook.yaml
@@ -124,8 +124,8 @@ spec:
             defaultMode: 0777
       containers:
         - name: mayastor-pre-upgrade-hook
-          image: {{ .Values.preUpgradeHook.image.registry }}/{{ .Values.preUpgradeHook.image.repo }}:{{ .Values.preUpgradeHook.image.tag }}
-          imagePullPolicy: {{ .Values.preUpgradeHook.image.pullPolicy }}
+          image: {{ default .Values.preUpgradeHook.image.registry .Values.global.imageRegistry }}/{{ .Values.preUpgradeHook.image.repo }}:{{ .Values.preUpgradeHook.image.tag }}
+          imagePullPolicy: {{ default .Values.preUpgradeHook.image.pullPolicy .Values.global.imagePullPolicy }}
           command:
             - "sh"
             - "-c"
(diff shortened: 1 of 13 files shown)


krishnaGajabi:
CI tests(images.sh script) are falling because we have set global.imageRegistry to an empty string. earlier the value used to be from localpv helm chart where it was explicitly set to docker.io. need the changes to be done in localpv before this changes are merged

chart/templates/pre-upgrade-hook.yaml
```diff
@@ -134,8 +134,9 @@ spec:
           volumeMounts:
             - name: scripts
               mountPath: /scripts
-      {{- if .Values.preUpgradeHook.imagePullSecrets }}
-      imagePullSecrets: {{ toYaml .Values.preUpgradeHook.imagePullSecrets | nindent 8 }}
+      {{- if or .Values.global.imagePullSecrets .Values.preUpgradeHook.imagePullSecrets }}
+      imagePullSecrets:
+{{- include "mayastor.preUpgradeHook.pullSecrets" . | indent 6 }}
```
tiagolobocastro:
```suggestion
{{- include "mayastor.preUpgradeHook.pullSecrets" . | indent 8 }}
```

Should this be 8?

krishnaGajabi: Its a valid yaml, without it too

tiagolobocastro: Yeah, but it's not consistent with the rest

tiagolobocastro:
```
Secrets:
- name
```
vs
```
Secrets:
  - name
```

krishnaGajabi: Done

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

PR #834, opened 2026-04-04
feat: add pool delete plugin command and opts

niladrih:
Changes:
1. Add `kubectl mayastor delete pool` command and purge CLI options.
2. Add `--cleanup-cr` and `--cleanup-cr-timeout` options for cleaning
   up DiskPool reasources after pool delete.


k8s/plugin/src/resources/mod.rs
```diff
@@ -321,3 +396,45 @@ pub async fn init_rest(cli_args: &CliArgs) -> Result<(), Error> {
...
+                "Timed out after {} waiting for DiskPool CR {pool_id} deletion \
+                 in namespace {namespace}. Check that the pool operator is running.",
+                humantime::format_duration(timeout)
+            )
+        })?
+        .map_err(|e| anyhow!("{e}"))?;
+
+    match result {
+        CrDeleteOutcome::Deleted => {
+            println!("DiskPool CR {pool_id} deleted from namespace {namespace}");
```
tiagolobocastro: How does this work if output is json/yaml?
IIRC the existing deletes don't print anything in these cases, we might want to the same here for consistency

k8s/plugin/Cargo.toml
```diff
@@ -27,9 +27,13 @@ shutdown = { path = "../../dependencies/control-plane/utils/shutdown" }
 
 kube = { workspace = true, features = ["derive", "runtime"] }
 k8s-openapi = { workspace = true }
+k8s-operators = { path = "../../dependencies/control-plane/k8s/operators", default-features = false }
 kube-proxy = { path = "../../dependencies/control-plane/k8s/proxy" }
 kube-forward = { path = "../../dependencies/control-plane/k8s/forward" }
-tokio = { workspace = true }
+tokio = { workspace = true, features = ["time"] }
```
tiagolobocastro, line 33: what is this for?

k8s/plugin/src/resources/diskpool_cleanup.rs
```diff
@@ -0,0 +1,106 @@
...
+pub async fn delete_diskpool_cr(
+    client: Client,
+    namespace: &str,
+    pool_id: &str,
+) -> Result<CrDeleteOutcome, CleanupError> {
+    let api: Api<DiskPool> = Api::namespaced(client, namespace);
+
+    // Fetch the CR to get its UID before deleting — is_deleted needs the
+    // UID to distinguish "gone" from "deleted and recreated".
+    let uid = match api.get(pool_id).await {
```
tiagolobocastro: you can use get_opt: https://github.com/openebs/mayastor-extensions/blob/81f769fe34db6fefadf1632578a0e476a58756a7/k8s/plugin/src/resources/mod.rs#L211

k8s/plugin/src/resources/diskpool_cleanup.rs
```diff
@@ -0,0 +1,106 @@
...
+        Ok(cr) => cr.uid().ok_or_else(|| CleanupError::MissingUid {
+            pool_id: pool_id.into(),
+            namespace: namespace.into(),
+        })?,
+        Err(kube::Error::Api(resp)) if resp.code == StatusCode::NOT_FOUND => {
+            return Ok(CrDeleteOutcome::NotFound);
```
tiagolobocastro: Do we really need this enum just for this, you can follow kube-rs and return Result<Option, _> ?

Cargo.toml
```diff
@@ -35,6 +35,7 @@ serde_yaml = "0.9.34"
 # errors
 snafu = "0.8.5"
 anyhow = "1.0.92"
+thiserror = "1.0.68"
```
tiagolobocastro: We already use snafu here..

k8s/plugin/Cargo.toml
niladrih: We wrap a diskpool watcher with tokio timeout and wait for it to get deleted. There is a timeout in the watch options, but that is for a server side timeout. Once that expires, our client side watcher reopens the server side watch automatically. Setting that doesn't signify 'give up after this duration'. That's what we're using time for.

Cargo.toml
niladrih: Using snafu for the diskpool_cleanup module's errors.

k8s/plugin/src/resources/diskpool_cleanup.rs
niladrih: Removed the enum.

k8s/plugin/src/resources/diskpool_cleanup.rs
niladrih: Using get_opt.

k8s/plugin/src/resources/mod.rs
niladrih: Removed the logs for successful deletes. Not logging errors for NOT_FOUND (doesn't necessarily mean a non-zero exit) if `--ignore-not-found` is supplied. If `--cleanup-cr` is used and both of the Mayastor Pool and the DiskPool CR are missing, it's an error (non-zero exit), unless --ignore-nor-found is given. If one is missing and not both and `--cleanup-cr` is given, it's an error log, but successful exit.

niladrih: Here's a matrix with the behaviour.

pool spec | CR | --ignore-not-found | result
-- | -- | -- | --
deleted | deleted | — | Ok (silent)
deleted | missing | no | Ok + stderr warning
deleted | missing | yes | Ok (silent)
missing | deleted | no | Ok + stderr warning
missing | deleted | yes | Ok (silent)
missing | missing | no | error
missing | missing | yes | Ok (silent)

k8s/plugin/src/resources/diskpool_cleanup.rs
```diff
@@ -0,0 +1,100 @@
...
+/// **This function does not enforce a timeout.**  The caller should wrap
+/// it with `tokio::time::timeout` or similar.
+pub async fn delete_diskpool_cr(
+    client: Client,
+    namespace: &str,
+    pool_id: &str,
+) -> Result<Option<()>, CleanupError> {
+    let api: Api<DiskPool> = Api::namespaced(client, namespace);
+
+    // Fetch the CR to get its UID before deleting — is_deleted needs the
```
tiagolobocastro: Not needed, delete api itself returns Either::left with the cr, if it's deleting

k8s/plugin/src/resources/mod.rs
```diff
@@ -204,6 +230,49 @@ impl ExecuteOperation for Operations {
...
+
+                        // Confirm deletion.
+                        plugin::resources::confirm(
+                            "Are you sure you want to delete the resource?",
+                            "This operation cannot be undone!",
+                            args.yes,
+                        )?;
+
+                        // REST delete via control-plane.
+                        let rest_result = pool::Pool::del(
```
tiagolobocastro: Why are we not using the `plugin::resources::DeleteArg` like for volome and snapshot? That also would include the question above

k8s/plugin/src/resources/diskpool_cleanup.rs
niladrih: Made this change.

k8s/plugin/src/resources/mod.rs
niladrih: Using DeleteArg pattern.

k8s/plugin/Cargo.toml
```diff
@@ -17,7 +17,7 @@ tls = ["openapi/tower-client-tls", "rest-plugin/tls"]
 
 
 [dependencies]
-openapi = { path = "../../dependencies/control-plane/openapi", default-features = false, features = ["tower-trace"] }
+openapi = { path = "../../dependencies/control-plane/openapi", default-features = false, features = ["tower-trace"], optional = true }
```
tiagolobocastro: why is this optional now?

k8s/plugin/src/resources/mod.rs
```diff
@@ -204,6 +229,42 @@ impl ExecuteOperation for Operations {
...
+                            ),
+                        }
+                        .execute(cli_args)
+                        .await?
+                    }
+                    DeleteResources::Pool(pool_args) => {
+                        // --cleanup-cr: run REST delete (never swallow 404), then
+                        // CR cleanup.
+                        let rest_result = plugin::resources::DeleteArgs {
+                            ignore_not_found: false,
```
tiagolobocastro, line 248: why is this hardcoded to false?

niladrih: We use the NOT_FOUND error to decide if we have to log the already missing warning.
We handle the ignore_not_found ourselves. So not passing it on to the REST.

k8s/plugin/Cargo.toml
niladrih: optional was set to true on the k8s-operators crate

niladrih: Going with a new feature on the Cargo.toml `openapi = []`, as some of the code is set to compile for that specific feature. `optional = true` is a hack which achieves a similar result.

k8s/plugin/src/resources/mod.rs
niladrih: I've removed the logging for when one of -- Mayastor Pool and DiskPool CR are deleted.

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

PR #843, opened 2026-04-22
fix(upgrade): add RBAC for OpenShift resources

niladrih:
Adds create, delete, patch, list, get RBAC permissions for all security.openshift.io resources (OpenShift api group).

### DIFF
--- a/k8s/upgrade/src/plugin/objects.rs
+++ b/k8s/upgrade/src/plugin/objects.rs
@@ -237,6 +237,12 @@ pub(crate) fn upgrade_job_cluster_role(
                 verbs: vec!["create", "list", "delete", "get", "patch"].into_vec(),
                 ..Default::default()
             },
+            PolicyRule {
+                api_groups: Some(vec!["security.openshift.io"].into_vec()),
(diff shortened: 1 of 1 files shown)


k8s/upgrade/src/plugin/objects.rs
```diff
@@ -237,6 +237,12 @@ pub(crate) fn upgrade_job_cluster_role(
                 verbs: vec!["create", "list", "delete", "get", "patch"].into_vec(),
                 ..Default::default()
             },
+            PolicyRule {
+                api_groups: Some(vec!["security.openshift.io"].into_vec()),
+                resources: Some(vec!["*"].into_vec()),
```
Abhinandan-Purkait: Let's trim it down as we discussed on the other PR.

Abhinandan-Purkait (approved):

k8s/upgrade/src/plugin/objects.rs
niladrih: Thinned it down. It just needed the resource securitycontextconstraints.

PR #851, opened 2026-05-09
fix(upgrade): respect plugin kube context instead of trying default

niladrih:
The upgrade plugin creates its own kube clients, doesn't have a unified context. Also it doesn't accept the kubectl mayastor plugin level context. This change fixes that.

### DIFF
--- a/k8s/plugin/src/resources/mod.rs
+++ b/k8s/plugin/src/resources/mod.rs
@@ -212,7 +212,8 @@ impl ExecuteOperation for Operations {
                 GetResourcesK8s::Rest(resource) => resource.execute(cli_args).await?,
                 GetResourcesK8s::UpgradeStatus(resources) => {
                     // todo: use generic execute trait
-                    resources.get_upgrade(&cli_args.namespace).await?
+                    let client = cli_args.client().await?;
+                    resources.get_upgrade(&cli_args.namespace, &client).await?
                 }
             },
             Operations::Drain(resource) => resource.execute(cli_args).await?,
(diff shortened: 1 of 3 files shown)


k8s/upgrade/src/plugin/error.rs
```diff
@@ -128,7 +128,7 @@ pub enum Error {
         source
     ))]
     ListPodsWithLabel {
-        source: kube::Error,
+        source: crate::common::error::Error,
```
tiagolobocastro: Why are these types changing?

niladrih: I'm using the paginated_list and paginated_list_metadata functions from upgrade. That function wraps kube::Error, via snafu. This error again wraps that snafu Error.

tiagolobocastro: Do we need that, are we going to have so many pods that require pagination?

Abhinandan-Purkait (approved):

k8s/upgrade/src/plugin/error.rs
niladrih: No we probably don't need that. Do you think I should remove the use of these paginated list APIs?

tiagolobocastro: Yeah, I'd say let's keep as is unless there's a need

niladrih: Done.

tiagolobocastro (approved):

PR #860, opened 2026-05-28
feat: add pool_name label to replica

cmrmahesh:
## Description

Add `pool_name` label to all replica I/O metrics (`replica_bytes_read`, `replica_num_read_ops`, `replica_bytes_written`, `replica_num_write_ops`, `replica_read_latency_us`, `replica_write_latency_us`).

The metrics-exporter now calls `ListReplicas` gRPC to build a replica name → pool_name mapping and includes it as a label when emitting replica metrics. The map is only refreshed when a new replica appears that isn't already cached, so there's zero additional gRPC overhead during steady-state operation.

## Motivation and Context

Described at [#1990](https://github.com/openebs/mayastor/issues/1990)

Replica I/O metrics currently have no `pool_name` label. The only shared dimension between replica metrics and disk pool metrics (`diskpool_*`) is `node`, which means correlating a specific replica's I/O to its backing pool requires visual inference ("same node = same pool") — this breaks on nodes with multiple disk pools.

With this change, users can directly query:
```promql
rate(replica_num_read_ops{pool_name="pool-1"}[5m])

```

or join with pool metrics:

`replica_num_read_ops * on(pool_name, node) group_left() diskpool_total_size_bytes`

**Regression**
No

How Has This Been Tested?

- [x] Verified that the Replica gRPC message already exposes poolname (field 8 in replica.proto)
- [x] Confirmed ListReplicaOptions::default() returns all replicas on the node
- [x] Validated that the cache-miss-only refresh logic avoids unnecessary gRPC calls in steady state
- [x] Existing metric labels (node, name, pv_name) remain unchanged; pool_name is additive
- [ ] Deploy the changes and validate on an actual cluster 

Types of changes

Checklist:

### DIFF
--- a/metrics-exporter/src/bin/io_engine/cache/mod.rs
+++ b/metrics-exporter/src/bin/io_engine/cache/mod.rs
@@ -8,7 +8,7 @@ use crate::client::{
     replica_stat::ReplicaIoStats,
 };
 use once_cell::sync::OnceCell;
-use std::sync::Mutex;
+use std::{collections::HashMap, sync::Mutex};
 
 /// NOTE: try to reference cache from the Collector.
 static CACHE: OnceCell<Mutex<Cache>> = OnceCell::new();
@@ -36,6 +36,8 @@ pub(crate) struct Data {
     nexus_stats: NexusIoStats,
     /// Contains Replica IOStats data.
     replica_stats: ReplicaIoStats,
+    /// Maps replica name to pool_name.
+    replica_pool_map: HashMap<String, String>,
 }
 
 impl Cache {
@@ -88,6 +90,16 @@ impl Cache {
     pub(crate) fn replica_iostat_mut(&mut self) -> &mut ReplicaIoStats {
         &mut self.data.replica_stats
     }
+
+    /// Get a reference to replica_pool_map.
+    pub(crate) fn replica_pool_map(&self) -> &HashMap<String, String> {
+        &self.data.replica_pool_map
+    }
+
+    /// Get mutable reference to replica_pool_map.
+    pub(crate) fn replica_pool_map_mut(&mut self) -> &mut HashMap<String, String> {
+        &mut self.data.replica_pool_map
+    }
 }
 
 impl Default for Data {
@@ -108,6 +120,7 @@ impl Data {
             replica_stats: ReplicaIoStats {
                 replica_stats: vec![],
             },
+            replica_pool_map: HashMap::new(),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


cmrmahesh:
I went with a cache-miss approach here — 
the map only refreshes when we see a replica in the stats that isn't already mapped. A replica doesn't migrate between pools; if one fails, it's destroyed and a new replica (with a new identity) is created on another pool. That new replica naturally triggers a cache miss and a refresh. In steady state this means zero additional gRPC calls.

metrics-exporter/src/bin/io_engine/client/grpc_client.rs
```diff
@@ -200,4 +203,23 @@ impl GrpcClient {
         }?;
         Ok(ReplicaIoStats { replica_stats })
     }
+
+    /// Lists all replicas to build a replica name → pool_name mapping.
+    pub(crate) async fn list_replicas(
```
abhilashshetty04:
```suggestion
    pub(crate) async fn fetch_replica_pool_mapping(
```

metrics-exporter/src/bin/io_engine/cache/mod.rs
```diff
@@ -118,4 +131,37 @@ pub(crate) async fn store_resource_data(client: &GrpcClient) {
     let _ = pool_stat::store_pool_stats_data(client).await;
     let _ = nexus_stat::store_nexus_stats_data(client).await;
     let _ = replica_stat::store_replica_stats_data(client).await;
+    store_replica_pool_map(client).await;
+}
+
+/// Fetches replica list and stores replica name → pool_name mapping in cache.
+/// Only refreshes when new replicas appear that aren't in the existing map.
+async fn store_replica_pool_map(client: &GrpcClient) {
+    let needs_refresh = {
```
abhilashshetty04, line 140: What if replicas get deleted from io-engine? Should we update map then as well?

cmrmahesh: When a replica is deleted, it disappears from the stats response, so its map entry is never looked up — it's a stale and harmless. The map gets fully replaced on the next refresh (triggered when a new replica appears ` *cache.replica_pool_map_mut() = new_map`), which cleans up stale entries. I could add explicit pruning but it adds complexity for no extra benefits. Let me know your thoughts.

metrics-exporter/src/bin/io_engine/client/grpc_client.rs
cmrmahesh: This is done.

tiagolobocastro:
How about adding the pool_name to the replica stat itself?

cmrmahesh:
> How about adding the pool_name to the replica stat itself?

@tiagolobocastro much cleaner to add it to the source rather than this map approach. Let me try that way. We can compare and finalise an approach. 

Edit:
Looks like that will be a multi repo change, I will raise PRs and then we can decide.

abhilashshetty04:
@cmrmahesh This is not needed now after https://github.com/openebs/mayastor/pull/1991#pullrequestreview-4426786492 ?

cmrmahesh:
> @cmrmahesh This is not needed now after [openebs/mayastor#1991 (review)](https://github.com/openebs/mayastor/pull/1991#pullrequestreview-4426786492) ?

Yes this pr is no longer valid.

PR #861, opened 2026-05-29
build: use static CDN to pull crates

Abhinandan-Purkait:
- Use static.crates.io (CDN) instead of crates.io/api to avoid the 1 req/sec rate limit on the API servers, which currently returns intermittent 403s. See https://github.com/rust-lang/crates.io/issues/13482

### DIFF
--- a/nix/lib/rust.nix
+++ b/nix/lib/rust.nix
@@ -27,6 +27,16 @@ rec {
     naersk_package = channel: pkgs.callPackage sources.naersk {
       rustc = channel.stable;
       cargo = channel.stable;
+      # Rewrite crates.io API URLs to static CDN to avoid intermittent 403s
+      fetchurl = attrs@{ url, ... }:
+        let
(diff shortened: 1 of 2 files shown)


Abhinandan-Purkait:
bors try

pchandra19 (approved):

Abhinandan-Purkait:
@mergifyio backport release/2.11

abhilashshetty04 (approved):

nix/lib/rust.nix
```diff
@@ -27,6 +27,16 @@ rec {
     naersk_package = channel: pkgs.callPackage sources.naersk {
       rustc = channel.stable;
       cargo = channel.stable;
+      # Rewrite crates.io API URLs to static CDN to avoid intermittent 403s
```
tiagolobocastro, line 30: Couldn't we have just set `cratesDownloadUrl` ?

PR #873, opened 2026-06-02
fix: release workflow missing nix-setup action

Abhinandan-Purkait:


.github/workflows/release.yml
tiagolobocastro:
```suggestion
          nix-shell --pure --run "./scripts/helm/update-reg-repo.sh --registry docker.io/ --repository ${{ github.repository_owner }}" ./scripts/helm/shell.nix
```

tiagolobocastro (approved):

PR #877, opened 2026-06-03
chore: increase max ingestion limits for loki

Abhinandan-Purkait:
- ETCD pod has 16 labels and loki by default supports maximum of 15 labels. We have this change on openebs to support this but was missing on mayastor. Increasing the max labels to 20. Not increasing too much as it slows down indexing.


niladrih (approved):

niladrih:
@Mergifyio backport develop

PR #879, opened 2026-06-03
feat: push events to ephemeral volume if Loki disabled and stdout

rohan2794:
## Description

- Added support to log events  to a JSON file on the pod's ephemeral volume when Loki is disabled.
- Integrated the Rust clap crate for command-line argument parsing and configuration management.
- Implemented log rotation for JSON event files to prevent unbounded file growth when Loki is not configured.


events-aggregator/src/main.rs
```diff
@@ -19,44 +24,126 @@ enum ProcessingResult {
...
+    nats_url: Url,
+
+    /// Enable JetStream subscription mode.
+    #[arg(long, default_value_t = false, action = clap::ArgAction::Set)]
+    jetstream_enabled: bool,
+
+    /// Optional Loki endpoint URL. If set, events will be sent to Loki;
+    /// otherwise they will be written to a local file.
+    #[arg(long, short)]
+    loki_url: Option<Url>,
```
tiagolobocastro: Why is this necessary, IIRC from last discussion we'd be only logging and let loki collect it?

events-aggregator/src/main.rs
```diff
@@ -19,44 +24,126 @@ enum ProcessingResult {
...
+        batch_max_size = exporter::BATCH_MAX_SIZE,
+        batch_timeout_ms = exporter::BATCH_TIMEOUT_MS,
+        "Exporter batch configuration"
+    );
 
     // Initialize NATS Manager
-    let nats_mgr = NatsManager::new(&nats_url).await?;
+    let nats_mgr = NatsManager::new(cli_args.nats_url.as_str())
+        .await
+        .map_err(|e| anyhow::anyhow!("Failed to connect to NATS: {}", e))?;
```
tiagolobocastro:
```suggestion
        .map_err(|error| anyhow::anyhow!("Failed to connect to NATS: {error}"))?;
```

events-aggregator/src/main.rs
```diff
@@ -19,44 +24,126 @@ enum ProcessingResult {
...
+
+    /// Local events directory for file exporter. Required when `--loki-url` is not provided.
+    #[arg(long, short)]
+    events_dir: Option<String>,
+
+    /// Maximum total size for local event files under `--events-dir`, including rotated history.
+    /// Supports human-readable units such as KiB, MiB, GiB, KB, MB, GB, bytes, and decimal values like 1.5GiB.
+    /// Required when `--loki-url` is not provided.
+    #[arg(long, short, value_parser = parse_dir_size_limit)]
+    dir_size_limit: Option<usize>,
```
tiagolobocastro: `parse_dir_size_limit` seems rather complicated, why not just:

```suggestion
    dir_size_limit: Option<byte_unit::Byte>,
```

events-aggregator/src/main.rs
```diff
@@ -19,44 +24,126 @@ enum ProcessingResult {
...
+impl CliArgs {
+    fn args() -> Self {
+        CliArgs::parse()
+    }
+}
+
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
     // Initialize structured logging with tracing-subscriber
     tracing_subscriber::fmt()
```
tiagolobocastro: Should replace this with utils

events-aggregator/Cargo.toml
```diff
@@ -1,21 +1,29 @@
...
 version = "0.1.0"
-edition = "2024"
+edition = "2021"
 
 [dependencies]
 # Async runtime
 tokio = { workspace = true }
 # NATS client
-async-nats = "0.37"
+async-nats = { version = "0.37.0", features = ["ring"] }
```
tiagolobocastro: Should come from workspace, and also why are you specifying ring?

events-aggregator/Cargo.toml
```diff
@@ -1,21 +1,29 @@
...
 # JSON serialization
 serde = { workspace = true }
 serde_json = { workspace = true }
 # Logging and Observability
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["json"]}
+tracing-appender = "0.2"
 # Error handling
 anyhow = { workspace = true }
+byte-unit = "5.1.4"
```
tiagolobocastro: Btw we already have `parse-size` in the deps, might be worth considering that one to avoid pulling in more dependencies

events-aggregator/src/exporter/engine.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    // The configured size limit allows up to 40% of the limit in `events.json` and
+    // another 40% in `events.1.json`. When `events.json` would exceed 40%, we rotate it.
+    let pending = buffer.len();
+    tracing::info!(action = "prepare_disk_write", dir = ?dir, pending_bytes = pending, size_limit = size_limit, "Preparing to write batch to disk");
+    rotate_file_if_needed(dir, size_limit, pending).await?;
+
+    let dir = dir.to_path_buf();
+    let buffer = buffer.clone();
```
tiagolobocastro: Why is this cloned, we don't seem to reuse it?

events-aggregator/src/exporter/engine.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    Ok(())
+}
+
+async fn rotate_file_if_needed(
+    dir: &Path,
+    size_limit: usize,
+    pending_bytes: usize,
+) -> anyhow::Result<()> {
+    let base_path = dir.join("events.json");
```
tiagolobocastro: Where is this file in a K8s deployment?

events-aggregator/Cargo.toml
rohan2794: Workspace does not have `async-nats` crate and `async-nats` default feature include ring for 0.37.0 version no need for explicit mention.

events-aggregator/src/exporter/engine.rs
rohan2794: Yes, clone not required as it's not reused

events-aggregator/src/exporter/engine.rs
rohan2794: This file will be created when events are persisted into ephemeral volume when Loki is disabled.

events-aggregator/Cargo.toml
rohan2794: replaced` byte-unit` with `parse-size`

events-aggregator/src/main.rs
rohan2794: replaced `byte-unit` with `parse-size`

events-aggregator/src/main.rs
rohan2794: Removed loki _url  args . We will only logging to console and loki will scrape logs.

events-aggregator/src/main.rs
rohan2794: replaced with tracing telemetry

events-aggregator/src/client/nats_client.rs
```diff
@@ -85,14 +93,13 @@ impl NatsManager {
         Ok(())
     }
 
-    // JetStream Consumer Setup with Backpressure Handling
+    // JetStream Consumer Setup with Robust Retry Logic and Backoff
```
tiagolobocastro: End doc comments with a period

```suggestion
    // JetStream Consumer Setup with Robust Retry Logic and Backoff.
```

events-aggregator/src/client/nats_client.rs
```diff
@@ -40,13 +46,14 @@ impl NatsManager {
...
-        info!("Attempting to connect to NATS at {}...", url);
+    /// Establish a connection to the NATS server with strict timeouts and a robust retry strategy.
+    pub async fn new(url: &str) -> Result<Self, async_nats::Error> {
+        info!("Attempting to connect to NATS at {url}...");
 
         // Configure native NATS backoff and retries
         let options = async_nats::ConnectOptions::new()
-            .connection_timeout(Duration::from_secs(1))
-            .request_timeout(Some(Duration::from_secs(5)))
+            .connection_timeout(CONNECTION_TIMEOUT)
```
tiagolobocastro: Should go on the clap to be overridable via cli arg

events-aggregator/src/client/nats_client.rs
```diff
@@ -59,12 +66,13 @@ impl NatsManager {
...
         // the timeouts and backoffs configured above without race conditions.
-        let client = options.connect(url).await.map_err(|e| {
-            anyhow::anyhow!("Failed to connect to NATS after configured attempts: {}", e)
-        })?;
+        let client = options.connect(url).await?;
 
         Ok(Self { client })
     }
+
+    /// Start subscribing to the specified subject. If JetStream is enabled,
```
tiagolobocastro: Shouldn't the consuming part of this be a crate in the control-plane repo?
We can deploy the eventing there so makes sense to be able to consume from there as well, also for aiding debug there

events-aggregator/Cargo.toml
```diff
@@ -1,21 +1,27 @@
...
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["json"]}
+tracing-appender = "0.2"
 # Error handling
 anyhow = { workspace = true }
-colored_json = "4"
+parse-size = { workspace = true }
 futures = { workspace = true }
 rand = { workspace = true }
+tokio-util = { version = "0.7.13", features = ["time"] }
```
tiagolobocastro: Use existing crate versions, also pull from workspace

events-aggregator/src/exporter/engine.rs
```diff
@@ -0,0 +1,276 @@
...
+    let mut buffer = String::new();
+    for event in batch {
+        buffer.push_str(&event.line);
+        buffer.push('\n');
+    }
+
+    // The configured size limit allows up to 40% of the limit in `events.json` and
+    // another 40% in `events.1.json`. When `events.json` would exceed 40%, we rotate it.
+    let pending = buffer.len();
+    tracing::info!(action = "prepare_disk_write", dir = ?dir, pending_bytes = pending, size_limit = size_limit, "Preparing to write batch to disk");
```
tiagolobocastro: Do we Log for every write to disk? Seems a bit much if so

events-aggregator/src/exporter/engine.rs
```diff
@@ -0,0 +1,276 @@
...
+    rotate_file_if_needed(dir, size_limit, pending).await?;
+
+    let dir = dir.to_path_buf();
+
+    tokio::task::spawn_blocking(move || -> anyhow::Result<()> {
+        let appender = RollingFileAppender::new(Rotation::NEVER, dir, EVENTS_JSON_FILE);
```
tiagolobocastro: If we're using tracing infra, can we not just send events and let the tracing library write them out?

events-aggregator/src/main.rs
tiagolobocastro: There's still a usize here

events-aggregator/src/main.rs
```diff
@@ -1,62 +1,139 @@
...
+// JetStream mode toggle, local file export options, and subject filter for subscriptions.
+#[derive(Parser)]
+#[command(name = package_description!(), version = version_info_str!())]
+struct CliArgs {
+    /// NATS server URL.
+    #[arg(long, short, default_value = "nats://mayastor-nats:4222")]
+    nats_url: Url,
+
+    /// Enable JetStream subscription mode.
+    #[arg(long, default_value_t = false, action = clap::ArgAction::Set)]
```
tiagolobocastro, line 34: When is this disabled?

events-aggregator/src/main.rs
```diff
@@ -123,36 +205,56 @@ fn process_message(subject_str: &str, payload_bytes: &[u8]) -> ProcessingResult
...
+                Ok(compact) => {
+                    if let Some(tx) = tx_exporter {
+                        let compact_log = compact.clone();
+                        match tx.try_send(LogEvent { line: compact }) {
+                            Ok(()) => {
+                                info!("{compact_log}");
+                                return ProcessingResult::Success;
+                            }
+                            Err(e) => match e {
+                                tokio::sync::mpsc::error::TrySendError::Full(_) => {
```
tiagolobocastro: This code is very nested here, I suggest you unwind these scopes as much as possible

tiagolobocastro: And even like here you can just do like:
```
Err(tokio::sync::mpsc::error::TrySendError::Full(_)) => ....
Err(tokio::sync::mpsc::error::TrySendError::Closed(_)) => ....
````

This will reduce nesting

File: events-aggregator/src/main.rs [hunk 17]
```diff
@@ -123,36 +205,56 @@ fn process_message(subject_str: &str, payload_bytes: &[u8]) -> ProcessingResult
...
+                                        error = %e,
+                                        compact_log = %compact_log,
+                                        "Exporter channel closed; treating this message as permanent failure"
+                                    );
+                                    return ProcessingResult::PermanentFailure;
+                                }
+                            },
+                        }
+                    } else {
+                        info!("{compact}");
```
tiagolobocastro: reverse if and return early

events-aggregator/src/main.rs
tiagolobocastro: What are we doing here, if the nats message is not json shouldn't we just ignore it?

events-aggregator/src/exporter/engine.rs
```diff
@@ -0,0 +1,276 @@
...
+        }
+    }
+
+    loop {
+        tokio::select! {
+            // Case A: Process a new incoming telemetry event from the channel pipeline
+            Some(event) = rx.recv() => {
+                if batch.is_empty() {
+                    let delay = Duration::from_millis(BATCH_TIMEOUT_MS);
+                    timeout_key = Some(delay_queue.insert((), delay));
```
tiagolobocastro: What is this delay queue giving us vs a timer firing off ?

tiagolobocastro (commented):
Please update the title, we're also pushing to stdout?

Also there are not chores, it's feat

events-aggregator/src/exporter/engine.rs
```diff
@@ -0,0 +1,276 @@
...
+        .unwrap_or(0);
+    let mut rotated_size = tokio::fs::metadata(&rotated_path)
+        .await
+        .map(|m| m.len())
+        .unwrap_or(0);
+    let pending_size = pending_bytes as u64;
+    let max_total_size = size_limit as u64;
+
+    let per_file_limit = max_total_size * 40 / 100;
+
```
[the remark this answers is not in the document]
rohan2794: added validation for file limit

events-aggregator/src/main.rs
rohan2794: Removed this, if message is non JSON then ignore

events-aggregator/src/main.rs
rohan2794: fixed

events-aggregator/src/main.rs
rohan2794: fixed

events-aggregator/src/main.rs
rohan2794: Helm chart have option to disable it with default JetStream is enabled. I have added this in case if JetStream is disabled

events-aggregator/src/exporter/engine.rs
rohan2794: added it to debug

events-aggregator/Cargo.toml
rohan2794: done

events-aggregator/src/client/nats_client.rs
rohan2794: moved connection and request timeout to args

events-aggregator/src/exporter/engine.rs
rohan2794: Replaced DelayQueue with Sleep

events-aggregator/src/exporter/engine.rs
rohan2794: Yes it can be done . Either implement custom subscriber or add file-layer option to exiting TracingTelemetry. If required this can be addressed as part of separate PR .

events-aggregator/src/client/nats_client.rs
rohan2794: Yes, we can move NATS connection management part with its reconnect/backoff logic to control plane .  The subscription setup (setup_jetstream, setup_core_nats) is more specific to this aggregator's needs and could stay here, built on top of the shared connection crate.

events-aggregator/src/main.rs
```diff
@@ -1,63 +1,141 @@
...
 
-    // Initialize NATS Manager
-    let nats_mgr = NatsManager::new(&nats_url).await?;
-    nats_mgr
-        .start_subscribing(subject, jetstream_enabled, tx)
-        .await?;
+    unified_event_processor(rx, tx_exporter).await;
 
-    // Unified Processor Loop
+    anyhow::bail!("NATS consumer exited unexpectedly")
```
tiagolobocastro: this doesn't seem right...
in another comment somewhere I had asked you about the shutdown when we give up, but we also need to handle graceful shutdown and exist cleanly on sigterm is received - pls check how we handle it in other places

events-aggregator/src/main.rs
```diff
@@ -103,84 +179,61 @@ async fn main() -> anyhow::Result<()> {
...
+    let json_payload = match serde_json::from_slice::<Value>(payload_bytes) {
+        Ok(v) => v,
+        Err(e) => {
+            warn!(subject = %subject_str, error = %e, "Received non-JSON event payload; discarding");
+            return ProcessingResult::PermanentFailure;
+        }
+    };
+
+    let log_line = serde_json::json!({
+        "fields": {
```
tiagolobocastro: Do we need another special marker so we can identify it's an event? fields is rather generic

events-aggregator/src/exporter/engine.rs
```diff
@@ -0,0 +1,212 @@
...
+/// Supports human-readable values such as `10MiB`, `1.5GiB`, `10000`, and `8 KB`.
+pub fn dir_size_limit(value: &str) -> Result<u64, String> {
+    parse_size(value).map_err(|e| format!("invalid size '{value}': {e}"))
+}
+
+/// A single event produced by the aggregator pipeline.
+/// `line` contains compact JSON for file export.
+#[derive(Debug)]
+pub struct LogEvent {
+    pub line: String,
```
tiagolobocastro, line 23: missing doc comment

events-aggregator/src/exporter/engine.rs
```diff
@@ -0,0 +1,212 @@
...
+pub async fn run_exporter(mode: ExporterMode, mut rx: mpsc::Receiver<LogEvent>) {
+    let mut batch: Vec<LogEvent> = Vec::with_capacity(BATCH_MAX_SIZE);
+    let flush_timeout = Duration::from_millis(BATCH_TIMEOUT_MS);
+    let mut timer_active = false;
+    let mut sleep = Box::pin(tokio::time::sleep(Duration::MAX));
+
+    match &mode {
+        ExporterMode::Disabled => {
+            tracing::info!(
+                "Batch Exporter in disabled mode; events will be consumed but not written."
```
tiagolobocastro: If disabled then do we even need to run this exporter?

rohan2794: No need to run exporter if Disabled. I will make the required changes.

events-aggregator/src/main.rs
rohan2794: @tiagolobocastro  Does this make sense ?
 let log_line = serde_json::json!({
        "type": "mbus_event",
        "subject": subject_str,
        "payload": json_payload,
    });

events-aggregator/src/main.rs
rohan2794: used exiting control plane crate shutdown to handle graceful shutdown

events-aggregator/src/main.rs
```diff
@@ -1,63 +1,157 @@
...
 
-    // Setup Internal Channel
-    let (tx, mut rx) = mpsc::channel::<UnifiedMessage>(512);
+    let error = tokio::select! {
+        signal = shutdown::Shutdown::wait_sig() => {
+            info!("Received {signal:?}; shutting down gracefully");
+            None
+        }
+        _ = unified_event_processor(rx, tx_exporter.clone()) => {
+            Some(anyhow::anyhow!("NATS consumer exited unexpectedly"))
```
tiagolobocastro: nit: you could just return result here, then no need to map
 
```suggestion
    let result = tokio::select! {
        signal = shutdown::Shutdown::wait_sig() => {
            info!("Received {signal:?}; shutting down gracefully");
            Ok(())
        }
        _ = unified_event_processor(rx, tx_exporter.clone()) => {
            Err(anyhow::anyhow!("NATS consumer exited unexpectedly"))
```

tiagolobocastro: Somewhere there was a failure mode when stopped retrying, can we not get that error wire up so we exit with the actual error?

tiagolobocastro (approved):

tiagolobocastro:
please squash the commits

Abhinandan-Purkait (approved):

events-aggregator/src/main.rs
rohan2794: @tiagolobocastro  I have raised PR to propagate error , https://github.com/openebs/mayastor-control-plane/pull/1124

rohan2794:
@mergifyio queue

PR #889, opened 2026-06-18
feat(chart): add initialDelaySeconds of 1 for api-rest health probes 

niladrih:
The api-rest Pod would try its probes at t=0 and fail. The next try
would eat through 20-30 seconds. Added a delay of 1s to both the
readiness and liveness probes so that they have a higher likelihood
of succeeding.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -289,9 +289,9 @@ apis:
         # -- Frequency for the agent-core liveness probe.
         agentCoreProbeFreq: "20s"
         # -- No. of failures the readiness probe will tolerate.
-        failureThreshold: 2
+        failureThreshold: 3
         # -- No. of seconds of delay before checking the readiness status.
-        initialDelaySeconds: 0
+        initialDelaySeconds: 1
         # -- No. of seconds between readiness probe checks.
         periodSeconds: 20
         # -- No. of seconds of timeout tolerance.
@@ -300,9 +300,9 @@ apis:
         # -- Toggle liveness probe.
         enabled: true
         # -- No. of failures the liveness probe will tolerate.
-        failureThreshold: 1
+        failureThreshold: 3
         # -- No. of seconds of delay before checking the liveness status.
-        initialDelaySeconds: 0
+        initialDelaySeconds: 1
         # -- No. of seconds between liveness probe checks.
         periodSeconds: 30
         # -- No. of seconds of timeout tolerance.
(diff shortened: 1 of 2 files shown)


chart/values.yaml
```diff
@@ -300,9 +300,9 @@ apis:
         # -- Toggle liveness probe.
         enabled: true
         # -- No. of failures the liveness probe will tolerate.
-        failureThreshold: 1
+        failureThreshold: 2
```
tiagolobocastro: Should this be even larger than 2?
https://kubernetes.io/docs/concepts/workloads/pods/probes/

niladrih: Maybe 3? Or what do you have in mind?

Abhinandan-Purkait (approved):

chart/values.yaml
tiagolobocastro: Sure, sounds good

tiagolobocastro (approved):

chart/values.yaml
```diff
@@ -291,7 +291,7 @@ apis:
         # -- No. of failures the readiness probe will tolerate.
         failureThreshold: 2
         # -- No. of seconds of delay before checking the readiness status.
-        initialDelaySeconds: 0
+        initialDelaySeconds: 3
```
```suggestion
        initialDelaySeconds: 1
```

chart/values.yaml
```diff
@@ -300,9 +300,9 @@ apis:
         # -- Toggle liveness probe.
         enabled: true
         # -- No. of failures the liveness probe will tolerate.
-        failureThreshold: 1
+        failureThreshold: 2
         # -- No. of seconds of delay before checking the liveness status.
-        initialDelaySeconds: 0
+        initialDelaySeconds: 3
```
```suggestion
        initialDelaySeconds: 1
```

tiagolobocastro (commented):
I think actually 1 should be enough for the initial delay, we don't do much there before we start?

chart/values.yaml
niladrih: Alright, sounds fair.

niladrih:
bors merge

niladrih:
@mergify queue

niladrih:
bors cancel

niladrih:
@mergify dequeue

niladrih:
@mergify queue

tiagolobocastro:
control-plane module has already been updated

niladrih:
bors merge

PR #892, opened 2026-06-22
Add TLS configuration to the public api in the helm-chart

tiagolobocastro:
feat(chart): add TLS support for the REST API server and its clients
    
    Introduces end-to-end TLS for the REST API, with three certificate
    engine options and optional mutual TLS.
    
    For now tls is disabled by default until the handling of the cert renewal
    is handled by the server and clients.
    
    Adds tls configuration similar to other open source charts.
    We add tls in 3 separate engines:
    1. pod
    2. helm
    3. cert-manager
    Also you can specify your secrets manually for each server and client.
    
    The recommend production approach is to use cert-manager.
    
    Values restructuring:
    - New top-level security.tls block owns the shared CA/issuer
    infrastructure: enabled, autoGenerated.engine
    (pod | helm | cert-manager), issuer settings, and default
    key parameters. This is intentionally cluster-scoped — it would
    be wrong to mix engines or CAs across REST and gRPC.
    - apis.rest.security.tls retains REST-specific leaf-cert settings:
    mutualAuth, existingSecret,
    certManager.{duration,renewBefore,secretName}.
    - New apis.rest.security.tls.clients.<component> blocks
    (csiController, csiNode, callhome, metricsExporter,
    diskpoolOperator, plugin) expose existingSecret and
    certManager.secretName for bring-your-own scenarios.
    
    Certificate engines:
    - pod: server generates a transient self-signed cert at startup
    (--auto-tls). No Secrets. Cannot be combined with mutualAuth.
    - helm: chart generates a self-signed CA and leaf certs stored as
    k8s Secrets. Certificates survive helm upgrade — CA is resolved
    with an explicit if/else so genCA is never called when a CA
    already exists, preventing silent CA rotation on upgrade. All
    helm-engine Secrets (CA + server + 6 clients) are generated in a
    single template file so the CA struct is in scope on first install
    (lookup cannot see same-pass Secrets).
    - cert-manager: emits a self-signed CA Issuer, a Certificate
    for the server, and (with mutualAuth) separate Certificate
    resources per client component. Supports
    existingIssuer/existingIssuerKind to reference a pre-existing
    issuer.
    
    When mutualAuth: true, each component gets its own key pair, this
    enables mTLS basically.
    
---

    refactor(kubectl/plugin): add debug-error cli switch
    
    This allows us to see the debug implementation of the error, which can
    show more useful information, especially when dealing with tls.
    
---

    feat(tls): add native cert store to container images
    
    This is required now that we're supporting tls.


Abhinandan-Purkait (approved):
LGTM. I guess you will add the NetworkPolicy and the corresponding http exposure on the rest.

chart/README.md
```diff
@@ -50,6 +50,103 @@ This removes all the Kubernetes components associated with the chart and deletes
...
+- **`apis.rest.security.tls`** — REST-specific leaf certificate settings (`mutualAuth`, `certManager.duration`, `existingSecret`).
+  Per-client overrides live under `apis.rest.security.tls.clients.<component>` (`existingSecret`, `certManager.secretName`).
+
+| Setting | Effect |
+|---------|--------|
+| `security.tls.enabled: false` | No TLS. All services use plain-text. |
+| `security.tls.autoGenerated.engine: pod` (default) | Server generates a transient cert at startup (`--auto-tls`). No Secrets created. Cannot be combined with `mutualAuth`. |
+| `security.tls.autoGenerated.engine: helm` | Chart generates self-signed certificates stored in k8s Secrets. Clients verify the server cert. |
+| `security.tls.autoGenerated.engine: certManager` | cert-manager provisions and rotates certificates. cert-manager must be installed. |
+| `+ apis.rest.security.tls.mutualAuth: true` | Full mutual TLS. Both sides verify. Separate client certs are provisioned for each component (csi-controller, csi-node, obs-callhome, io-engine, operator-diskpool, plugin). Requires `engine: helm` or `engine: certManager`. |
```
niladrih:
```suggestion
| `+ apis.rest.security.tls.mutualAuth: true` | Full mutual TLS (mTLS). Both sides verify. Separate client certs are provisioned for each component (csi-controller, csi-node, obs-callhome, io-engine, operator-diskpool, plugin). Requires `engine: helm` or `engine: certManager`. |
```

chart/templates/_helpers.tpl
```diff
@@ -448,3 +448,304 @@ Usage:
...
+*/}}
+{{- define "validate_tls_mode" -}}
+{{- $globalTls := .Values.security.tls -}}
+{{- $restTls := .Values.apis.rest.security.tls -}}
+{{- $mutualAuth := dig "mutualAuth" $globalTls.mutualAuth $restTls -}}
+{{- $ag := $globalTls.autoGenerated -}}
+{{- if and $globalTls.enabled (not $ag.enabled) -}}
+{{- if not $restTls.existingSecret -}}
+{{- fail "tls: apis.rest.security.tls.existingSecret must be set when tls.autoGenerated.enabled is false" -}}
+{{- end -}}
```
niladrih: It'd nice to add indentation to the code inside the if blocks

chart/templates/_helpers.tpl
```diff
@@ -448,3 +448,304 @@ Usage:
...
+Usage: {{ include "validate_tls_mode" . }}
+*/}}
+{{- define "validate_tls_mode" -}}
+{{- $globalTls := .Values.security.tls -}}
+{{- $restTls := .Values.apis.rest.security.tls -}}
+{{- $mutualAuth := dig "mutualAuth" $globalTls.mutualAuth $restTls -}}
+{{- $ag := $globalTls.autoGenerated -}}
+{{- if and $globalTls.enabled (not $ag.enabled) -}}
+{{- if not $restTls.existingSecret -}}
+{{- fail "tls: apis.rest.security.tls.existingSecret must be set when tls.autoGenerated.enabled is false" -}}
```
niladrih: This error message instruction won't work with the openebs/openebs chart.

chart/templates/_helpers.tpl
```diff
@@ -448,3 +448,304 @@ Usage:
...
+{{- end -}}
+
+{{- define "rest_api_tls_csi_controller_client_secret_name" -}}
+{{- $globalTls := .Values.security.tls -}}
+{{- $client := .Values.apis.rest.security.tls.clients.csiController -}}
+{{- if $globalTls.enabled -}}
+{{- if not $globalTls.autoGenerated.enabled -}}
+{{- $client.existingSecret -}}
+{{- else if eq $globalTls.autoGenerated.engine "cert-manager" -}}
+{{- $client.certManager.secretName | default (printf "%s-api-rest-csi-controller-crt" .Release.Name) -}}
```
niladrih: The defaults seem to add lengthy suffixes to the release name. These could exceed the 63 character limit. Maybe we should put the defaults inside a template function where they'd be trimmed down to 63 characters, if the printf output is >=64.

chart/templates/mayastor/apis/api-rest-network-policy.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.security.tls.enabled }}
```
niladrih, line 1: There maybe should be an additional `networkPolicy.enabled` inside the `security.tls.enabled`? For people who want to enable TLS, but not use the network policy.

chart/templates/mayastor/csi/csi-controller-cert.yaml
```diff
@@ -0,0 +1,31 @@
...
+    {{ include "label_prefix" . }}/version: {{ .Chart.Version }}
+spec:
+  secretName: {{ include "rest_api_tls_csi_controller_client_secret_name" . }}
+  commonName: {{ printf "%s-api-rest-csi-controller" .Release.Name }}
+  issuerRef:
+    name: {{ printf "%s-tls-ca-issuer" .Release.Name }}
+    kind: Issuer
+  subject:
+    organizations:
+      - "OpenEBS"
```
niladrih, line 23: This should be configurable

chart/values.yaml
```diff
@@ -991,10 +1098,11 @@ nats:
   exporter:
     image:
       registry: docker.io
+
 obs:
   callhome:
     # -- Enable callhome
-    enabled: true
+    enabled: false # missing native roots
```
niladrih: Is this required?

k8s/plugin/src/main.rs
```diff
@@ -62,21 +65,33 @@ impl Deref for CliArgs {
 
 #[tokio::main]
 async fn main() {
+    fn fmt_err(e: &(impl std::fmt::Debug + std::fmt::Display), debug: bool) -> String {
+        if debug {
+            format!("{e:?}")
+        } else {
+            format!("{e}")
+        }
```
niladrih, line 73:
```suggestion
        if debug {
            return format!("{e:?}");
        }
        return format!("{e}");
```

chart/templates/_helpers.tpl
tiagolobocastro: On it

chart/templates/_helpers.tpl
tiagolobocastro: hmm AFAIK it will show the sub-chart?

chart/templates/_helpers.tpl
tiagolobocastro: We could though we don't follow this anywhere else in the codebase
I'd rather follow that as another piece of work if that's ok

Also to consider removing the helm name prefix completely

chart/templates/mayastor/apis/api-rest-network-policy.yaml
tiagolobocastro: Is there any use-case for disabling it?
Also this is an interim until we change to HTTP only for the probes

chart/templates/mayastor/csi/csi-controller-cert.yaml
tiagolobocastro: I don't think so, this is in relation to us
The other charts don't do that

chart/values.yaml
tiagolobocastro: oops, good catch no

k8s/plugin/src/main.rs
tiagolobocastro: not sure there's any benefit, present seems fine to me

chart/templates/_helpers.tpl
tiagolobocastro: Done

chart/templates/_helpers.tpl
niladrih: It will, I suppose.

chart/templates/_helpers.tpl
niladrih: We could do this as a separate task across the codebase.

chart/templates/mayastor/apis/api-rest-network-policy.yaml
niladrih: > interim until we change to HTTP only for the probes

Alright

niladrih (approved):

tiagolobocastro:
@mergify dequeue

tiagolobocastro:
@mergify queue

matns:
Please consider allowing not to install NetworkPolicy resources. Standardized K8s NetworkPolicy resources sometimes are lacking and one needs to use CNI-specific (like CiliumNetworkPolicy) resources.

tiagolobocastro:
> Please consider allowing not to install NetworkPolicy resources. Standardized K8s NetworkPolicy resources sometimes are lacking and one needs to use CNI-specific (like CiliumNetworkPolicy) resources.

Hi @matns, I've made it optional here: #937
It's also now less important since I've changed the rest itself to allow specifying enabling HTTP only for the live/ready probes.

PR #897, opened 2026-06-24
feat(deployment): mount kubelet directory in io-engine pods

susobhandey:
Mount the kubelet directory into io-engine pods to enable access to CPUManager state information required for dynamic reactor affinity updates.

Add a hostPath mount for /var/lib/kubelet in the
io-engine DaemonSet template.

Signed-off-by: Susobhan Dey [susobhan.dey@datacore.com](mailto:susobhan.dey@datacore.com)

### DIFF
--- a/chart/templates/mayastor/io/io-engine-daemonset.yaml
+++ b/chart/templates/mayastor/io/io-engine-daemonset.yaml
@@ -152,6 +152,8 @@ spec:
           mountPath: /var/local/{{ .Release.Name }}/io-engine/
         - name: hugepages-2mi
           mountPath: /dev/hugepages-2mi
+        - name: kubelet-dir
+          mountPath: /var/lib/kubelet
         {{- if .Values.io_engine.resources.requests.hugepages1Gi }}
         - name: hugepages-1gi
           mountPath: /dev/hugepages-1gi
@@ -226,3 +228,7 @@ spec:
           path: /var/local/{{ .Release.Name }}/io-engine/
           type: DirectoryOrCreate
       {{- include "rest_api_tls_client_volume" (dict "ctx" . "secretName" (include "rest_api_tls_metrics_exporter_client_secret_name" .)) | nindent 6 }}
+      - name: kubelet-dir
+        hostPath:
+          path: {{ .Values.csi.node.kubeletDir }}
+          type: Directory


chart/templates/mayastor/io/io-engine-daemonset.yaml
```diff
@@ -220,3 +222,7 @@ spec:
         hostPath:
           path: /var/local/{{ .Release.Name }}/io-engine/
           type: DirectoryOrCreate
+      - name: kubelet-dir
```
Abhinandan-Purkait, line 231: Don't you need mount propagation? By default its None.

Abhinandan-Purkait: Also you might not need to mount the entire /var/lib/kubelet as that has a lot of things, maybe the specific path you are interested in.

Also I wonder without mount propagation will you get the required inotify events

tiagolobocastro: > Also I wonder without mount propagation will you get the required inotify events

I suspect not since we want only the file inside it, no other mount, but probably might still be worth adding propagation and see if there's any difference on your testing

Also we should not hardcode `/var/lib/kubelet`, should use the helm values since it may be different on different distributions of K8s?

chart/templates/mayastor/io/io-engine-daemonset.yaml
```diff
@@ -220,3 +222,7 @@ spec:
...
+        hostPath:
+          path: /var/lib/kubelet
```
niladrih: This path cannot be hardcoded, it's a configurable parameter. You can use it like [how it's done on the csi-node](https://github.com/openebs/mayastor-extensions/blob/develop/chart/templates/mayastor/csi/csi-node-daemonset.yaml#L165C1-L168C26).

niladrih: I see that this review comment is already posted by Tiago above. Feel free to resolve my comment.

Abhinandan-Purkait (approved):

niladrih (approved):

abhilashshetty04 (approved):

tiagolobocastro (approved):

abhilashshetty04:
@susobhandey , Okay to merge this?

PR #928, opened 2026-07-08
refactor(supportability): expose LokiClient for external reuse

rohan2794:
Make LokiClient and LokiError pub, add fetch_lines() for callers that need a Vec<String> rather than file output, and add with_logql_filters() and with_page_size() builder methods. 
No behaviour change to existing fetch_and_dump_logs path.

### DIFF
--- a/k8s/supportability/src/collect/logs/loki.rs
+++ b/k8s/supportability/src/collect/logs/loki.rs
@@ -120,7 +120,7 @@ impl LogDirection {
 /// Http client to interact with Loki (a log management system)
 /// to fetch historical log information.
 #[derive(Debug)]
-pub(crate) struct LokiClient {
+pub struct LokiClient {
     /// Address of Loki service.
     uri: String,
     /// Loki client
@@ -132,15 +132,18 @@ pub(crate) struct LokiClient {
     /// Determines the sort order of logs. Supported values are "forward" or "backward".
     /// Defaults to forward
     direction: LogDirection,
-    /// Maximum number of entries to return on one http call.
+    /// Maximum number of entries to return per HTTP request (page size).
     limit: u64,
     /// Tenant id to be used for querying.
     tenant_id: String,
+    /// LogQL pipeline filter expressions appended after the stream selector,
+    /// e.g. `vec!["| json", "| payload_category=\"Volume\""]`.
+    logql_filters: Vec<String>,
 }
 
 impl LokiClient {
     /// Instantiate new instance of Http Loki client.
-    pub(crate) async fn new(
+    pub async fn new(
         uri: Option<String>,
         kubeconfig_args: crate::KubeConfigArgs,
         namespace: String,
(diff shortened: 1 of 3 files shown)


Abhinandan-Purkait (approved):

k8s/supportability/src/collect/logs/loki.rs
```diff
@@ -205,9 +208,91 @@ impl LokiClient {
...
+            .split(',')
+            .map(|key_value_pair| {
+                let pairs = key_value_pair.split('=').collect::<Vec<&str>>();
+                format!("{}=\"{}\",", pairs[0], pairs[1])
+                    .replace('.', "_")
+                    .replace('/', "_")
+            })
+            .collect::<String>();
+        if !label_filters.is_empty() {
+            label_filters.pop();
```
tiagolobocastro: Why do we need to remove the last item?

k8s/supportability/src/collect/logs/loki.rs
```diff
@@ -205,9 +208,91 @@ impl LokiClient {
...
+                        lines.push(line);
+                        if limit > 0 && lines.len() >= limit {
+                            return Ok(lines);
+                        }
+                    }
+                }
+                Ok(None) => break,
+                Err(e) => return Err(e),
+            }
+        }
```
tiagolobocastro, line 283: I think you can simplify this

```suggestion
        while let Some(batch) = poller.poll_next().await? {
            for line in batch {
                 lines.push(line);
                 if limit > 0 && lines.len() >= limit {
                     return Ok(lines);
                 }
             }
        }
```

k8s/supportability/src/collect/logs/loki.rs
```diff
@@ -205,9 +208,91 @@ impl LokiClient {
...
+        let mut poller = LokiPoll {
+            uri: self.uri.clone(),
+            endpoint: self.logs_endpoint.clone(),
+            since: self.since,
+            query_params,
+            next_start_epoch_timestamp: 0,
+            client: self,
+        };
+
+        let mut lines = Vec::new();
```
tiagolobocastro:
```suggestion
        let mut lines = Vec::with_capacity(limit);
```

k8s/supportability/src/collect/logs/loki.rs
```diff
@@ -205,9 +208,91 @@ impl LokiClient {
...
+    /// applying any `logql_filters` set via `with_logql_filters()`.
+    ///
+    /// Paginates automatically. Stops as soon as `limit` lines have been
+    /// accumulated; pass `0` to fetch all available lines (no cap).
+    /// Returns one string per log line, in forward time order.
+    pub async fn fetch_lines(
+        &mut self,
+        label_selector: String,
+        container_name: String,
+        limit: usize,
```
tiagolobocastro, line 246: How does the limit work here?
Is it like a max the caller wants to enforce? It not for like paging right?

k8s/supportability/src/collect/logs/loki.rs
rohan2794: yes not required to pop last item. If it's a trailing-comma-then-pop antipattern. Each format! call appends a ,, so after .collect::<String>() we get app="eventing-aggregator", and need to pop the last character.

tiagolobocastro: ah, I missed the comma being re-added

k8s/supportability/src/collect/logs/loki.rs
```diff
@@ -205,9 +208,91 @@ impl LokiClient {
...
+    ) -> Result<Vec<String>, LokiError> {
+        let mut label_filters: String = label_selector
+            .split(',')
+            .map(|key_value_pair| {
+                let pairs = key_value_pair.split('=').collect::<Vec<&str>>();
+                format!("{}=\"{}\",", pairs[0], pairs[1])
```
tiagolobocastro: This will crash if the label_selector doesn't follow the `key=val` pattern

rohan2794: I will make https://github.com/openebs/mayastor-extensions/blob/develop/k8s/supportability/src/collect/logs/loki.rs#L225 as private function and reuse for both methods( `fetch_and_dump_logs`  and `fetch_lines` )

rohan2794: @tiagolobocastro  Should we silently skips entries which doesn't follow the key=val pattern ?

k8s/supportability/src/collect/logs/loki.rs
rohan2794: Yes, exactly — the limit parameter on fetch_lines is the total cap the caller wants, not the page size.
The page size is self.limit (the LokiClient field, default 3000, overridable via with_page_size()). That controls how many entries Loki returns per HTTP request.

The limit parameter on fetch_lines is the caller's ceiling across all pages — once lines.len() >= limit, it stops fetching and returns early, even if there are more pages available. Passing 0 means "no cap, fetch everything.

So the flow is:
- Each poll_next() call fetches up to self.limit (3000) lines from Loki
- fetch_lines accumulates them across pages and stops as soon as it hits the caller's limit

tiagolobocastro: Then you have to adjust the poller with a new limit, otherwise it's always getting the per request max (ie 3000)
Let's say you wanted 3001 pages - with current code I think you'll get 3000 + 3000, and ditch the last 2999?

k8s/supportability/src/collect/logs/loki.rs
tiagolobocastro: I think it should be an error?

k8s/supportability/src/collect/logs/loki.rs
rohan2794: Yeah, we can make the page size on LokiPoll adjustable, and cap it to remaining before each request.

k8s/supportability/src/collect/logs/loki.rs
rohan2794: Should we keep the silent skip and and add debug assert ?

k8s/supportability/src/collect/logs/loki.rs
```diff
@@ -205,9 +208,84 @@ impl LokiClient {
...
+        let mut lines = Vec::with_capacity(limit);
+        loop {
+            if limit > 0 {
+                // Cap the page size to how many lines we still need so we don't
+                // ask Loki for more entries than the caller wants.
+                poller.page_size = poller.page_size.min((limit - lines.len()) as u64);
+            }
+            match poller.poll_next().await? {
+                Some(batch) => {
+                    for line in batch {
```
tiagolobocastro: Can't we simply extend here?

```rust
lines.extend(batch);
```

k8s/supportability/src/collect/logs/loki.rs
tiagolobocastro: Should be an error, but I see this is already present on existing code, so can do this in another change

tiagolobocastro (approved):

k8s/supportability/src/collect/logs/loki.rs
rohan2794: I will handle this in support dump collection PR change.

k8s/supportability/src/collect/logs/loki.rs
rohan2794: Yes, loop and push can collapse to extend.

abhilashshetty04:
bors merge

PR #930, opened 2026-07-10
chore: patch images before release chart publish

pchandra19:


.github/workflows/release.yml
```diff
@@ -145,6 +145,12 @@ jobs:
       - name: Update the registry and the repository
         run: |
           nix-shell --pure --run "./scripts/helm/update-reg-repo.sh --registry docker.io --repository ${{ github.repository_owner }}" ./dependencies/control-plane/utils/dependencies/scripts/staging/shell.nix
+      
+      - name: Generate image list
+        run: nix-shell ./scripts/helm/shell.nix --run "./scripts/helm/images.sh generate --dependency-update --exit-code"
+        
+      - name: Patch chart/Chart.yaml
+        run: nix-shell ./scripts/helm/shell.nix --run "./scripts/helm/images.sh patch --exit-code"
```
Abhinandan-Purkait: We probably can't use the --exit-code here

pchandra19: makes sense. Have removed.

.github/workflows/release.yml
```diff
@@ -145,6 +145,12 @@ jobs:
       - name: Update the registry and the repository
         run: |
           nix-shell --pure --run "./scripts/helm/update-reg-repo.sh --registry docker.io --repository ${{ github.repository_owner }}" ./dependencies/control-plane/utils/dependencies/scripts/staging/shell.nix
+      
+      - name: Generate image list
+        run: nix-shell ./scripts/helm/shell.nix --run "./scripts/helm/images.sh generate --dependency-update"
+
+      - name: Patch chart/Chart.yaml
+        run: nix-shell ./scripts/helm/shell.nix --run "./scripts/helm/images.sh patch"
```
niladrih, line 153: nit: two different steps make this unnecessarily slow, it'll spin up two different shells.
```suggestion
      - name: Generate and patch image list
        run: |
          nix-shell ./scripts/helm/shell.nix --run "./scripts/helm/images.sh generate --dependency-update"
          nix-shell ./scripts/helm/shell.nix --run "./scripts/helm/images.sh patch"
```

niladrih (approved):

.github/workflows/release.yml
pchandra19: Yes, doing it in one step is efficient but in case if any of the command fails troubleshooting is easier this way which each step having its own logs.

Abhinandan-Purkait (approved):

tiagolobocastro:
@mergify backport release/2.11

PR #935, opened 2026-07-13
feat(plugin/purge): make diskpool cleanup optional

niladrih:
This lib may be used by binaries which want to handle the diskpool cleanup on their own
terms. This change adds an option to skip the diskpool CR cleanup entirely via a new
CleanupDspArg. This option is no user-configurable, and only meant for use by importing
codebases.

### DIFF
--- a/k8s/plugin/src/resources/mod.rs
+++ b/k8s/plugin/src/resources/mod.rs
@@ -1,4 +1,4 @@
-mod diskpool_cleanup;
+pub mod diskpool_cleanup;
 mod events;
 mod io_engine_label_check;
 
@@ -175,14 +175,14 @@ pub enum Operations {
 pub struct DeleteArgs {
     /// Ignore error if resource is not found.
     #[clap(long, short, global = true)]
-    ignore_not_found: bool,
+    pub ignore_not_found: bool,
 
     /// Automatically confirm and assume yes for all prompts.
     #[clap(long, short, global = true)]
     pub yes: bool,
 
     #[clap(subcommand)]
-    resource: DeleteResources,
+    pub resource: DeleteResources,
 }
 
 /// The type of resources which support the delete operation.
@@ -216,6 +216,12 @@ pub struct CleanupDspArgs {
     /// Timeout per DiskPool CR deletion when using --cleanup-dsp.
     #[clap(long, default_value = "60s")]
     pub cleanup_dsp_timeout: humantime::Duration,
+
+    /// Skip the DiskPool CR cleanup step.
+    /// When set, the REST pool/node delete still runs but the DiskPool CR deletion is skipped.
+    /// Intended for downstream plugins that handle CR cleanup themselves.
+    #[clap(skip)]
+    pub skip_dsp_cr_cleanup: bool,
 }
 
 /// Arguments for deleting a pool.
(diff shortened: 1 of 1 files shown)


k8s/plugin/src/resources/mod.rs
```diff
@@ -234,17 +234,121 @@ pub struct PoolDeleteArgs {
...
     #[clap(flatten)]
-    cleanup: CleanupDspArgs,
+    pub cleanup: CleanupDspArgs,
 }
 
+/// Strategy for cleaning up DiskPool Custom Resources after a pool or node
+/// delete via the REST API.
+///
+/// The default implementation deletes DiskPool CRs
+/// directly.  Downstream consumers can supply an alternative that deletes an
```
Abhinandan-Purkait:
```suggestion
/// directly. Downstream consumers can supply an alternative that deletes an
```

Abhinandan-Purkait (approved):

k8s/plugin/src/resources/mod.rs
```diff
@@ -211,6 +211,14 @@ pub struct CleanupDspArgs {
     /// Timeout per DiskPool CR deletion when using --cleanup-dsp.
     #[clap(long, default_value = "60s")]
     pub cleanup_dsp_timeout: humantime::Duration,
+
+    /// Skip the DiskPool CR cleanup step.{n}
+    /// When set, the REST pool/node delete still runs but the DiskPool CR
+    /// deletion is skipped.{n}
+    /// Intended for downstream plugins that handle CR
+    /// cleanup themselves.
```
tiagolobocastro: Not sure why you're adding `{n}` since we're skipping clap?

Also use longer lines pls

```suggestion
    /// Skip the DiskPool CR cleanup step.
    /// When set, the REST pool/node delete still runs but the DiskPool CR deletion is skipped.
    /// Intended for downstream plugins that handle CR cleanup themselves.
```

k8s/plugin/src/resources/mod.rs
```diff
@@ -334,16 +342,22 @@ impl ExecuteOperation for Operations {
...
+                            diskpool_cleanup::cleanup_dsp(
+                                client,
+                                &cli_args.namespace,
+                                &pool_args.rest_args.pool_id,
+                                pool_args.cleanup.cleanup_dsp_timeout,
+                                pool_deleted,
+                                args.ignore_not_found,
+                            )
+                            .await?;
+                        }
```
tiagolobocastro: Not sure why you're using this syntax here, won't this do?

```suggestion
                        if !pool_args.cleanup.skip_dsp_cr_cleanup {
                            let client = cli_args.client().await?;
                            diskpool_cleanup::cleanup_dsp(
                                client,
                                &cli_args.namespace,
                                &pool_args.rest_args.pool_id,
                                pool_args.cleanup.cleanup_dsp_timeout,
                                pool_deleted,
                                args.ignore_not_found,
                            )
                            .await?;
                        }
```

niladrih: Yes, it will, it was like that initially, okay, I'll remove the labels.
I just liked to set it apart, because that's how we're looking at it now, no other reason.

tiagolobocastro (approved):

PR #942, opened 2026-07-16
feat(plugin): add volume fallback and typed filters for get events

rohan2794:
Fall back to eventing-aggregator pod exec (kube ws) when Loki is not
found in the cluster. Explicit --loki-endpoint errors on failure;
auto-discovery silently falls back so the command always produces results.

Add eight new filter flags: --component, --pool, --volume, --replica,
--rebuild-status, --state, --filter. Typed flags use prost struct access;
--filter accepts a full JSON dot-path with leading/trailing * wildcards
and walks the raw event JSON so no hardcoded field list is needed.

Skip REST client init for the events subcommand since it uses kube
directly and never touches the REST API.

### DIFF
--- a/k8s/plugin/src/main.rs
+++ b/k8s/plugin/src/main.rs
@@ -1,5 +1,5 @@
 use plugin::ExecuteOperation;
-use resources::{init_rest, Error, Operations};
+use resources::{Error, Operations};
 
 use clap::Parser;
 use std::{env, ops::Deref, path::PathBuf};
@@ -105,9 +105,9 @@ async fn main() {
 
 impl CliArgs {
     async fn execute(self) -> Result<(), Error> {
-        // Initialise the REST client.
-        init_rest(&self.args).await?;
-
+        if self.operations.needs_rest_init() {
+            resources::init_rest(&self.args).await?;
+        }
         tokio::select! {
             shutdown = shutdown::Shutdown::wait_sig() => {
                 Err(anyhow::anyhow!("Interrupted by {shutdown:?}").into())
--- a/k8s/plugin/src/resources/events.rs
+++ b/k8s/plugin/src/resources/events.rs
@@ -1,22 +1,32 @@
 use anyhow::anyhow;
+use chrono::{DateTime, Utc};
 use clap::builder::TypedValueParser;
-use events_api::event::{Component, EventAction, EventCategory, EventMessage};
+use events_api::event::{
+    Component, EventAction, EventCategory, EventDetails, EventMessage, RebuildStatus,
+};
+use k8s_openapi::api::core::v1::Pod;
+use kube::{
+    api::{AttachParams, ListParams},
+    Api,
+};
 use plugin::resources::utils::{optional_cell, print_table, CreateRow, GetHeaderRow, OutputFormat};
 use prettytable::{row, Row};
 use serde::Serialize;
-use std::path::PathBuf;
 use strum::IntoEnumIterator;
 use supportability::KubeConfigArgs;
+use tokio::io::{AsyncBufRead, AsyncBufReadExt, BufReader};
 
 const EVENTS_CONTAINER: &str = "eventing-aggregator";
 const EVENTS_LABEL_SELECTOR: &str = "app=eventing-aggregator";
 const MBUS_EVENT_TYPE: &str = "mbus_event";
 const DEFAULT_LIMIT: usize = 1000;
+const EVENTS_VOLUME_DIR: &str = "/var/events";
 
 /// Args for `kubectl mayastor get events`.
 #[derive(Debug, clap::Args)]
 pub struct EventsArgs {
-    /// Loki base URL. If omitted, Loki is auto-discovered via the K8s service.
+    /// Loki base URL. If omitted, Loki is auto-discovered via the K8s service;
+    /// if discovery also fails, events are read from the eventing-aggregator pod volume.
     #[arg(long)]
     loki_endpoint: Option<String>,
 
@@ -36,6 +46,38 @@ pub struct EventsArgs {
     #[arg(long)]
     target: Option<String>,
 
+    /// Filter by source component (repeatable or comma-separated).
+    #[arg(long = "component", value_delimiter = ',', value_parser = component_parser())]
+    components: Vec<Component>,
+
+    /// Filter events for a pool (substring match on target).
+    #[arg(long)]
+    pool: Option<String>,
+
+    /// Filter events touching a volume (target, snapshot, volume_id).
+    #[arg(long)]
+    volume: Option<uuid::Uuid>,
+
+    /// Filter events by replica UUID.
+    #[arg(long)]
+    replica: Option<uuid::Uuid>,
+
+    /// Filter rebuild events by outcome (repeatable or comma-separated).
+    #[arg(long = "rebuild-status", value_delimiter = ',', value_parser = rebuild_status_parser())]
+    rebuild_statuses: Vec<RebuildStatus>,
+
+    /// Filter state-change events by state value (substring on previous or next state).
+    #[arg(long)]
+    state: Option<String>,
+
+    /// Filter by JSON field path and pattern: path=value (repeatable, AND logic).
+    /// Path is relative to the event payload, e.g.
+    /// --filter "metadata.source.eventDetails.replicaDetails.poolUuid=abc*".
+    /// Supports leading/trailing * wildcards. Events with an unknown or absent path
+    /// are excluded.
+    #[arg(long = "filter")]
+    filters: Vec<String>,
+
     /// Events from last duration (e.g. 1h, 30m).
     #[arg(long, default_value = "24h")]
     since: humantime::Duration,
@@ -44,7 +86,7 @@ pub struct EventsArgs {
     #[arg(long, default_value_t = DEFAULT_LIMIT)]
     limit: usize,
 
-    /// Loki tenant ID (X-Scope-OrgID header).
+    /// Loki tenant ID / X-Scope-OrgID header (loki source only).
     #[arg(long, default_value = "openebs")]
     tenant_id: String,
 }
@@ -69,6 +111,9 @@ pub struct EventRecord {
     pub node: String,
     #[serde(skip)]
     pub component: String,
+    /// Raw outer JSON — used by --filter dot-path traversal.
+    #[serde(skip)]
+    raw: serde_json::Value,
     /// Full typed payload; serialised for JSON/YAML output.
     #[serde(flatten)]
     pub event_message: EventMessage,
@@ -116,6 +161,7 @@ impl EventRecord {
             target,
             node,
             component,
+            raw: serde_json::Value::Null,
             event_message: msg,
         }
     }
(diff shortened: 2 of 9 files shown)


k8s/plugin/src/resources/events.rs
```diff
@@ -264,6 +324,94 @@ fn build_logql_filters() -> Vec<String> {
...
+/// Exec into the eventing-aggregator pod and read its ephemeral event files.
+///
+/// Reads `/var/events/events.json` (current) and `/var/events/events.1.json`
+/// (rotated) in a single `cat` call; both files are optional — if the pod has
+/// not yet written them the cat will exit non-zero, which we treat as zero lines
+/// rather than an error so the caller sees an empty-table diagnostic instead.
+async fn fetch_from_volume(
+    kubeconfig_args: &KubeConfigArgs,
+    namespace: &str,
+) -> anyhow::Result<Vec<String>> {
```
Abhinandan-Purkait: Why is it not reusing the client from the other places? We should not create client like this on the fly?

rohan2794: [Fixed](https://github.com/openebs/mayastor-extensions/compare/98e648c89878a7fd28899c0cd0d8502b2f9e5ed8..3002586648b630a519741096515426783fefe760) , reused existing client

k8s/plugin/src/resources/events.rs
```diff
@@ -36,15 +45,47 @@ pub struct EventsArgs {
...
+
+    /// Filter events for a pool (substring match on target).
+    #[arg(long)]
+    pool: Option<String>,
+
+    /// Filter events touching a volume (target, snapshot volume_id, or clone source_uuid).
+    #[arg(long)]
+    volume: Option<String>,
+
+    /// Filter events touching a replica (target or replica_details.replica_name).
```
tiagolobocastro: what's a replica_name?

k8s/plugin/src/resources/events.rs
```diff
@@ -36,15 +45,47 @@ pub struct EventsArgs {
...
+    #[arg(long = "component", value_delimiter = ',', value_parser = component_parser())]
+    components: Vec<Component>,
+
+    /// Filter events for a pool (substring match on target).
+    #[arg(long)]
+    pool: Option<String>,
+
+    /// Filter events touching a volume (target, snapshot volume_id, or clone source_uuid).
+    #[arg(long)]
+    volume: Option<String>,
```
tiagolobocastro: volume and replica should be uuid::Uuid ?

k8s/plugin/src/resources/events.rs
```diff
@@ -193,30 +240,43 @@ impl EventsArgs {
...
+                (lines, "Loki")
+            }
+            None if self.loki_endpoint.is_some() => {
+                return Err(anyhow!(
+                    "Could not connect to Loki at {}. Check the endpoint and try again.",
+                    self.loki_endpoint.as_deref().unwrap_or_default()
+                ));
+            }
+            None => {
+                eprintln!(
```
tiagolobocastro: should add this only on human output?

k8s/plugin/src/resources/mod.rs
```diff
@@ -256,6 +256,10 @@ impl ExecuteOperation for Operations {
     type Args = CliArgs;
     type Error = Error;
     async fn execute(&self, cli_args: &CliArgs) -> Result<(), Error> {
+        // Subcommands that talk to kube directly skip REST client init.
+        if !matches!(self, Operations::Get(GetResourcesK8s::Events(_))) {
+            init_rest(cli_args).await?;
```
tiagolobocastro: This should not exist in the execute interface, otherwise will enforce it upstream
This needs to be in the binary and not the library

k8s/plugin/src/resources/events.rs
```diff
@@ -264,6 +324,77 @@ fn build_logql_filters() -> Vec<String> {
...
+///
+/// Reads `/var/events/events.json` (current) and `/var/events/events.1.json`
+/// (rotated) in a single `cat` call; both files are optional — if the pod has
+/// not yet written them the cat will exit non-zero, which we treat as zero lines
+/// rather than an error so the caller sees an empty-table diagnostic instead.
+async fn fetch_from_volume(
+    kubeconfig_args: &KubeConfigArgs,
+    namespace: &str,
+) -> anyhow::Result<Vec<String>> {
+    let client = kube_proxy::client_from_kubeconfig(
```
tiagolobocastro: we should not do this, we should pass cli_args to the `get_events` and then we can get client from there

k8s/plugin/src/resources/events.rs
```diff
@@ -264,6 +324,77 @@ fn build_logql_filters() -> Vec<String> {
...
+        kubeconfig_args.path.clone(),
+        kubeconfig_args.opts.context.clone(),
+    )
+    .await?;
+
+    let pods: Api<Pod> = Api::namespaced(client, namespace);
```
tiagolobocastro: should come from cli_args as well

k8s/plugin/src/resources/events.rs
```diff
@@ -264,6 +324,77 @@ fn build_logql_filters() -> Vec<String> {
...
+        .await
+        .map_err(|e| anyhow!("Failed to exec into pod {pod_name}: {e}"))?;
+
+    let mut stdout_reader = attached
+        .stdout()
+        .ok_or_else(|| anyhow!("No stdout from pod exec"))?;
+
+    let mut output = String::new();
+    stdout_reader
+        .read_to_string(&mut output)
```
tiagolobocastro: Shouldn't this be streamed? This may be a huge amount of data in-memory?

tiagolobocastro: we could also copy to local file, and then just read from there since we also want some mechanism to support reading from local file right?

k8s/plugin/src/resources/events.rs
```diff
@@ -315,7 +446,154 @@ fn apply_filters(records: Vec<EventRecord>, args: &EventsArgs) -> Vec<EventRecor
...
+            let inner = &pattern[1..pattern.len().saturating_sub(1)];
+            value.contains(inner)
+        }
+        (true, false) => value.ends_with(&pattern[1..]),
+        (false, true) => value.starts_with(&pattern[..pattern.len() - 1]),
+        (false, false) => value == pattern,
+    }
+}
+
+fn component_parser() -> impl clap::builder::TypedValueParser<Value = Component> {
```
tiagolobocastro, line 593: we don't need the parser now that we have the types directly on clap right?

k8s/plugin/src/resources/events.rs
rohan2794: That was wrong, i have updated the comment.

k8s/plugin/src/resources/events.rs
rohan2794: changed it to uuid::Uuid

k8s/plugin/src/resources/events.rs
rohan2794: done

k8s/plugin/src/resources/mod.rs
rohan2794: restored this to binary

k8s/plugin/src/resources/events.rs
rohan2794: fixed

k8s/plugin/src/resources/events.rs
rohan2794: fixed

k8s/plugin/src/resources/events.rs
rohan2794: Now events from pod volume are getting streamed

k8s/plugin/src/resources/events.rs
rohan2794: The parser is still needed  because RebuildStatus and the other types have EnumString , but Component does not.

k8s/plugin/src/resources/events.rs
```diff
@@ -315,7 +447,156 @@ fn apply_filters(records: Vec<EventRecord>, args: &EventsArgs) -> Vec<EventRecor
...
+    }
+}
+
+/// Pattern matching with leading/trailing `*` wildcards. Case-sensitive.
+/// A bare `*` matches everything. Interior `*` is treated as a literal character.
+fn matches_pattern(value: &str, pattern: &str) -> bool {
+    match (pattern.starts_with('*'), pattern.ends_with('*')) {
+        (true, true) => {
+            // Guard: a single '*' means match all; also avoids [1..0] panic.
+            let inner = &pattern[1..pattern.len().saturating_sub(1)];
```
tiagolobocastro: this doesn't make any sense at all, if pattern is `*` then this will panic won't it?

k8s/plugin/src/resources/events.rs
tiagolobocastro: Why do some have it and others don't?
Anyhow, then I can infer at least then `rebuild_status_parser` isn't needed?

k8s/plugin/src/resources/events.rs
```diff
@@ -264,6 +341,61 @@ fn build_logql_filters() -> Vec<String> {
...
+            &pod_name,
+            vec![
+                "/bin/eventing-aggregator",
+                "--print-events",
+                &format!("--events-dir={EVENTS_VOLUME_DIR}"),
+            ],
+            &ap,
+        )
+        .await
+        .map_err(|e| anyhow!("Failed to exec into pod {pod_name}: {e}"))?;
```
niladrih, line 405: This API's behaviour needs to be tested for timeouts, etc. We might want to know at what point this starts to fail or require pagination.

k8s/plugin/src/resources/events.rs
rohan2794: [Fixed](https://github.com/openebs/mayastor-extensions/compare/a99f5451549dad7a07d1c2de04fc26a270efe6d4..bd650233b49c230bfde9a51e27d4cef82929396a), in case of *, return true

k8s/plugin/src/resources/events.rs
rohan2794: Yeah this needs to be tested. Added --since at source, the plugin computes Utc::now() - since and passes it as --since=<rfc3339> in the exec command. The aggregator binary parses each line's payload.metadata.timestamp and skips lines before the cutoff before they even enter the HashSet or get streamed over exec.

k8s/plugin/src/main.rs
```diff
@@ -105,9 +105,9 @@ async fn main() {
 
 impl CliArgs {
     async fn execute(self) -> Result<(), Error> {
-        // Initialise the REST client.
-        init_rest(&self.args).await?;
-
+        if self.operations.needs_rest_init() {
+            resources::init_rest(&self.args).await?;
```
abhilashshetty04, line 109: `init_rest` could have been imported here?

https://github.com/openebs/mayastor-extensions/pull/942/changes#diff-42f1a2eff9283e023b3db8a3104c73f269bca7100dee448589405aa8daef7d74

abhilashshetty04:
@rohan2794 , Can you share the sample of cmd execution. Is `--component` an arg and `pool`, `volume`, `replica` etc are its possible values?

rohan2794:
> @rohan2794 , Can you share the sample of cmd execution. Is `--component` an arg and `pool`, `volume`, `replica` etc are its possible values?
```

./kubectl-mayastor get events --category 
error: a value is required for '--category <CATEGORIES>' but none was supplied
  [possible values: pool, volume, nexus, replica, node, high-availability, nvme-path, host-initiator, io-engine-category, snapshot, clone]

For more information, try '--help'.
```

```
 ./kubectl-mayastor get events --category io-engine-category
Loki is not found, continuing...
Loki not found in cluster; falling back to eventing-aggregator pod volume.
 TIMESTAMP                       CATEGORY          ACTION    TARGET    NODE           COMPONENT 
 2026-07-21T06:34:44.275821845Z  IoEngineCategory  Start     mayastor  node-2-481267  IoEngine 
 2026-07-21T06:34:44.232410831Z  IoEngineCategory  Stop      mayastor  node-2-481267  IoEngine 
 2026-07-21T06:34:44.232223299Z  IoEngineCategory  Shutdown  mayastor  node-2-481267  IoEngine 
 2026-07-21T06:34:44.205252633Z  IoEngineCategory  Start     mayastor  node-1-481267  IoEngine 
 2026-07-21T06:34:44.158788038Z  IoEngineCategory  Stop      mayastor  node-1-481267  IoEngine 
 2026-07-21T06:34:44.158536520Z  IoEngineCategory  Shutdown  mayastor  node-1-481267  IoEngine 
 2026-07-21T06:34:41.637113617Z  IoEngineCategory  Start     mayastor  node-0-481267  IoEngine 
 2026-07-21T06:34:41.592434524Z  IoEngineCategory  Stop      mayastor  node-0-481267  IoEngine 
```

tiagolobocastro (approved):

abhilashshetty04 (approved):

k8s/plugin/src/main.rs
rohan2794: Yes, it can be used like that also

k8s/plugin/src/resources/events.rs
rohan2794: rebuild_status_parser cannot be removed — the PossibleValuesParser wrapper is the real reason these parsers exist (valid values in help + Unknown gating). component_parser differs only in the .map() body because Component lacks EnumString

PR #952, opened 2026-07-22
feat(supportability): collect eventing-aggregator events in system dump

rohan2794:
Add mayastor/events.ndjson to the system dump archive. Events are collected from Loki when available (eventing-aggregator in Loki mode) or by exec-ing --print-events into the aggregator pod volume otherwise.

- New collect/events.rs with two paths:
  - Loki: queries {app=eventing-aggregator} with |= mbus_event filter, strips the tracing prefix from each line so the output is valid NDJSON
  - Volume: execs /bin/eventing-aggregator --print-events --since=<rfc3339>
    and streams stdout bytes directly to disk (O(1) memory, no parsing)
- Both paths write events-source.txt recording Loki or the pod name used
- Add take_loki_client() to the Logger trait so the existing LokiClient from log collection is reused rather than opening a second connection
- Store since on SystemDumper and pass it to the volume exec so both paths respect the dump's --since window consistently
- Graceful skip with a warning log if no running aggregator pod is found
- Added LokiError::InvalidLabelSelector(String) variant
- label_selector_to_logql now returns Result<String, LokiError> —
  uses map + collect::<Result<Vec<_>, _>>() so the first malformed entry
  (missing =) returns an Err immediately instead of silently skipping it

### DIFF
--- a/k8s/supportability/src/collect/events.rs
+++ b/k8s/supportability/src/collect/events.rs
@@ -1,0 +1,181 @@
+use crate::collect::{
+    error::Error,
+    k8s_resources::client::ClientSet,
+    logs::LokiClient,
+    utils::{log, write_to_log_file},
+};
+
+use k8s_openapi::api::core::v1::Pod;
+use kube::{
+    api::{AttachParams, ListParams},
+    Api,
+};
+use std::{io::Write, path::Path};
+
+const EVENTS_LABEL_SELECTOR: &str = "app=eventing-aggregator";
+const EVENTS_CONTAINER: &str = "eventing-aggregator";
+const MBUS_EVENT_TYPE: &str = "mbus_event";
+const EVENTS_VOLUME_DIR: &str = "/var/events";
+
+fn build_logql_filters() -> Vec<String> {
+    vec![format!("|= \"{MBUS_EVENT_TYPE}\"")]
+}
+
+/// Collect events into `path/events.ndjson`.
+///
+/// Prefers Loki when a client is provided (eventing-aggregator was deployed in Loki mode).
+/// Falls back to exec-ing the eventing-aggregator pod volume when Loki is unavailable.
+/// `since` is passed to the aggregator exec so only events within the dump window are streamed.
+pub(crate) async fn collect_events_to_file(
+    loki_client: Option<&mut LokiClient>,
+    client_set: &ClientSet,
+    path: &Path,
+    since: humantime::Duration,
+) -> Result<(), Error> {
+    std::fs::create_dir_all(path)?;
+    match loki_client {
+        Some(client) => collect_events_from_loki(client, path).await,
+        None => collect_events_from_volume(client_set, path, since).await,
+    }
+}
+
+/// Collect events from Loki into `path/events.ndjson`.
+///
+/// Queries Loki for the eventing-aggregator stream and applies a `|= "mbus_event"` substring
+/// filter so only event lines are written (tracing/other log lines are dropped server-side).
+/// Loki lines carry a tracing prefix before the JSON payload; this function strips everything
+/// before the first `{` so the output file contains valid NDJSON.
+/// Lines are written page-by-page as they arrive from Loki — no full result set is held in
+/// memory, keeping allocation bounded to a single page (~3000 lines) at a time.
+async fn collect_events_from_loki(loki_client: &mut LokiClient, path: &Path) -> Result<(), Error> {
+    log("Collecting events from Loki...");
+
+    let events_path = path.join("events.ndjson");
+    let mut file = std::fs::File::create(&events_path)?;
+    let mut count = 0usize;
+
+    loki_client.set_logql_filters(build_logql_filters());
+    loki_client
+        .fetch_lines_paged(
+            EVENTS_LABEL_SELECTOR.to_string(),
+            EVENTS_CONTAINER.to_string(),
+            |page| {
+                for line in page {
+                    // Loki lines have a tracing prefix before the JSON payload
+                    // (e.g. "2024-01-01 INFO eventing_aggregator: {...}").
+                    // Strip everything before the first '{' so the file contains
+                    // the same raw NDJSON format as the pod-volume path.
+                    if let Some(json_start) = line.find('{') {
+                        writeln!(file, "{}", line[json_start..].trim_end())?;
+                        count += 1;
+                    }
+                }
+                Ok(())
+            },
+        )
+        .await
+        .map_err(|e| Error::Generic(format!("Failed to fetch events from Loki: {e:?}")))?;
+
+    std::fs::write(path.join("events-source.txt"), "loki\n")
+        .map_err(|e| Error::Generic(format!("Failed to write events-source.txt: {e}")))?;
+
+    log(format!("Collected {count} events from Loki"));
+    let _ = write_to_log_file(format!("Events written to {}\n", events_path.display()));
+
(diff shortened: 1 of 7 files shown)


k8s/supportability/src/collect/events.rs
```diff
@@ -0,0 +1,158 @@
...
+            &ap,
+        )
+        .await
+        .map_err(|e| Error::Generic(format!("Failed to exec into pod {pod_name}: {e}")))?;
+
+    let mut stdout = attached
+        .stdout()
+        .ok_or_else(|| Error::Generic("No stdout from pod exec".to_string()))?;
+
+    let events_path = path.join("events.ndjson");
```
Abhinandan-Purkait, line 164: What is ndjson?

Abhinandan-Purkait (approved):

k8s/supportability/src/collect/events.rs
rohan2794: NDJSON (Newline-Delimited JSON) is one complete JSON object per line, no wrapping array:

```
{"type": "mbus_event", "payload": {"component": "volume", ...}}
{"type": "mbus_event", "payload": {"component": "io-engine", ...}}
{"type": "mbus_event", "payload": {"component": "pool", ...}}
```

Each line is independently valid JSON. No commas between records, no outer [ ].

JSON stores a collection as a single document — one root array wrapping everything:

```
[
  {"type": "mbus_event", "payload": {"component": "volume", ...}},
  {"type": "mbus_event", "payload": {"component": "io-engine", ...}},
  {"type": "mbus_event", "payload": {"component": "pool", ...}}
]
```

The entire file must be parsed before you can read a single record. If the file is 500 MB, you load 500 MB into memory.

k8s/supportability/src/collect/events.rs
```diff
@@ -0,0 +1,158 @@
...
+    std::fs::write(path.join("events-source.txt"), "loki\n")
+        .map_err(|e| Error::Generic(format!("Failed to write events-source.txt: {e}")))?;
+
+    log(format!("Collected {} events from Loki", lines.len()));
+    let _ = write_to_log_file(format!("Events written to {}\n", events_path.display()));
+
+    Ok(())
+}
+
+async fn collect_events_from_volume(
```
abhilashshetty04, line 97: missing doc

k8s/supportability/src/collect/events.rs
```diff
@@ -0,0 +1,158 @@
...
+    since: humantime::Duration,
+) -> Result<(), Error> {
+    std::fs::create_dir_all(path)?;
+    match loki_client {
+        Some(client) => collect_events_from_loki(client, path).await,
+        None => collect_events_from_volume(client_set, path, since).await,
+    }
+}
+
+async fn collect_events_from_loki(loki_client: LokiClient, path: &Path) -> Result<(), Error> {
```
abhilashshetty04: missing doc

abhilashshetty04 (approved):

k8s/supportability/src/collect/events.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    let lines = loki_client
+        .with_logql_filters(build_logql_filters())
+        .fetch_lines(
+            EVENTS_LABEL_SELECTOR.to_string(),
+            EVENTS_CONTAINER.to_string(),
+            0, // no limit — collect all stored events
+        )
+        .await
+        .map_err(|e| Error::Generic(format!("Failed to fetch events from Loki: {e:?}")))?;
```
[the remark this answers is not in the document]
rohan2794: The code compiles and runs correctly as written.
 when you call a &mut self method on the return value of a by-value method in a chain, the compiler implicitly places the temporary in a slot and mutably borrows it. The temporary lives for the entire statement, including across the .await.

k8s/supportability/src/collect/events.rs
rohan2794: added comment

k8s/supportability/src/collect/events.rs
```diff
@@ -0,0 +1,158 @@
...
+async fn collect_events_from_loki(loki_client: LokiClient, path: &Path) -> Result<(), Error> {
+    log("Collecting events from Loki...");
+
+    let lines = loki_client
+        .with_logql_filters(build_logql_filters())
+        .fetch_lines(
+            EVENTS_LABEL_SELECTOR.to_string(),
+            EVENTS_CONTAINER.to_string(),
+            0, // no limit — collect all stored events
+        )
```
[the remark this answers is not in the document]
rohan2794: fixed

k8s/supportability/src/collect/events.rs
```diff
@@ -0,0 +1,158 @@
...
+    };
+
+    log(format!("Collecting events from pod volume ({pod_name})..."));
+
+    let ap = AttachParams::default()
+        .stdin(false)
+        .stdout(true)
+        .stderr(false);
+
+    let since_cutoff = chrono::Utc::now() - chrono::Duration::from_std(*since).unwrap_or_default();
```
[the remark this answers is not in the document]
rohan2794: fixed

k8s/supportability/src/collect/logs/mod.rs
```diff
@@ -247,6 +251,9 @@ pub(crate) trait Logger {
         &self,
         logging_label_selectors: String,
     ) -> Result<HashSet<LogResource>, LogError>;
+    /// Take the inner Loki client, if one was successfully connected at construction time.
+    /// Returns `None` when Loki was not found or could not be reached.
+    fn take_loki_client(&mut self) -> Option<LokiClient>;
```
tiagolobocastro: What if we need the Loki client for something later, then we won't have it anymore?
Can't we just clone it?

rohan2794: @tiagolobocastro  Cloning LokiClient isn't straightforward — the inner kube_proxy::LokiClient wraps a tower service (and in the port-forward case, a live port-forward connection) that doesn't implement Clone.

We can change take_loki_client → loki_client_mut returning Option<&mut LokiClient>, and update the call chain to work through a mutable reference.

tiagolobocastro: I see, what about building a new client?

rohan2794: Using new loki client will be simple and clean but i thought of using existing one . Should i build new loki client ?

rohan2794: @tiagolobocastro replaced take() with mutable reference https://github.com/openebs/mayastor-extensions/pull/952/changes/65875c7c2c84da36d8d199e76f42e9fc4b9d1780

k8s/supportability/src/collect/logs/loki.rs
```diff
@@ -220,6 +221,10 @@ impl LokiClient {
         self
     }
 
+    pub fn set_logql_filters(&mut self, filters: Vec<String>) {
```
tiagolobocastro, line 226: missing doc comments

tiagolobocastro (commented):
Please squash the commits

k8s/supportability/src/collect/logs/loki.rs
rohan2794: added comment

rohan2794:
> Please squash the commits

done

tiagolobocastro (approved):

PR #953, opened 2026-07-22
feat(chart): add custom envs for io-engine, update readme

mk-raven:
## Description

Add support for custom environment variables in the `io-engine` container through a new Helm value:

```yaml
io_engine:
  extraEnv: []
```

The value uses the standard Kubernetes `EnvVar` list format.

Example:

```yaml
io_engine:
  extraEnv:
    - name: IOBUF_SMALL_POOL_COUNT
      value: "32768"
    - name: IOBUF_LARGE_POOL_COUNT
      value: "8192"
```

It also supports Kubernetes `valueFrom` sources:

```yaml
io_engine:
  extraEnv:
    - name: CUSTOM_SETTING
      valueFrom:
        secretKeyRef:
          name: io-engine-settings
          key: custom-setting
```

When `io_engine.extraEnv` is empty or not configured, the rendered `io-engine` DaemonSet remains unchanged.

## Motivation and Context

The Helm chart currently defines a fixed set of environment variables for the `io-engine` container but does not provide a supported way to add additional variables.

This is required when io-engine or SPDK settings are exposed only through environment variables.

For example, increasing the NVMe/TCP shared buffer count:

```yaml
io_engine:
  nvme:
    tcp:
      numSharedBuf: "3000"
```

can produce the following warnings and errors:

```text
The num_shared_buffers value (3000) is larger than the available iobuf pool size (2048). Please increase the iobuf pool sizes.

Failed to populate 'nvmf_TCP' iobuf large buffer cache at 32/64 entries. You may need to increase spdk_iobuf_opts.large_pool_count (2048).

Unable to reserve the full number of buffers for the pg buffer cache.
```

The issue can be resolved by increasing the SPDK iobuf pools:

```text
IOBUF_SMALL_POOL_COUNT=32768
IOBUF_LARGE_POOL_COUNT=8192
```

Currently, these values must be added manually:

```bash
kubectl -n openebs set env daemonset/mayastor-io-engine \
  IOBUF_SMALL_POOL_COUNT=32768 \
  IOBUF_LARGE_POOL_COUNT=8192
```

Because the DaemonSet is managed by Helm, these manual changes can be removed during a Helm upgrade or reconciliation.

Adding a generic `io_engine.extraEnv` value allows these settings, and other future io-engine settings, to be managed declaratively without adding a dedicated Helm value for every environment variable.

Fixes #ISSUE_NUMBER.

## Regression

No.

This is a new optional configuration value. It defaults to an empty list and does not change the rendered resources or existing io-engine behavior unless explicitly configured.

## How Has This Been Tested?

The chart was rendered with the following values:

```yaml
io_engine:
  nvme:
    tcp:
      numSharedBuf: "3000"

  extraEnv:
    - name: IOBUF_SMALL_POOL_COUNT
      value: "32768"
    - name: IOBUF_LARGE_POOL_COUNT
      value: "8192"
```

The rendered `mayastor-io-engine` DaemonSet contains:

```yaml
- name: IOBUF_SMALL_POOL_COUNT
  value: "32768"
- name: IOBUF_LARGE_POOL_COUNT
  value: "8192"
```

The following cases were tested:

1. Rendering the chart with literal custom environment variable values.
2. Rendering the chart with an environment variable using `valueFrom.secretKeyRef`.
3. Rendering the chart without `io_engine.extraEnv`.
4. Running Helm lint against the chart.
5. Deploying the updated chart in a Kubernetes cluster and verifying the environment variables inside the io-engine pod.
6. Verifying that the previous SPDK iobuf pool warnings were no longer present after configuring the increased pool sizes.

Test commands:

```bash
helm lint ./chart
```

```bash
helm template mayastor ./chart \
  --namespace openebs \
  --set io_engine.nvme.tcp.numSharedBuf=3000 \
  --set io_engine.extraEnv[0].name=IOBUF_SMALL_POOL_COUNT \
  --set-string io_engine.extraEnv[0].value=32768 \
  --set io_engine.extraEnv[1].name=IOBUF_LARGE_POOL_COUNT \
  --set-string io_engine.extraEnv[1].value=8192
```

Deployment environment:

* Kubernetes: `v1.31.4`
* Operating system: Ubuntu 24.04 LTS
* Mayastor chart: `2.11.1`
* Transport: NVMe/TCP

## Types of changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [x] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

* [x] My code follows the code style of this project.
* [x] My change requires a change to the documentation.
* [x] I have updated the documentation accordingly.
* [ ] I have added unit tests to cover my changes.

### DIFF
--- a/chart/values.schema.json
+++ b/chart/values.schema.json
@@ -1,0 +1,2529 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "type": "object",
+    "properties": {
+        "agents": {
+            "type": "object",
+            "properties": {
+                "core": {
+                    "type": "object",
+                    "properties": {
+                        "allowNonPersistentDevlink": {
+                            "type": "boolean"
+                        },
+                        "capacity": {
+                            "type": "object",
+                            "properties": {
+                                "thin": {
+                                    "type": "object",
+                                    "properties": {
+                                        "poolCommitment": {
+                                            "type": "string"
+                                        },
+                                        "snapshotCommitment": {
+                                            "type": "string"
+                                        },
+                                        "volumeCommitment": {
+                                            "type": "string"
+                                        },
+                                        "volumeCommitmentInitial": {
+                                            "type": "string"
+                                        }
+                                    }
+                                }
+                            }
+                        },
+                        "encryptedPoolsSoftScheduling": {
+                            "type": "boolean"
+                        },
+                        "logLevel": {
+                            "type": "string"
+                        },
+                        "maxCreateVolume": {
+                            "type": "integer"
+                        },
+                        "minTimeouts": {
+                            "type": "boolean"
+                        },
+                        "nodeSelector": {
+                            "type": "object"
+                        },
+                        "poolClusterSize": {
+                            "type": "string"
+                        },
+                        "priorityClassName": {
+                            "type": "string"
+                        },
+                        "rebuild": {
+                            "type": "object",
+                            "properties": {
+                                "maxConcurrent": {
+                                    "type": "string"
+                                },
+                                "partial": {
+                                    "type": "object",
+                                    "properties": {
+                                        "enabled": {
+                                            "type": "boolean"
+                                        },
+                                        "waitPeriod": {
+                                            "type": "string"
+                                        }
+                                    }
+                                }
+                            }
+                        },
+                        "requestTimeout": {
+                            "type": "null"
+                        },
+                        "resources": {
+                            "type": "object",
+                            "properties": {
+                                "limits": {
+                                    "type": "object",
+                                    "properties": {
+                                        "cpu": {
+                                            "type": "string"
+                                        },
+                                        "memory": {
+                                            "type": "string"
+                                        }
+                                    }
+                                },
+                                "requests": {
+                                    "type": "object",
+                                    "properties": {
+                                        "cpu": {
+                                            "type": "string"
+                                        },
+                                        "memory": {
+                                            "type": "string"
+                                        }
+                                    }
+                                }
+                            }
+                        },
+                        "tolerations": {
+                            "type": "array"
+                        },
+                        "volumeHealth": {
+                            "type": "boolean"
+                        }
+                    }
+                },
+                "ha": {
+                    "type": "object",
+                    "properties": {
+                        "cluster": {
+                            "type": "object",
+                            "properties": {
+                                "logLevel": {
+                                    "type": "string"
+                                },
+                                "resources": {
+                                    "type": "object",
+                                    "properties": {
+                                        "limits": {
+                                            "type": "object",
+                                            "properties": {
+                                                "cpu": {
+                                                    "type": "string"
+                                                },
+                                                "memory": {
+                                                    "type": "string"
+                                                }
+                                            }
+                                        },
+                                        "requests": {
+                                            "type": "object",
+                                            "properties": {
+                                                "cpu": {
+                                                    "type": "string"
+                                                },
+                                                "memory": {
+                                                    "type": "string"
+                                                }
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                        },
+                        "enabled": {
+                            "type": "boolean"
+                        },
+                        "node": {
+                            "type": "object",
+                            "properties": {
+                                "logLevel": {
+                                    "type": "string"
+                                },
+                                "nodeSelector": {
+                                    "type": "object"
+                                },
+                                "port": {
+                                    "type": "integer"
+                                },
+                                "priorityClassName": {
+                                    "type": "string"
+                                },
+                                "resources": {
+                                    "type": "object",
+                                    "properties": {
+                                        "limits": {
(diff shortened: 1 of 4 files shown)


chart/values.schema.json
```diff
@@ -0,0 +1,2529 @@
+{
```
Abhinandan-Purkait, line 1: Why do we need this file?

mk-raven: in a Helm chart defines the allowed structure, types, and validation rules for values.yaml.

It is useful for:
- Validation before deployment — Helm rejects invalid values during helm install, helm upgrade, helm lint, and helm template.
- Type checking — prevents mistakes such as using "true" instead of true, or a string instead of a number.
- Required fields — ensures mandatory configuration is provided.
- Allowed values — limits fields to supported options using enum.
- Range and format checks — validates minimum values and similar constraints.

https://helm.sh/docs/topics/charts/#schema-files

Abhinandan-Purkait: Okay, then we need to make sure this is updated everytime something changes, we would need to modify the CI to catch this as well otherwise this will not be maintained properly.

mk-raven: You could use helm plugin `helm schema` for generate values.schema.json file.
https://github.com/losisin/helm-values-schema-json
Its integrate with github actions.

tiagolobocastro: Let's not add these changes I'd say
None of the engines use this approach today, so we'd need to figure out if we want to start supporting this, and ensure it's kept up to date

tiagolobocastro (commented):
How about having explicit variables for this, like `numSharedBuf`?

mk-raven:
> How about having explicit variables for this, like `numSharedBuf`?

My concern is that explicit values do not fully solve the operational use case. In large, high-throughput big-data clusters, io-engine and SPDK often require deployment-specific tuning, and some parameters are available only through environment variables.

Adding an explicit Helm value for every possible parameter means that each new tuning requirement needs a chart change and release. Otherwise, operators must patch the Helm-managed DaemonSet, and the changes are lost during upgrade or reconciliation.

I suggest supporting:

```
io_engine:
  extraEnv: []
```

tiagolobocastro:
> > How about having explicit variables for this, like `numSharedBuf`?
> 
> My concern is that explicit values do not fully solve the operational use case. In large, high-throughput big-data clusters, io-engine and SPDK often require deployment-specific tuning, and some parameters are available only through environment variables.
> 
> Adding an explicit Helm value for every possible parameter means that each new tuning requirement needs a chart change and release. Otherwise, operators must patch the Helm-managed DaemonSet, and the changes are lost during upgrade or reconciliation.
> 
> I suggest supporting:
> 
> ```
> io_engine:
>   extraEnv: []
> ```

That makes sense,  though `extraEnv` may clash with the explicit variables, what do we do then?

mk-raven:
> > > How about having explicit variables for this, like `numSharedBuf`?
> > 
> > 
> > My concern is that explicit values do not fully solve the operational use case. In large, high-throughput big-data clusters, io-engine and SPDK often require deployment-specific tuning, and some parameters are available only through environment variables.
> > Adding an explicit Helm value for every possible parameter means that each new tuning requirement needs a chart change and release. Otherwise, operators must patch the Helm-managed DaemonSet, and the changes are lost during upgrade or reconciliation.
> > I suggest supporting:
> > ```
> > io_engine:
> >   extraEnv: []
> > ```
> 
> That makes sense, though `extraEnv` may clash with the explicit variables, what do we do then?

For this case I will suggest to validate predefined envs.
For example:
```

{{- $reservedEnv := dict
  "RUST_LOG" true
  "NVMF_TCP_NUM_SHARED_BUFFERS" true
  "NVMF_TCP_MAX_QUEUE_DEPTH" true
  "NVMF_TCP_MAX_QPAIRS_PER_CTRL" true
-}}

{{- range .Values.io_engine.extraEnv }}
  {{- if hasKey $reservedEnv .name }}
    {{- fail (printf "io_engine.extraEnv contains reserved environment variable %q; use the corresponding io_engine Helm value instead" .name) }}
  {{- end }}
{{- end }}
```

Then the custom variables can be rendered only after validation:

```
{{- with .Values.io_engine.extraEnv }}
{{- toYaml . | nindent 12 }}
{{- end }}
```

mk-raven:
if my suggestions will be ok, I could add changes to my PR.

tiagolobocastro:
Sorry @mk-raven I'm away so have not been able to keep up with these

Something like that sounds good;
would it be possible to have it as a map?
like ["NVMF_TCP_NUM_SHARED_BUFFERS": .Values.io-engine.xxxx]

In this case then can you add the `IOBUF_SMALL_POOL_COUNT` as a .Values.io_engine.xxxx ?
We can keep the env for the "unknown" ones

Also on the values schema let's do that as a separate PR please and we'd also need the CI to ensure it's not left outdated
Also there's workflows which modify the values.yaml

Thank you!

PR #954, opened 2026-07-28
feat(plugin): add --from-file for offline event analysis from system dump

rohan2794:
Allows querying a local events.ndjson without a live cluster. --since applies as a client-side timestamp filter; all typed filters work as usual. 

```
Usage:
kubectl mayastor get events --from-file ./events.ndjson
kubectl mayastor get events --from-file ./events.ndjson --since 48h --component io-engine --state degraded
```

### DIFF
--- a/k8s/plugin/src/resources/events.rs
+++ b/k8s/plugin/src/resources/events.rs
@@ -4,6 +4,7 @@ use clap::builder::TypedValueParser;
 use events_api::event::{
     Component, EventAction, EventCategory, EventDetails, EventMessage, RebuildStatus,
 };
+use futures::{Stream, StreamExt};
 use k8s_openapi::api::core::v1::Pod;
 use kube::{
     api::{AttachParams, ListParams},
@@ -12,6 +13,7 @@ use kube::{
 use plugin::resources::utils::{optional_cell, print_table, CreateRow, GetHeaderRow, OutputFormat};
 use prettytable::{row, Row};
 use serde::Serialize;
+use std::path::PathBuf;
 use strum::IntoEnumIterator;
 use supportability::KubeConfigArgs;
 use tokio::io::{AsyncBufRead, AsyncBufReadExt, BufReader};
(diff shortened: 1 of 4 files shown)


k8s/plugin/src/resources/events.rs
```diff
@@ -89,6 +90,12 @@ pub struct EventsArgs {
     /// Loki tenant ID / X-Scope-OrgID header (loki source only).
     #[arg(long, default_value = "openebs")]
     tenant_id: String,
+
+    /// Read events from a local NDJSON file (e.g. extracted from a system dump archive)
+    /// instead of querying a live cluster. No cluster connection is made.
```
tiagolobocastro: Please check --help, also for the existing ones.
With clap `{n}` is needed for newlines.

```suggestion
    /// instead of querying a live cluster. No cluster connection is made. {n}
```

k8s/plugin/src/resources/events.rs
```diff
@@ -254,6 +266,26 @@ impl EventsArgs {
         // utility — that subsystem is only wired up in system-dump context.
         supportability::init_no_log_file();
 
+        // --from-file: read a local NDJSON dump; no cluster connection needed.
+        if let Some(path) = &self.from_file {
+            let file = tokio::fs::File::open(path)
+                .await
+                .map_err(|e| anyhow!("Cannot open {}: {e}", path.display()))?;
+            let records = read_events_from_reader(BufReader::new(file)).await;
```
tiagolobocastro: Is this loading the entire file contents to memory one-shot?

rohan2794: @tiagolobocastro  Yes, read_events_from_reader collects all records into a Vec<EventRecord> before returning, so the full parsed output lives in memory. However, it reads line-by-line via lines.next_line(), so the raw file bytes are streamed — only the parsed EventRecord structs accumulate.

k8s/plugin/src/resources/events.rs
rohan2794: fixed

k8s/plugin/src/resources/events.rs
rohan2794: changed to events stream https://github.com/openebs/mayastor-extensions/pull/954/changes/109628e305cb38bd760d55285ac0d5b41649cb54

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

PR #962, opened 2026-08-04
feat(chart): add coresFromCgroup option for io-engine

therayquaza:
## Summary
- Adds `io_engine.coresFromCgroup` (default `false`) to `values.yaml`.
- When enabled, appends `--cores-from-cgroup` to the io-engine daemonset args, alongside the existing static `-l` core list.
- Companion to openebs/mayastor#2050, which adds the `--cores-from-cgroup` flag to io-engine so it derives its reactor core list from the pod's kubelet-assigned cgroup cpuset (addresses openebs/mayastor#1458).

## Test plan
- [x] `helm lint`
- [x] `helm template` with `coresFromCgroup=true` and default (`false`) — flag present/absent as expected
- [x] Manual deploy on a kubelet node with CPU Manager static policy enabled (Cgroup V2 only)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -314,6 +314,7 @@ Each Secret must contain `tls.crt`, `tls.key`, and `ca.crt` keys.
 | image.&ZeroWidthSpace;repo | Image registry's namespace | `"openebs"` |
 | image.&ZeroWidthSpace;tag | Release tag for our images | `"develop"` |
 | io_engine.&ZeroWidthSpace;coreList | If not empty, overrides the cpuCount and explicitly sets the list of cores. Example: --set='io_engine.coreList={30,31}' | `[]` |
+| io_engine.&ZeroWidthSpace;coresFromCgroup | Derive the io-engine core list from the pod's cgroup cpuset instead of cpuCount/coreList. See: https://github.com/openebs/mayastor/issues/1458 | `false` |
 | io_engine.&ZeroWidthSpace;cpuCount | The number of cores that each io-engine instance will bind to. | `"2"` |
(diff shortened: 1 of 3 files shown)


Abhinandan-Purkait (approved):

Scorplionn:
+1 — we hit exactly this and independently built the same thing before finding
this PR (see #1015, closed in favour of it).

Environment: Kubernetes v1.35, Rocky Linux 10 (kernel 6.12), Mayastor v2.11.1,
32-core nodes with

    cpuManagerPolicy: static
    cpuManagerPolicyOptions:
      strict-cpu-reservation: "true"
    reservedSystemCPUs: "0,1"

Every alternative we tried fails:

- the generated core list starts at 1 (`untilStep 1 (cpuCount+1)`), which
  collides with `reservedSystemCPUs` — `EAL: Cannot set affinity`;
- setting `coreList` to match what the CPU manager granted works only by
  coincidence: it picks the lowest free cores, so the values line up until
  another Guaranteed pod claims them first;
- adding the io-engine cores to `reservedSystemCPUs` removes them from every
  container's cpuset, io-engine included — same failure;
- dropping `-l` entirely doesn't help: EAL falls back to core 0, which is
  reserved. Verified — `core_list: None` followed by `EAL: Cannot set affinity`.

With #1014 merged the pod can reach Guaranteed through values alone, so the
cpuset this PR reads is the exclusively assigned cores rather than the shared
pool.

Happy to test on our cluster once openebs/mayastor#2050 lands.

abhilashshetty04 (approved):

PR #963, opened 2026-08-07
fix(supportability): fix Loki auto-discovery timeout, raise default --timeout

datacore-gjanoti:
## Description
LokiClient's auto-discovery path (no explicit `--loki-endpoint`, the common case) never called `.with_timeout()` on its `ConfigBuilder`, so it silently used the builder's own hardcoded 5s default instead of `--timeout`, too short for large windows (a 30-day search needs 60s+). Fixed.

Also raised the shared `--timeout` default from 10s to 30s, since it's shared across etcd, REST, Loki, and k8s log fetches (each call still gets its own fresh timeout window, not one shared deadline).

Earlier this PR also added a manual timeout on the k8s-native per-container log fetch, since a hung kubelet was seen blocking `dump system` indefinitely. Removed after review, since `kube::Client` already has a built-in 295s read timeout at the transport layer, making the manual wrapper redundant. Will revisit if the underlying hang is found to slip past that built-in timeout.

## Motivation and Context
Investigating repeated `dump system` timeouts/hangs against a real cluster.

## Regression
No, only widens/corrects timeout windows; normal, responsive calls are unaffected.

## How Has This Been Tested?
Verified live: a 30-day `--since` with a 60s `--timeout` completes successfully end-to-end where it previously failed with `Loki(Tower(Elapsed(())))` on every service. `cargo check -p supportability` passes after the manual-timeout removal.

## Types of changes

## Checklist:

### DIFF
--- a/k8s/supportability/src/collect/logs/loki.rs
+++ b/k8s/supportability/src/collect/logs/loki.rs
@@ -153,10 +153,17 @@ impl LokiClient {
     ) -> Option<Self> {
         let (uri, client) = match uri {
             None => {
+                // Without this, ConfigBuilder's own Default impl applies a
+                // hardcoded 5s timeout here regardless of the CLI's --timeout
+                // value - fine for a short --since window, but Loki
+                // genuinely needs longer than that to search a large one
+                // (e.g. 30d), so every query timed out well before Loki
+                // could respond.
                 let (uri, svc) = match kube_proxy::ConfigBuilder::default_loki()
                     .with_kube_config(kubeconfig_args.path.clone())
                     .with_context(kubeconfig_args.opts.context.clone())
                     .with_target_mod(|t| t.with_namespace(namespace))
+                    .with_timeout(*timeout)
                     .build()
                     .await
                 {
--- a/k8s/supportability/src/lib.rs
+++ b/k8s/supportability/src/lib.rs
@@ -19,7 +19,7 @@ pub use collect::logs::{LokiClient, LokiError};
 #[derive(Debug, Clone, clap::Args)]
 pub struct SupportArgs {
     /// Specifies the timeout value to interact with other modules of system
-    #[clap(global = true, long, short, default_value = "10s")]
+    #[clap(global = true, long, short, default_value = "30s")]
     timeout: humantime::Duration,
 
     /// Period states to collect all logs from last specified duration


k8s/supportability/src/collect/logs/k8s_log.rs
```diff
@@ -50,15 +50,36 @@ impl From<kube::Error> for K8sLoggerError {
 #[derive(Clone)]
 pub(crate) struct K8sLoggerClient {
     k8s_client: ClientSet,
+    /// How far back to fetch logs from, mirroring the same `--since` this
+    /// module's Loki-backed counterpart already honors - without this, a
+    /// direct kube-apiserver log fetch (used whenever Loki isn't
+    /// available/configured) always streamed the entire available log
+    /// buffer with no way to limit it.
+    since: humantime::Duration,
```
tiagolobocastro: This is a break in behaviour
It's actually been very useful to get the entire logs, so not sure how we want to handle this.. for now let's leave as is and collect the entire K8s logs?

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
Looks fair to me

Abhinandan-Purkait:
What's the default timeout?

datacore-gjanoti:
> 10s (the `--timeout` default), same value the Loki path already used.

10s (the --timeout default), same value the Loki path already used.

k8s/supportability/src/collect/logs/k8s_log.rs
```diff
@@ -229,10 +250,38 @@ impl K8sLoggerClient {
         container_name: &str,
         mut writer: W,
         previous_logs: bool,
+    ) -> Result<(), K8sLoggerError> {
+        match tokio::time::timeout(
```
tiagolobocastro: if we use the timeout for the K8s client requests, then this is not needed

datacore-gjanoti: > if we use the timeout for the K8s client requests, then this is not needed

I looked into this exact question before, for Loki/REST. The Kubernetes client does have a built-in timeout, but we tried to test if it alone would catch this kind of hang and couldn't get a clear answer, the cluster kept failing for unrelated reasons during our test. The manual timeout is the one we know for sure works, since it's what I saw fix the actual bug live. Should we remove it and trust the built-in timeout only, or keep both to be safe?

k8s/supportability/src/collect/logs/k8s_log.rs
datacore-gjanoti: Reverted in 69191ee , the k8s-native path is back to collecting full logs, no `--since` limit. The timeout fix stays as-is.

k8s/supportability/src/collect/logs/k8s_log.rs
tiagolobocastro: If it has it then trust it, if we find it doesn't work then we can fix

datacore-gjanoti: Done, removed the manual timeout and merged fetch_logs_stream back into write_logs_stream. kube::Client already has a built-in 295s read timeout, so trusting that for now. Will add it back if we see this hang again in practice.

PR description updated to match.

tiagolobocastro: Oh so we're not overriding the K8s timeout of 295s !
This may be something we may want to revisit

k8s/supportability/src/collect/logs/k8s_log.rs
tiagolobocastro: Also we could make this an Option<>, disabled by default

k8s/supportability/Cargo.toml
tiagolobocastro: This isn't required anymore?

tiagolobocastro (commented):
Please squash commits

k8s/supportability/src/collect/logs/k8s_log.rs
datacore-gjanoti: You're right, and it's worse than a theoretical gap - reproduced it live today. Ran a dump against a cluster with one unresponsive kubelet, and it sat stuck 13+ minutes with no error, last log line showing it had moved on to that node's containers and never came back. The 295s built-in timeout never fired.

Reinstated the manual timeout.

k8s/supportability/Cargo.toml
datacore-gjanoti: It is now - reinstated the manual tokio::time::timeout call this depends on (see the other [thread](https://github.com/openebs/mayastor-extensions/pull/963#discussion_r3850387008)), since the built-in K8s client timeout turned out not to catch the hang it was added for.

k8s/supportability/src/collect/logs/k8s_log.rs
tiagolobocastro: As I said previously, please remove this manual timeout
If there's a bug causing a hang it should be dealt separately, as it likely may affect any other component using the kube client, thanks

tiagolobocastro (changes requested):

k8s/supportability/src/collect/logs/k8s_log.rs
datacore-gjanoti: > Also we could make this an Option<>, disabled by default

@tiagolobocastro - Thanks for the suggestion. This was actually already removed a little while back, following your earlier comment about collecting the entire K8s logs. If you'd still like it back as an opt-in Option<> (disabled by default), happy to add it , just let me know and I'll go ahead. Wanted to check with you first before making any change.

k8s/supportability/src/collect/logs/k8s_log.rs
datacore-gjanoti: > As I said previously, please remove this manual timeout If there's a bug causing a hang it should be dealt separately, as it likely may affect any other component using the kube client, thanks

Fair point, done. Removed the manual timeout again, back to relying on the built-in K8s client timeout. Agreed a systemic hang like this is better tracked and fixed at that layer rather than patched locally here.

k8s/supportability/Cargo.toml
datacore-gjanoti: Correction to my earlier reply: it's unused again now that the manual timeout is [removed](https://github.com/openebs/mayastor-extensions/pull/963#discussion_r3851910400). Dropped the tokio dependency for real this time.

k8s/supportability/src/collect/logs/k8s_log.rs
tiagolobocastro: Let's do that on another PR, need to figure out how to expose that to user as well

tiagolobocastro (commented):
Please squash commits

datacore-gjanoti:
> Please squash commits

done, squashed. 
All review comments taken care of, pls check if can be approved now.

tiagolobocastro (approved):

tiagolobocastro:
@mergify backport release/2.12

PR #968, opened 2026-08-10
feat(plugin): add NATS JetStream source for get events

rohan2794:
Adds --nats-endpoint to kubectl mayastor get events:
- No value → auto-discovers the NATS service via K8s port-forward (app.kubernetes.io/name=nats, named port client); errors if not found.
- Explicit URL → connects directly; errors on failure. No fallback.

Creates an ephemeral push consumer with DeliverPolicy::ByStartTime so --since is pushed to the NATS server, limiting messages delivered to the plugin. AckPolicy::None is used since reads are fire-and-forget. Subscription is closed automatically when the consumer is dropped.

NATS messages carry raw EventMessage JSON; the mbus_event envelope is added on read so --filter dot-path traversal works identically to the Loki and file paths. All existing filters (--category, --action, --volume, --filter, --limit, etc.) are applied via finalize_and_print.

### DIFF
--- a/k8s/plugin/Cargo.toml
+++ b/k8s/plugin/Cargo.toml
@@ -45,6 +45,8 @@ snafu = { workspace = true }
 # These dependencies are required to build that file.
 futures = { workspace = true }
 chrono = { workspace = true }
+async-nats = { workspace = true }
+time = { workspace = true }
 parse-size = { workspace = true }
 strum_macros = { workspace = true }
 serde = { workspace = true }
--- a/k8s/plugin/src/resources/events.rs
+++ b/k8s/plugin/src/resources/events.rs
@@ -23,6 +23,7 @@ const EVENTS_LABEL_SELECTOR: &str = "app=eventing-aggregator";
 const MBUS_EVENT_TYPE: &str = "mbus_event";
 const DEFAULT_LIMIT: usize = 1000;
 const EVENTS_VOLUME_DIR: &str = "/var/events";
+const NATS_STREAM: &str = "events-stream";
 
 /// Args for `kubectl mayastor get events`.
 #[derive(Debug, clap::Args)]
@@ -95,6 +96,17 @@ pub struct EventsArgs {
     /// instead of querying a live cluster. All other filters work normally.
     #[arg(long, value_name = "PATH", conflicts_with = "loki_endpoint")]
     from_file: Option<PathBuf>,
+
+    /// Read events directly from NATS JetStream instead of Loki or the eventing-aggregator pod volume.{n}
+    /// Pass without a value to auto-discover the NATS service via the K8s cluster,{n}
+    /// or pass a URL (e.g. nats://host:4222) to connect directly.
+    #[arg(
+        long,
+        num_args = 0..=1,
+        default_missing_value = "",
+        conflicts_with_all = ["loki_endpoint", "from_file"],
+    )]
+    nats_endpoint: Option<String>,
 }
 
 /// A single parsed event record.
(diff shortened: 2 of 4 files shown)


k8s/plugin/src/resources/events.rs
```diff
@@ -95,6 +98,16 @@ pub struct EventsArgs {
     /// instead of querying a live cluster. All other filters work normally.
     #[arg(long, value_name = "PATH", conflicts_with = "loki_endpoint")]
     from_file: Option<PathBuf>,
+
+    /// NATS base URL. If omitted, NATS is auto-discovered via the K8s service;{n}
+    /// if discovery also fails, it will error out.
+    #[arg(
+        long,
+        num_args = 0..=1,
+        default_missing_value = "",
```
tiagolobocastro, line 106: Why would this be needed?

```suggestion
```

k8s/plugin/Cargo.toml
```diff
@@ -45,6 +45,8 @@ snafu = { workspace = true }
 # These dependencies are required to build that file.
 futures = { workspace = true }
 chrono = { workspace = true }
+async-nats = { workspace = true }
+time = "0.3"
```
tiagolobocastro: workspace too

k8s/plugin/src/resources/events.rs
```diff
@@ -462,6 +494,107 @@ async fn fetch_from_volume(
...
+        TargetSelector::ServiceLabel(NATS_LABEL.to_string()),
+        NATS_CLIENT_PORT,
+        namespace,
+    );
+    let (port, handle) = PortForward::new(target, None, kube_client)
+        .port_forward()
+        .await
+        .map_err(|e| anyhow!("NATS service not found in cluster: {e}"))?;
+
+    Ok((format!("nats://127.0.0.1:{port}"), handle))
```
tiagolobocastro: Seems like this should go on kube_proxy like etcd and loki?

k8s/plugin/src/resources/events.rs
```diff
@@ -462,6 +494,107 @@ async fn fetch_from_volume(
...
+/// NATS server only delivers messages after the cutoff — equivalent to the
+/// server-side `--since` filter used in the Loki path.
+///
+/// NATS messages are raw `EventMessage` JSON (no outer envelope). The envelope
+/// `{"type":"mbus_event","payload":{…}}` is added here so `--filter` dot-path
+/// traversal works identically to the Loki and file paths.
+///
+/// Dropping the returned `Messages` subscription closes the consumer;
+/// NATS removes the ephemeral consumer after its inactivity threshold.
+async fn fetch_from_nats(
```
tiagolobocastro, line 520: This should be present in the `rest-plugin` itself, along with the command there
Though happy for this to be as another refactor later since that plugin is for devel only

k8s/plugin/src/resources/events.rs
rohan2794: These are needed because 
- with --nats-endpoint, plugin knows source will be NATS and auto discovery will be tried
- with --nats-endpint <nats_url>, plugin will try NATS source with given url

 num_args = 0..=1 + default_missing_value = "" combination used by clap to support an optional value for a flag

k8s/plugin/src/resources/events.rs
rohan2794: ok

k8s/plugin/Cargo.toml
rohan2794: added to root workspace

k8s/plugin/src/resources/events.rs
rohan2794: fixed now, added NATS support to kube proxy https://github.com/openebs/mayastor-control-plane/pull/1167

k8s/plugin/src/resources/events.rs
Abhinandan-Purkait: Okay, I believe you. :-)

Abhinandan-Purkait (approved):

k8s/plugin/src/resources/events.rs
tiagolobocastro: Then you should update the help above, because in the help it says if "omitted", which doesn't seem true then; it's only if an empty value `""` that we do auto-dicovery right?
Please make sure help is clear, because this is an unusual way of doing cli args

tiagolobocastro (approved):

k8s/plugin/src/resources/events.rs
rohan2794: updated helper comments.

PR #983, opened 2026-08-14
feat(supportability): add dump loki-limit, let LOKI_ENDPOINT env override discovery

datacore-gjanoti:
## Summary
- New `dump loki-limit` subcommand: connects to Loki via the same client every other Loki operation uses, reads its configured `max_query_length` from `/config`, prints it as JSON (`{"maxQueryLength":"<duration>"}`, or `null` if it can't be determined). Lets a caller (e.g. the Headlamp plugin) learn Loki's real query-range limit without building its own separate connection to Loki.
- `loki_endpoint` now also reads from a `LOKI_ENDPOINT` environment variable (via clap's `env` support) when `--loki-endpoint` isn't passed on the command line, ahead of the existing port-forward-based Service discovery fallback. Lets a Helm chart hand the CLI Loki's address directly at deploy time (since the chart already knows it), skipping port-forward entirely for that case.

## Testing
- `cargo check -p supportability` passes.
- Live-tested against a real cluster: `dump loki-limit` returns the correct configured limit in well under a second.
- Live-tested `LOKI_ENDPOINT`: a bogus address fails with a connection error (proving it's read and used, not silently ignored); the real Service address works correctly with no port-forward involved.

### DIFF
--- a/k8s/supportability/src/operations.rs
+++ b/k8s/supportability/src/operations.rs
@@ -28,6 +28,13 @@ pub(crate) enum Resource {
 
     /// Collects the Loki logs from the product's components
     Loki,
+
+    /// Reports Loki's own configured max query-time-range limit, as JSON
+    /// (`{"maxQueryLength":"<duration>"}`, or `null` if it can't be determined).
+    /// Uses the same TLS-aware discovery/connection as every other Loki
+    /// operation this tool performs - a caller wanting this value should use
+    /// this instead of building its own connection to Loki.
+    LokiLimit,
 }
 
 impl SystemDumpArgs {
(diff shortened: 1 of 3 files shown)


k8s/supportability/src/operations.rs
```diff
@@ -28,6 +28,13 @@ pub(crate) enum Resource {
 
     /// Collects the Loki logs from the product's components
     Loki,
+
+    /// Reports Loki's own configured max query-time-range limit, as JSON
+    /// (`{"maxQueryLength":"<duration>"}`, or `null` if it can't be determined).
+    /// Uses the same TLS-aware discovery/connection as every other Loki
+    /// operation this tool performs - a caller wanting this value should use
+    /// this instead of building its own connection to Loki.
+    LokiLimit,
```
Abhinandan-Purkait, line 37: Maybe Hide this or something, as this is not useful for the usual consumers of the plugin? cc @tiagolobocastro wydt?

PR #1014, opened 2026-08-26
feat(chart): allow Guaranteed QoS for the io-engine pod

Scorplionn:
## Problem

Exclusive CPUs under the kubelet `static` CPU manager policy are only granted
to pods in QoS class Guaranteed, which requires matching requests and limits on
**every** container in the pod. Two chart-side gaps make that unreachable for
io-engine today:

- the init containers (`agent-core-grpc-probe`, `etcd-probe`) carry no
  resources, and attaching them means restating the whole
  `base.initContainers.containers` list in values;
- `metrics-exporter-io-engine` has no `resources` section at all, so enabling
  metrics silently drops the pod back to Burstable.

## Changes

- `base.initContainers.defaultResources` — applied only to init containers that
  do not declare their own. Passed explicitly from the io-engine call site, so
  the other five call sites of `render_init_containers` (agent-core, ha-node,
  csi-node, api-rest, jaeger) are unaffected.
- `base.metrics.resources` — for the exporter sidecar.

Both default to empty; rendered output is unchanged for existing installations.

## Why this matters

This is a prerequisite for #962 / mayastor#2050: `--cores-from-cgroup` reads the
pod's cpuset, but without Guaranteed QoS that cpuset is the whole shared pool
rather than the exclusively assigned cores.

Verified on a 32-core node, Kubernetes v1.35, kernel 6.12, with
`cpuManagerPolicy: static` and `strict-cpu-reservation: true`:

    {
        "policyName": "static",
        "defaultCpuSet": "4-31",
        "entries": {
            "00e08908-035e-4e3b-a7db-a8bf82170873": { "io-engine": "2-3" }
        }
    }

## Test plan

- [x] `helm lint`
- [x] `helm template` with defaults — output byte-identical to upstream
- [x] `helm template` with `defaultResources` set — resources appear on the
      io-engine init containers only
- [x] `chart/README.md` regenerated

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -238,12 +238,14 @@ Each Secret must contain `tls.crt`, `tls.key`, and `ca.crt` keys.
 | apis.&ZeroWidthSpace;rest.&ZeroWidthSpace;tolerations | Set tolerations, overrides global | `[]` |
 | base.&ZeroWidthSpace;cache_poll_period | Cache timeout for core agent & diskpool deployment | `"30s"` |
 | base.&ZeroWidthSpace;default_req_timeout | Request timeout for rest & core agents | `"5s"` |
+| base.&ZeroWidthSpace;initContainers.&ZeroWidthSpace;defaultResources | Resources applied to io-engine init containers which do not define their own. Init containers without resources force the pod to QoS class Burstable, which disqualifies it from exclusive CPUs under the static CPU manager policy. Scope is limited to the io-engine DaemonSet: the shared helper receives these defaults explicitly from that call site only. | <pre>{<br><br>}</pre> |
 | base.&ZeroWidthSpace;initContainers.&ZeroWidthSpace;image.&ZeroWidthSpace;registry | Image registry for init containers | `""` |
 | base.&ZeroWidthSpace;logging.&ZeroWidthSpace;color | Enable ansi color code for Pod StdOut/StdErr | `true` |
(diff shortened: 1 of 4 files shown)


tiagolobocastro:
Hey @Scorplionn there's already some pre-existing proposal here:
- https://github.com/openebs/mayastor/pull/2050
- https://github.com/openebs/mayastor-extensions/pull/962

There's also another issue here which is the core list may be changed whilst the pod is running, so just binding to the core list at the time of pod start may not work:
- https://github.com/openebs/spdk-rs/pull/124
- https://github.com/openebs/mayastor/pull/2017

Scorplionn:
Thanks @tiagolobocastro — apologies, I pasted the wrong description into this
PR; it described the cpuManagerAware work from #1015, not what this branch
actually changes. Fixed now.

This PR only adds `base.initContainers.defaultResources` and
`base.metrics.resources`. It doesn't touch the core list at all — it's the QoS
prerequisite that #962 needs: `--cores-from-cgroup` reads the pod's cpuset, but
without Guaranteed QoS that cpuset is the full shared pool rather than the
exclusively assigned cores. Today reaching Guaranteed means restating the whole
`base.initContainers.containers` list in values just to attach resources.

Closing #1015 in favour of #962 and mayastor#2050 — detection inside io-engine
is clearly the better place for it, and good to know about the runtime cpuset
changes; a start-time-only binding wouldn't have covered that.

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

PR #1039, opened 2026-09-08
feat(chart): expose offline volume rebuild settings

yugchaudhari:
The core agent has supported offline volume rebuild since v2.12 behind `--offline-rebuild-enabled`, but the flag was never surfaced in the chart, so there was no way to enable it from values.

Adds `agents.core.rebuild.offline.{enabled,gracePeriod}` next to the existing partial-rebuild settings. Off by default, so installs and upgrades are unaffected. A blank `gracePeriod` omits the arg entirely and leaves the core agent's own default in place rather than pinning one in the chart.

Rendered and checked: defaults emit no args; `enabled=true` emits the flag; both emit together when a grace period is set; and an older values file with no `offline` key (or no `rebuild` key at all) renders without error.

Part of the offline volume rebuild OEP, openebs/openebs#4208.


chart/values.yaml
```diff
@@ -272,6 +272,15 @@ agents:
...
         waitPeriod: ""
+      offline:
+        # -- Rebuild degraded volumes which are not currently published. Without this, a volume
+        # which loses a replica while it has no application attached stays degraded until it is
+        # published again.
+        enabled: false
+        # -- How long a volume must stay degraded before an offline rebuild is started, so that a
+        # node which is only briefly unavailable does not trigger one.
+        # A blank value "" means the internally derived value will be used.
+        gracePeriod: ""
```
tiagolobocastro, line 283: What is the default value?

tiagolobocastro (approved):

chart/values.yaml
yugchaudhari: `10m`, it's a default_value on the core agent's `--offline-rebuild-grace-period`, so a blank chart value just omits the arg and the binary applies its own default.

I've modified the comment and README to reflect that.

Abhinandan-Purkait (approved):

PR #1043, opened 2026-09-09
ci: run basic K8s test on arm

tiagolobocastro:
ci(k8s): run the kind tests on arm64 as well as x86
    
    The k8s CI job now runs on a runner matrix of x86 and arm64. With
    helm-pins it stays x86 only: the pinned chart is architecture
    independent and is pushed once.
    
    The upgrade test is skipped on arm64 hosts: it upgrades from the latest
    released chart, and no release ships arm64 images yet.

---

    feat(chart): stop pinning pods to amd64 nodes
    
    The default nodeSelector, and io_engine's, no longer set
    kubernetes.io/arch to amd64: the images are built for amd64 and arm64.
    A nodeSelector cannot express an OR, so this admits any architecture;
    set kubernetes.io/arch explicitly to confine the pods to one.

### DIFF
--- a/.github/workflows/k8s-ci.yml
+++ b/.github/workflows/k8s-ci.yml
@@ -17,7 +17,13 @@ on:
 
 jobs:
   k8s-ci:
-    runs-on: cncf-ubuntu-8-32-x86
+    strategy:
+      fail-fast: false
+      matrix:
+        # The pinned chart is architecture-independent and is pushed once, so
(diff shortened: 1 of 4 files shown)


tiagolobocastro:
All the E2E tests run entirely on DC via hetzner, so we can't schedule those here
We're looking to open source those so that we can also run them on CNCF infra, though it'd have to be small scale compared to the full suite we have today...

Abhinandan-Purkait (approved):

